PR #10, opened 2022-11-09
tuple variants

StuartHarris:
Now that we are using `serde` to serialize over the boundary, messages can have tuple variants where appropriate, which means they can also be used directly as continuation callbacks. 

e.g. this
```rust
cmd.http.get(FACT_API_URL.to_owned(), Msg::SetFact),
```

rather than this:
```rust
cmd.http.get(FACT_API_URL.to_owned(), |bytes| Msg::SetFact { bytes }),
```


charypar (approved):
Lush!

PR #14, opened 2022-11-15
Update readme

charypar:
Small readme updates to cover the `Request`/`Response` mechanics and clarify some of the technical details of `uniffi`, `wasm-pack` and `serde_generate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Rust Multi-platform Mobile (RMM) ;-)
 
-This repo is a "hello world"-style demonstration of how a single shared library (written in Rust) can act as the single, central core for both mobile apps deployed on either iOS or Android devices, or the Web.
+This repo is a "hello world"-style demonstration of how a single shared library (written in Rust) can act as the single, central core for cross-platform apps deployed on iOS, Android, or the Web. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,70 +11,79 @@ This repo is a "hello world"-style demonstration of how a single shared library
 The fundamental architectural concept used here is the strict separation of pure computational tasks from tasks that cause side effects.
 This concept has been borrowed from [Elm](https://guide.elm-lang.org/architecture/).
 
-### Pure WebAssembly Core
+### Side-effect-free core
+
+In the above diagram, the inner "Cross-platform Rust core" is compiled and linked to the shell on each platform
+as a library (native on iOS, [Java Native Access](https://github.com/java-native-access/jna) and with WebAssembly in the web browsers).
+
+As such, the core is completely isolated and secure against software supply-chain attacks, as it has
```
charypar, line 19: On second thought, that's not strictly true 🤔 a rust library secretly using the network would likely still succeed to make a network call on iOS and Android, wouldn't it...

StuartHarris: It would. I added a bit to suggest that by having Wasm as a target, the core is _forced_ to be pure

StuartHarris (approved):

PR #16, opened 2022-11-16
A documentation guide

charypar:
A longer form documentation guide to make it easier for people to play with this.

```sh
cd docs && mdbook serve --open
```


docs/previous/local-execution.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Make sure you have the following rust targets installed (e.g. `rustup target add <target-name>`). Note there is a `rust-toolchain.toml` in the root directory, so you should be able to just `rustup target list --installed` (in or below the root directory) and the targets should be installed if they are not already present.
+
+   ```txt
+   aarch64-apple-darwin
+   aarch64-apple-ios
+   aarch64-apple-ios-sim
+   aarch64-linux-android
+   wasm32-unknown-unknown
+   ```
```
obmarg, line 13: Need `x86_64-apple-ios` on intel macs

```suggestion
   ```txt
   aarch64-apple-darwin
   aarch64-apple-ios
   aarch64-apple-ios-sim
   aarch64-linux-android
   wasm32-unknown-unknown
   x86_64-apple-ios
   ```
```

docs/src/getting_started/core.md
```diff
@@ -0,0 +1,209 @@
...
+   pub enum Event {
+       Increment,
+       Decrement,
+   }
+   
+   #[derive(Effect)]
+   pub struct Capabilities {
+       pub render: Render<Event>,
+   }
+   ```
```
obmarg, line 187: Not one for now, but might be worth just including the example files here?  https://rust-lang.github.io/mdBook/format/mdbook.html#including-files

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+
+This makes the core portable between platforms, and, importantly, really easy to test. It also separates the intent, the "functional" requirements, from the implementation of the side-effects and the "non-functional" requirements (NFRs). For example, your application knows it wants to store data in a SQL database, but it doesn't need to know or care whether that database is local or remote. That decision can even change as the application evolves. You can read a lot more about how side-effects work in Crux in the chapter on [capabilities](./capabilities.md).
+
+To _ask_ the Shell for side effects, we will need to describe the possible side-effects the Core will request. And to help us create these requests (generated by the `Effect` derive macro), we can use capabilities - reusable bits of functionality which give us a nice API for performing side-effects. We'll look at them in a lot more detail later.
```
obmarg, line 13: Not sure how to fix but I think that "generated by the effect derive macro" bit probably won't be very clear to new users?  Maybe we just don't mention it in this paragraph and leave it for the more detailed bit later?

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+
+```rust
+use serde::{Serialize, Deserialize};
+use crux_core::render::Render;
+use crux_macros::Effect;
+
+#[derive(Effect, Default)]
+struct Capabilities {
+    render: Render<Effect>
```
obmarg, line 22: Capabilities are generic over the event type not the effect type, so this isn't right. Since there's no event type below does this need to be:

```suggestion
    render: Render<()>
```

Although I'm also moderately sure that won't compile with the way the effect macro works now 😢

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+}
+```
+
+For now, we will need a single capability, `Render`, which is built into Crux and available from the `crux_core` crate. It simply tells the shell to update the screen using the latest information. That means the core also produces a single `Effect`, but it will soon be more than one, so we make Effect an enum to prepare for that. Effect also needs to be serializable, so it can be passed across the language boundary between Core and Shell.
```
obmarg, line 26: Think this paragraph also needs updates now we're doing a derive.  Maybe something like:

```suggestion
For now, we will need a single capability, `Render`, which is built into Crux and available from the `crux_core` crate. It simply tells the shell to update the screen using the latest information. The core needs to produce an `Effect` enum, which we derive using `crux_macros::Effect`. Effect needs to be serializable, so it can be passed across the language boundary between Core and Shell, but the macro takes care of that.
```

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+    _marker: PhantomData<fn() -> (E, C)>
+};
+
+impl Default for Hello {
+    fn default() -> Self {
+        Self {
+            _marker: MarkerData
+        }
+    }
+}
```
obmarg, line 47: Pretty sure we can simplify this now:

```suggestion
#[derive(Default)]
pub struct Hello;
```

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+The struct is generic over two types parameters, `E` (for effect) and `C` (for capabilities). This supports composability and reusability of Crux apps (you can read more about that in the chapter on [Composing apps](./composing.md)). These are type parameters corresponding to the types we defined in the previous section. We also need to implement `Default` so that Crux can construct the app for us.
```
obmarg, line 50: Most of this could go in the bin

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+    fn update(&self, _e: Event, _s: &mut Model, caps: &C) -> Vec<Command<Event>> {
+        let render: &Render<_> = caps.get();
+
+        vec![render.render()]
+    }
+
+    fn view(&self, model: &Model) -> ViewModel {
+        "Hello World".to_string()
+    }
+}
```
obmarg, line 75: Lots no longer correct about this, but here's some of the bits:

```suggestion
impl App for Hello where
{
    type Event = ();
    type Model = ();
    type ViewModel = String;
    type Capabilities = Capabilities

    fn update(&self, _e: Event, _s: &mut Model, caps: &Capabilities) -> Vec<Command<Event>> {
        let render: &Render<_> = caps.get();

        vec![render.render()]
    }

    fn view(&self, model: &Model) -> ViewModel {
        "Hello World".to_string()
    }
}
```

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+There's a fair bit going on, so lets look at all of it one by one:
+
+Like the `Hello` struct, the `impl` is generic over `E`, the effect, and `C`, the capabilities. The effect type needs to support cloning and serialisation. The trait bound on `C` says that this app requires the capabilities provided to include the `Render` capability. This may seem quite awkward at first but will become important as a form of dependency injection when we start [composing](./composing.md) apps from reusable modules.
```
obmarg, line 80: Binable (probably)

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+
+The trait has three associated types. The `Event` type describes the possible actions this app can perform. For now its's a unit type. The `Model` describes the internal state of the app, also a unit type for now, as there is no state to hold. The `ViewModel` describes the user interface to display.
```
obmarg, line 82:
```suggestion
The trait has four associated types. The `Event` type describes the possible actions this app can perform. For now its's a unit type. The `Model` describes the internal state of the app, also a unit type for now, as there is no state to hold. The `ViewModel` describes the user interface to display.  The `Capabilities` type lets us control what capaiblities the core has access to.
```

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+        let Command { effect: actual_effect, .. } = commands[0];
+        let expected_effect = Effect::Render;
+
+        // Check update asked us to `Render`
+        assert_eq!(actual_effect, expected_effect);
+
+        let actual_view = hello.view(&model);
+        let expected_view = "Hello World".to_string();
+
+        assert_eq!(actual_view, expected_view);
```
obmarg, line 116: Probably update this to use AppTester?

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+}
+```
+
+We need `Default` implemented to define the initial state. For now we derive it, as our state is quite simple. We also update the app to use the model:
+
+```rust
+impl<E, C> App<E, C> for Hello where
+    E: Clone + Serialize,
+    C: Capabilities<Render<E>>
+{
```
obmarg, line 140:
```suggestion
impl App for Hello where {
```

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+Great. All that's left is adding the behaviour. That's where `Event` comes in:
+
+```rust
+#[derive(Serialize, Deserialize)]
+enum Event {
+    Increment,
+    Decrement,
+    Reset,
+}
+```
```
obmarg, line 162: Suspect we might need an empty event type at the very start of this so the `Effect` macro doesn't go mad

docs/src/guide/hello_world.md
```diff
@@ -0,0 +1,290 @@
...
+
+The event type covers all the possible events we can respond to. "Will that not get massive really quickly??" I hear you ask. Don't worry about that, there is [a way to make this scale](./composing.md). Let's carry on. We need to actually handle those messages.
+
+```rust
+impl<E, C> App<E, C> for Hello where
+    E: Clone + Serialize,
+    C: Capabilities<Render<E>>
+{
```
obmarg, line 170:
```suggestion
impl App for Hello where {
```

obmarg (approved):
Spotted a few things, but could easily be fixed in another PR

docs/src/getting_started/core.md
charypar: Potentially, although they all have extra stuff to explain, that's why I didn't. Maybe we need a dumb hello world example to point to for the basic scaffolding.

docs/src/guide/hello_world.md
charypar: Maybe we can quote that line of code in a parenthesis? Or skip it like you suggest.

In general, we should probably sit a few people down to read it, try to follow it and report every bit of poor clarity back.

docs/src/guide/hello_world.md
charypar: Hmmm, this whole guide is out of date now, isn't it. 😬 Completely missed it, need to go and fix that... I'd rather have no docs than wrong docs

docs/src/getting_started/core.md
StuartHarris: I'll write your hello world as an example in a separate PR

StuartHarris: The example is in this [PR](https://github.com/redbadger/crux/pull/37/files). The order in which the example is built up is tricky because the `Effect` derive macro will fail until the rest of the example is in place.

docs/src/guide/hello_world.md
StuartHarris: agree. I had a go at explaining the macro a bit (it _is_ relevant here). LMK if that works for now

docs/src/getting_started/core.md
StuartHarris: I updated the code in this tutorial to match the (as yet unmerged) example, so hopefully should be OK

StuartHarris (approved):

docs/src/getting_started/core.md
StuartHarris: Once the other PR is merged, we could include the whole example in this page as per @obmarg's excellent suggestion above.

PR #22, opened 2022-11-23
Add basic cors policy to nextjs example repo

niall-rb:
Right now the cat facts example is showing a runtime error when we start the repo locally:

Screenshot 2022-11-23 at 11 39 05

The reason why is that Nextjs implements a Same Origin Policy by default and so our attempt to return an image from aws.random.cat/meow is met with a 503. 

This PR adds a basic cors policy as part of our nextjs config to remove this error.


StuartHarris (approved):
Thanks Niall! 🙏

PR #23, opened 2022-11-24
Fix Android code generation for any package name

matt-thomson:
Currently, the type generation assumes that the target code will live under the com.redbadger.crux_core.shared_types package.  However, users of crux can specify their own package, so this isn't always the case.

This commit changes the type generation to set the package in the Requests.java file, and copy it to the correct package directory.  It also changes the package in the cat facts example, to demonstrate this working correctly.


StuartHarris (approved):
Amazing! Thanks, Matt 🙏

PR #25, opened 2022-12-04
Add `x86_64-apple-ios` to targets list

obmarg:
This seems to be needed for the iOS simulator on intel macs.


StuartHarris:
Thanks Graeme! Ahh, that's what it's for! I did wonder. Should we add it to `rustfmt.toml` as well?

StuartHarris (approved):
❤️

PR #26, opened 2022-12-07
Async capabilities

obmarg:
There's been some work recently to formalise some capability APIs in crux.  The
existing implementation uses the `Command` abstraction to pass an effect back
up to the platform specific shell.  When the shell is done carrying out this
effect it passes back a response, which is fed through a callback to produce an
Event, which is then fed into the app.  Nice and simple.

However, this limits what can be acheived in the rust part of capabilities:

- There's always a 1:1 mapping between the `Command`s created by an app and the
  responses back into the core.  This is often what you need, but not always.
- If a capability requires further processing before handing an event back to the
  app the app specifically needs to wire up the communication for that.

This PR exposes an alternative approach to capabilities:

- Each capability gets a `CapabilityContext`.  This exposes three functions to
  capabilities:
  - `spawn` allows a capability to spawn a future onto an executor that is run
    whenever the `Core` receives input from the shell.
  - `effect` which passes an effect back up to the core and returns a future
    that resolves after the shell has carried out the effect.
  - `send_event` allows the capability to schedule an event to be sent into the
    `App::update` function.

Putting these together allows capabilities to spawn some async code, run one or
more effects through the shell (without requiring the app to get involved) and
when it's done send one or more events back into the App.

Some other misc changes in this code:

1. Since capability side effects are now smuggled through channels, the return
   value of the `update` function is (imo) not too useful anymore.  So it's gone.
2. `Capabilities` is now an associated type on the `App` trait rather than a generic
   param.
3. These two combined allowed the `App` trait to be a fair bit simpler.
4. `ContinuationStore` now uses a Mutex.  It was using a `RwLock` when it didn't
   really need to, and this led to some stricter bounds on the callbacks
   put into it than neccesary.
5. Fixed a bug (or at least bad behaviour imo) in the yew-web impl where it ignored
   any effects in a batch that were after a render effect (because `any` is short
   circuiting).


crux_core/src/lib.rs
```diff
@@ -261,3 +291,11 @@ pub struct Request<Effect> {
     pub uuid: Vec<u8>,
     pub effect: Effect,
 }
+
+// This name is too confusing, since we kind of have two "Effects"
+// We've got the `Effect` enum that gets serialised to the shell
+// and the capability level stuff, which this represnets.
+// TODO: Think up a better name...
```
obmarg, line 298: Any thoughts on this?

charypar: I'm not sure I fully underatand what this is - is it the case that the `Effect` defined by the app to enumerate the possible effects the core will emit implements this trait?

crux_core/src/render.rs
```diff
@@ -1,27 +1,34 @@
...
-        Self { make_effect }
+    pub fn new(context: CapabilityContext<(), Ev>) -> Self {
+        Self { context }
     }
 
-    pub fn render<Ev>(&self) -> Command<Ef, Ev>
+    pub fn render(&self) {
+        let ctx = self.context.clone();
+        self.context
+            .spawn(async move { ctx.run_command(Command::new_without_callback(())) })
```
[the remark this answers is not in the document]
charypar: Hmm, reading the other capabilities, I'm not sure why this runs a command (rather than emitting an effect), or what command is now.

charypar (approved):
Overall, looks great, happy to merge as is and make any further tweaks separately. These revamped capabilities are a lot more powerful than the previous one which is very cool, and I can't believe it just works in webassembly. Great work!

Just as general feedback: 

* I get a little bit lost in the names of the futures machinery (spawner, executor, waker, ...). I also think we can probably find a nicer structure and naming for the capability context API, making it a bit more clear what each interaction is doing and "who" is it doing it to (shell, or app). 
* It would be helpful to somehow highlight that the capability calls from `App::update` behave a little like go routines - they get scheduled to execute "concurrently" once the update call is done. Not sure that's clear and also not sure how to make it clearer that you can get to the effects/futures that have been queued up from those calls and investigate them (so that people's first instinct for testing isn't to mock the capability API surface).
* I'm not entirely sure I understand the role of command now, it feels like it has been entirely superseded, but maybe I'm missing something?

crux_core/src/capability.rs
```diff
@@ -1,6 +1,119 @@
...
+        });
+
+        CapabilityContext { inner }
+    }
+
+    pub(crate) fn run_command(&self, cmd: Command<Ef>) {
+        self.inner.command_sender.send(cmd);
+    }
+
+    pub fn spawn(&self, f: impl Future<Output = ()> + 'static + Send) {
```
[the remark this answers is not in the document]
obmarg: > I can't quite put my finger on it, but I think something around this API should nod to the fact that this is essentially the capability's Shell interface, especially the effect method.

Yeah, broadly speaking I'm not against renaming/rejigging some of this stuff.  It's the most "user" (well capability implementor) visibile bit of this so yeah.

> It might be that the parts talking to the shell (effect) and the parts talking back to the app (send_event) should be separate?

I don't know.  I definitely see where you're coming from but wouldn't a capability be useless without both of them?  Like you'd just end up having two parameters/fields everywhere instead of one.

Exposing an API like `ctx.core.send_event()` could be nice mind you, although that would mean introducing more things so not sure it's actually a good idea...?

> Maybe that would also avoid the need to clone the context? Is the context doing to many jobs...?

I'm not sure reorganising things would remove the need for a clone.  Our futures are `'static` so they can't borrow.  They need to be able to send events to the core and since they can't borrow they need to own the thing that does that.  So there needs to be a clone _somewhere_.

We could maybe go down the path of associating a lifetime with our futures.  Not inconceivable that they could inherit a lifetime from the core.  But I do not know that it's worth it - don't think it's an expensive clone, and putting lifetimes on things could open up a whole new world of pain.  You'd still not be able to borrow the `&self` so would probably still need to do some `let ctx = &ctx` type thing. 

One "simplification" I was thinking about was making the futures `impl Future<Output=Ev>`, which would remove the need for `send_event`.  But then it'd be impossible to build any kind of subscription capability on top of that API, so not sure it's actually better?

crux_core/src/lib.rs
```diff
@@ -191,29 +180,55 @@ where
...
             bcs::from_bytes(msg).expect("Message deserialization failed.");
 
-        let commands = {
+        let mut model = self.model.write().expect("Model RwLock was poisoned.");
+        self.app.update(event, &mut model, &self.capabilities);
+        drop(model);
+
+        self.executor.run_all();
+
+        while let Some(event) = self.event_receiver.receive() {
```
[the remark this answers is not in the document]
obmarg: Yeah, I mean _arguably_ you might not need this.  The shell hasn't passed anything in so it is unlikely that any new events have been generated by capabilities.  But I think it would be a mistake to assume that's always true.

Possible that a capability could contain a cache for example and could answer (or partially answer) a request without ever needing to send off an effect.

crux_core/src/render.rs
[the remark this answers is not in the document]
obmarg: The `()` is just because `Command` has to package up some effect.  The render command doesn't have any payload so I just put a `()` in here.  The `CapabilityFactory` provided by the user handles mapping that into the `Effect::Render` the user provides.

`Command` still exists to ferry effects & a continuation callback up to the `Core`.  It's what `fn effect` is using under the hood.

But render was a bit of an odd capability that didn't have a callback - and the `effect` function always registers one.  For the sake of not changing it I just made this one run a command directly.  Atlhough could easily change it if we'd rather consistency?

obmarg:
> I get a little bit lost in the names of the futures machinery (spawner, executor, waker, ...). I also think we can probably find a nicer structure and naming for the capability context API, making it a bit more clear what each interaction is doing and "who" is it doing it to (shell, or app).

Agreed re: the futures machinery.  Although it's standard shit so not sure we'd come up with anything better 😆.

> It would be helpful to somehow highlight that the capability calls from App::update behave a little like go routines - they get scheduled to execute "concurrently" once the update call is done. Not sure that's clear.

Yeah, open to any suggestions on this.

> also not sure how to make it clearer that you can get to the effects/futures that have been queued up from those calls and investigate them (so that people's first instinct for testing isn't to mock the capability API surface).

So I've got that testing API I showed you lined up on another branch that I wanted to submit for review once this is merged.  That + some documentation might be the way?  Think we've got rusts staticly typed nature on our side as well - people might _want_ to mock things, but eh... good luck doing it 😆.


> I'm not entirely sure I understand the role of command now, it feels like it has been entirely superseded, but maybe I'm missing something?

It has been removed as something a user needs to care about (and is now private API) but is still useful as a thing that contains an effect and the continuation callback for that effect.

StuartHarris:
> people might want to mock things, but eh... good luck doing it 😆.
Haha! Love this! I think that all the docs, test helpers, examples, tests etc will quickly get folks out of reaching for mocks. Anyone who is here already knows.

crux_core/src/lib.rs
```diff
@@ -191,29 +180,55 @@ where
...
+        <A as App>::Event: Deserialize<'de>,
     {
-        let msg: <A as App<_, _>>::Event =
+        let event: <A as App>::Event =
             bcs::from_bytes(msg).expect("Message deserialization failed.");
 
-        let commands = {
+        let mut model = self.model.write().expect("Model RwLock was poisoned.");
+        self.app.update(event, &mut model, &self.capabilities);
+        drop(model);
```
StuartHarris, line 216: Is this explicit `drop` for a reason I don't understand, as in, is it equivalent to wrap the previous 2 lines in braces?

crux_core/src/lib.rs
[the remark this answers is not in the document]
StuartHarris: Yeah, I could easily see caching capabilities

crux_core/src/lib.rs
StuartHarris: Does the name `Protocol` resonate? I dunno, the capabilaitiy's protocol with the shell is basically it's tx/rx right? So maybe 🤷

StuartHarris (approved):

crux_core/src/lib.rs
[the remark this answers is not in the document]
charypar: Yes, ok that all makes sense, and caching capabilities sound like a useful thing

crux_core/src/lib.rs
charypar: Stealing from Smithy, we could call it `Operation` with an associated type called `Result` or `ReturnValue`

charypar:
I think this is ready to go, if you're both happy @obmarg @StuartHarris

crux_core/src/capability.rs
```diff
@@ -43,64 +44,68 @@ impl<Ef, Ev> Clone for CapabilityContext<Ef, Ev> {
...
+    pub fn notify_shell(&self, operation: T) {
+        self.inner.steps.send(Step::once(operation));
     }
 
-    pub fn map_effect<NewEf, F>(&self, func: F) -> CapabilityContext<NewEf, Ev>
+    pub fn update_app(&self, event: Ev) {
+        self.inner.events.send(event);
+    }
+
+    pub fn with_effect<OtherT, F>(&self, func: F) -> CapabilityContext<OtherT, Ev>
```
obmarg, line 78: Nice, yeah - this is a better name.

crux_core/src/steps.rs
```diff
@@ -0,0 +1,88 @@
...
+    where
+        F: Fn(&[u8]) + Send + 'static,
+    {
+        Self {
+            payload,
+            resolve: Some(Box::new(resume)),
+        }
+    }
+
+    pub fn once(payload: T) -> Self {
```
obmarg, line 24: Much better name this

crux_core/src/steps.rs
```diff
@@ -0,0 +1,88 @@
...
+            payload: effect,
+            resolve,
+        } = step;
+
+        let uuid = *Uuid::new_v4().as_bytes();
+
+        if let Some(resolve) = resolve {
+            self.0
+                .lock()
+                .expect("Continuation Mutex poisoned.")
```
obmarg, line 66:
```suggestion
                .expect("StepRegistry Mutex poisoned.")
```

crux_core/src/steps.rs
```diff
@@ -0,0 +1,88 @@
...
+            uuid: uuid.to_vec(),
+            effect,
+        }
+    }
+
+    pub(crate) fn resume(&self, uuid: &[u8], body: &[u8]) {
+        let resolve = self
+            .0
+            .lock()
+            .expect("Continuation Mutex poisoned.")
```
obmarg, line 81:
```suggestion
            .expect("StepRegistry Mutex poisoned.")
```

crux_core/src/steps.rs
```diff
@@ -0,0 +1,88 @@
...
+            .0
+            .remove(uuid)
+            .unwrap_or_else(|| panic!("Continuation with UUID {uuid:?} not found."));
```
obmarg, line 84:
```suggestion
            .unwrap_or_else(|| panic!("Step with UUID {uuid:?} not found."));
```

crux_core/src/lib.rs
obmarg: I was actually thinking `protocol.rs` for the file where we define this stuff in my expanded http capability, so I do quite like it as the name for this trait.  But I'm also happy with `Operation` as the name for a this?  Suppose a protocol _could_ be made up of more than one of these `Operations`, so the `Operation` name makes a bit more sense...

PR #28, opened 2022-12-12
Counter example

StuartHarris:
Add a simple counter example (increments and decrements a single global integer, via an API).

- [x] Server API, deployed to fly.io, that supports
    - [x] GET /
    - [x] POST /inc
    - [x] POST /dec
    - [x] GET /sse (for server-sent events, and subscriptions)
- [x] Shared core with tests
- [x] CLI
- [x] iOS app
- [x] Android app
- [x] Yew app
- [x] NextJS app


examples/counter/shared/src/app/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+        let app = AppTester::<App, _>::default();
+        let mut model = Model::default();
+
+        let update = app.update(Event::Set(Counter { value: 1 }), &mut model);
+
+        let actual = &update.effects[0];
+        let expected = &Effect::Render;
+        assert_eq!(actual, expected);
+
+        let actual = model.count;
```
StuartHarris, line 120: should we also test `view()` here?

crux_http/src/lib.rs
```diff
@@ -87,6 +87,26 @@ where
...
+        let ctx = self.context.clone();
+        self.context.spawn(async move {
+            let request = HttpRequest {
+                method: HttpMethod::Post.to_string(),
+                url: url.to_string(),
+            };
+            let resp = ctx.request_from_shell(request).await;
+
+            let data = serde_json::from_slice::<Res>(&resp.body)
+                .expect("TODO: do something sensible here");
```
charypar, line 104: Should `Output` be `Result<HttpResponse, HttpError>`? Http is (famously) fallible after all.

examples/counter/cli/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+                        Ok(bytes) => {
+                            queue.push_back(CoreMessage::Response(
+                                uuid,
+                                Outcome::Http(HttpResponse {
+                                    status: 200,
+                                    body: bytes,
+                                }),
+                            ));
+                        }
+                        Err(e) => bail!("Could not HTTP GET from {}: {}", &url, e),
```
charypar, line 88: As in the other comment, suggest Http should support error handling

examples/counter/server/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+    let stream = stream::repeat_with(move || {
+        Event::default().data(
+            &serde_json::to_string(&Counter {
+                value: *counter.value.lock().unwrap(),
+            })
+            .unwrap(),
+        )
+    })
+    .map(Ok)
+    .throttle(Duration::from_secs(1));
```
charypar, line 97: Is it an event every second, or an event every time a counter changes, but _at most_ once a second? The stream could easily watch the counter / get updated every time anyone runs inc / dec, right?

examples/counter/shared/src/app/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#[derive(Default)]
+pub struct Model {
+    count: isize,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ViewModel {
+    pub count: isize,
```
charypar, line 51: This could be an already formatted string, to make the distinction between model and view model a bit clerer?

examples/counter/shared/src/app/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn update(&self, msg: Self::Event, model: &mut Self::Model, caps: &Self::Capabilities) {
+        match msg {
+            Event::Get => {
+                caps.http.get_json(Url::parse(API_URL).unwrap(), Event::Set);
+            }
+            Event::Set(counter) => {
+                model.count = counter.value;
+                caps.render.render();
+            }
+            Event::Increment => {
```
charypar, line 26: We could do optimistic state updates here too (maybe with an extra bit of state indicating whether the value is pending confirmation from the server or something). It would make for a better incremental example in the docs, maybe, starting with a local one, then adding in the remote API on top.

charypar: Alternatively, some state keeping track of the fact that we've sent something and are awaiting server response, so feedback can be given in the UI.

examples/counter/shared/src/app/mod.rs
charypar: I would, possibly before checking the model. In some sense it's black box (via view) vs white box (via model) testing.

charypar (approved):
Nice! Made a few suggestions here and there, but good example code!

crux_http/src/lib.rs
StuartHarris: yeah I agree. We should also look at @obmarg 's http capability based on surf. (so didn't want to jump to far just yet)

examples/counter/server/src/main.rs
StuartHarris: yeah, it feels as though it should give you the value once, and then when it changes. But I just did a copy-paste for now. We should look at this when we update the capability to support subscriptions?

examples/counter/shared/src/app/mod.rs
StuartHarris: yeah, like the optimistic update idea

examples/counter/shared/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+    CORE.response(uuid, data)
+}
+
+#[wasm_bindgen]
+pub fn view() -> Vec<u8> {
+    CORE.view()
+}
+
+// TODO macro effect generation:
+// crux_macros::generate_effect!(Effect, MyCapabilities);
```
obmarg, line 38: An idea I had for this was to use [typetag](https://github.com/dtolnay/typetag).  Suspect we could set it up to serlalize a `dyn Operation` and get rid of the `Effect` enum altogether.

obmarg: Of course I just looked into it a tiny bit, and _suspect_ it won't work on wasm.  So never mind 😢

StuartHarris: Hmm. That would've been great. I got quite excited for a moment. We can write a macro to hide it

examples/counter/shared/src/app/mod.rs
StuartHarris: Although our API is sooo fast you wouldn't see it anyway

examples/counter/shared/src/app/mod.rs
StuartHarris: Yeah we'll need to expose it

examples/counter/shared/src/lib.rs
charypar: Would that lose us the exhaustive pattern matching / switch on the shell side? It would have no way of knowing what the full list of tags is that it needs to handle, wouldn't it?

obmarg: Pretty sure that would have been solvable.  With `serde_generate` I'm 90% sure it'd just work, with `go-away` I could implement a similar thing to what typetag does under the hood.

examples/counter/shared/src/app/mod.rs
StuartHarris: will do this in another pr

PR #29, opened 2022-12-14
Rustdoc for version 0.2

charypar:
Update rustdoc for all the public APIs

## Todo
- [x] crux_core
   - [x] Documentation for app facing API
   - [x] Documentation for capabilities facing API
- [x] crux_http
- [x] crux_kv
- [x] crux_time

### DIFF
--- a/crux_core/src/capability.rs
+++ b/crux_core/src/capability.rs
@@ -364,8 +364,8 @@ where
         CapabilityContext { inner }
     }
 
-    /// Transform the CapabilityContext into one maps each event passed to
-    /// `update_app` using the provided function.
+    /// Transform the CapabilityContext into one which uses the profived function to
+    /// map each event dispatched with `update_app` to a different event type.
     ///
     /// This is useful when composing apps from modules to wrap a submodule's
     /// event type with a specific variant of the parent module's event, so it can
--- a/crux_core/src/capability.rs
+++ b/crux_core/src/capability.rs
@@ -1,14 +1,108 @@
+//! Capabilities provide a user-friendly API to request side-effects from the shell.
+//!
+//! Typically, capabilities provide I/O and host API access. Capabilities are external to the
+//! core Crux library. Some are part of the Crux core distribution, others are expected to be built by the
+//! community. Apps can also build single-use capabilities where necessary.
+//!
+//! # Example use
+//!
+//! A typical use of a capability would look like the following:
+//!
+//! ```rust,ignore
+//! fn update(&self, event: Self::Event, model: &mut Self::Model, caps: &Self::Capabilities) {
+//!     match event {
+//!         ...         
+//!         Event::Increment => {
+//!             model.count += 1;
+//!             caps.render.render(); // Render capability
+//!
+//!             let base = Url::parse(API_URL).unwrap();
+//!             let url = base.join("/inc");
+//!             caps.http.post(url.unwrap(), Event::Set) // HTTP client capability
+//!         }
+//!     }
+//! }
+//! ```
+//!
+//! Capabilities don't _perform_ side-effects themselves, they request them from the Shell. As a consequence
+//! The capability calls within the `update` function **only queue up the requests**. The side-effects themselves
+//! are performed concurrently and don't block the update function.
+//!
+//! In order to use a capability, the app needs to include it in its `Capabilities` associated type and `WithContext`
+//! trait implementation. For example:
+//!
+//! ```rust,ignore
+//! use crux_http::{Http, HttpRequest};
+//!
+//! enum Effect {
+//!     Http(HttpRequest)
+//! }
+//!
+//! #[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+//! struct Capabilities {
+//!     http: Http<Effect>
+//! }
+//!
+//! impl crux_core::WithContext<App, Effect> for Capabilities {
+//!    fn new_with_context(context: CapabilityContext<Effect, Event>) -> Capabilities {
+//!        Capabilities {
+//!            http: Http::new(context.with_effect(Effect::Http)),
+//!        }
+//!    }
+//! }
+//! ```
+//!
+//! This links the (reusable) capabilities with the app-specific `Effect` type, used by the Shell to dispatch
+//! side-effect requests to the right capability implementation (and, in some languages, checking that all necessary
+//! capabilities are implemented).
+//!
+//! # Implementing a capability
+//!
+//! TODO. See the crux_http crate as a working example.
+
 use std::sync::Arc;
 
 use futures::Future;
 
 use crate::{channels::Sender, Step};
 
+/// Operation trait links together input and output of a side-effect.
+///
+/// You implement `Operation` on the payload sent by the capability to the shell using [`CapabilityContext::request_from_shell`].
+///
+/// For example (from `crux_http`):
+///
+/// ```rust,ignore
+/// impl Operation for HttpRequest {
+///     type Output = HttpResponse;
+/// }
+/// ```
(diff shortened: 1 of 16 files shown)


crux_core/src/capability.rs
```diff
@@ -1,14 +1,231 @@
+//! Capabilities provide a user-friendly API to request side-effects from the shell.
```
obmarg, line 1: We might want to put `#![warn(missing_docs)]` on this at some point?  Makes it harder to forget to add them.  Ofc it's also annoying af when you're working on stuff, so maybe one for later

crux_core/src/capability.rs
```diff
@@ -63,10 +319,14 @@ where
         CapabilityContext { inner }
     }
 
+    /// Spawn a task to do the asynchronous work. Within the task, async code
+    /// can be used to interact with the Shell.
```
obmarg, line 323:
```suggestion
    /// can be used to interact with the Shell and the App.
```

crux_core/src/capability.rs
```diff
@@ -93,6 +364,28 @@ where
         CapabilityContext { inner }
     }
 
+    /// Transform the CapabilityContext into one maps each event passed to
+    /// `update_app` using the provided function.
```
obmarg, line 368: This paragraph definitely needs fixed (i'd make a suggestion but I'm not certain what it was meant to say)

crux_core/src/capability.rs
```diff
@@ -93,6 +364,28 @@ where
...
+    ///
+    /// This is useful when composing apps from modules to wrap submodule's
```
obmarg, line 370:
```suggestion
    /// This is useful when composing apps from modules to wrap a submodule's
```

crux_core/src/capability.rs
```diff
@@ -93,6 +364,28 @@ where
...
+    /// event type with a specific variant of parent module event, so it can
```
obmarg, line 371:
```suggestion
    /// event type with a specific variant of the parent module's event, so it can
```

crux_core/src/future.rs
```diff
@@ -1,3 +1,5 @@
+//! Async support for implementing capabilities
```
obmarg, line 1: This module isn't particularly useful from a documentation perspective...

Wonder if we want to re-jig it a bit.  Maybe make it a submodule of `capability` and just re-export `EffectFuture` from `capability`?  Not sure.

crux_core/src/lib.rs
```diff
@@ -132,24 +180,30 @@ pub use capability::{Capability, WithContext};
...
     /// Model, typically a `struct` defines the internal state of the application
     type Model: Default;
-    /// Message, typically an `enum`, defines the actions that can be taken to update the application state.
+    /// Event, typically an `enum`, defines the actions that can be taken to update the application state.
     type Event: Send + 'static;
     /// ViewModel, typically a `struct` describes the user interface that should be
     /// displayed to the user
     type ViewModel: Serialize;
-
+    /// Capabilites, typically a `struct` lists the capabilities used by this application
```
obmarg, line 188:
```suggestion
    /// Capabilites, typically a `struct`, lists the capabilities used by this application
```

obmarg (commented):
Looks decent overall.  Should we get someone who's not super familiar with the code to review at some point?

A general comment (that I'm not sure how to solve) - it'd be nice if we hadn't `ignore`d so many of the code samples.  I recognise that most of them need an uncomfortable amount of context in order to actually compile, but seems very likely that we'll miss breakages if we're not compiling them...

charypar:
> A general comment (that I'm not sure how to solve) - it'd be nice if we hadn't `ignore`d so many of the code samples. I recognise that most of them need an uncomfortable amount of context in order to actually compile, but seems very likely that we'll miss breakages if we're not compiling them...

Yea, I tried getting the key ones to compile and test, but there's a lot of ignored ones for sure. Wonder if there's a way to hide some of the context from code samples, or whether people just need to deal with longer samples...

crux_core/src/capability.rs
```diff
@@ -1,14 +1,231 @@
...
+//! side-effect requests to the right capability implementation (and, in some languages, checking that all necessary
+//! capabilities are implemented).
+//!
+//! # Implementing a capability
+//!
+//! Capabilities provide an interface to request side-effects. The interface has asynchronous semantics
+//! with a form of callback. A typical capability call can look like this:
+//!
+//! ```rust,ignore
+//! caps.ducks.get_in_a_row(10, Event::RowOfDucks)
```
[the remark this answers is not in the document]
charypar: I know! Feel exactly the same way... 😔

obmarg:
> Yea, I tried getting the key ones to compile and test, but there's a lot of ignored ones for sure. Wonder if there's a way to hide some of the context from code samples, or whether people just need to deal with longer samples...

You can hide context by sticking a ` #` at the start of the code in a docstring.  I tried it with a few of them though and it was still uncomfortable writing it.  Although maybe better than eventually having wrong docs

crux_core/src/future.rs
charypar: The module isn't, but it contains one of the most important capability APIs. It'd be nice if we can make it part of `capability`.

obmarg:
We could arguably also stick a `#[doc(hidden)]pub mod __docstring_helpers { ... }` somewhere to share common doctest code.

A bit messy, but could make it more concise...

crux_core/src/future.rs
obmarg: The function is defined in this file but it's part of `CapabilityContext` which isn't defined in this file, so rustdoc puts the docs for it elsewhere. 

Ofc the future module itself isn't public so doesn't appear in the documentation anyway.  So lets consider this comment resolved

crux_core/src/capability.rs
```diff
@@ -1,14 +1,231 @@
...
+//!            }
+//!        }
+//!     }
+//! }
+//! }
+//! ```
+//!
+//! This links the (reusable) capabilities with the app-specific `Effect` type, used by the Shell to dispatch
+//! side-effect requests to the right capability implementation (and, in some languages, checking that all necessary
+//! capabilities are implemented).
```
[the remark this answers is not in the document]
charypar: I gave it a go.

charypar:
> You can hide context by sticking a ` #` at the start of the line in a docstring. I tried it with a few of them though and it was still uncomfortable writing it. Although maybe better than eventually having wrong docs

Oh that's useful, I might have a go at that as a second pass. Do you think this is now good to go @obmarg?

crux_core/src/capability.rs
```diff
@@ -1,14 +1,231 @@
...
+//!
+//! Typically, capabilities provide I/O and host API access. Capabilities are external to the
+//! core Crux library. Some are part of the Crux core distribution, others are expected to be built by the
+//! community. Apps can also build single-use capabilities where necessary.
+//!
+//! # Example use
+//!
+//! A typical use of a capability would look like the following:
+//!
+//! ```rust,ignore
```
StuartHarris, line 11: these `ignore` attributes (that I peppered around the place a while back) should go. We could use `no_run`, but the code would need to compile so we'd need to add enough code around the example (using `#` to hide) to make it compile. Probs a job for another day though

crux_core/src/capability.rs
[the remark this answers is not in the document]
StuartHarris: me too. But I think more explicit is usually better, so `capabilities`. Also if your IDE is implementing the trait for you, there shouldn't be any typing

StuartHarris (approved):

crux_core/src/capability.rs
StuartHarris: ha! Sorry just read the comments above! Glad we landed in the same place

PR #30, opened 2022-12-15
Flesh out `crux_http` significantly.

obmarg:
The existing crux_http API is really very simple.  It supports a handful of things, but nowhere near everything an actual HTTP library would need to do.

This PR updates `crux_http` with a lot of borrowed code from [`surf`](https://github.com/http-rs/surf).  I'm a big fan of the `surf` API, and don't think that re-inventing the wheel on this would be a good idea.  

There are a few awkward bits of this conversion: surf is an async/await API, and crux is obviously based on callbacks into the `update` function.  So I've had to remove a lot of async bits. `surf` also supports middlewares though, which continue to use `async/await` in this conversion.  This has probably left the API in a pretty weird state.  But hopefully this is something we can come back to and clean up as people start to use it

Worth noting that this PR doesn't update the protocol between the shell and the capability, so some of the API (particularly around headers) won't actually do anything at present.  But that can probalby come in a future, smaller PR.

I've also not really figured out how to allow people to set config or middleware yet.  The functionality is there in the code, but since we're expecting people to derive `WithContext` for the `Capabilities` struct it's not really clear how they'd set stuff like this up.  Can probably tweak some APIs to make it possible, but I didn't want to do that in this PR.


StuartHarris (approved):
Amazing!

StuartHarris:
Just noticed `-no-default-features` in `Cargo.toml`, which I think is a CLI flag instead, so changed it to `default-features = false, features = ["fs"]`

PR #31, opened 2022-12-16
Make `channels` & `executor` private.

obmarg:
Noticed these sitting undocumented in the 0.2 docs.  At one point I thought these might be public API, but we've ended up hiding them sufficiently.  So rather than documenting them, lets just hide them from users.


StuartHarris (approved):
Good idea!

PR #33, opened 2022-12-24
Macro to generate effect enum and `WithContext` impl

StuartHarris:
Generate code to declare the Effect enum, and implement `WithContext` on the capabilities struct.

This implementation is as good as we can do I think. Proc macros can't use any info that is not contained in the incoming token streams (2 streams for attribute macros — params and attributed item, or 1 stream for derive macros — the item itself, including non-macro attributes), so we can't go looking round the codebase for referenced types :-(

from the readme...

# Crux Macros

This crate provides macros that can be used in conjunction with [`crux_core`](https://crates.io/crates/crux_core) and associated Capabilities.

## Effect

The `Effect` derive macro can be used to create an effect enum for use by the Shell when performing side effects for the Core. It also derives `WithContext` for the Capabilities struct.

The name of the generated enum is `Effect` by default, but can be overridden (see below). The macro also needs to know the name of your app struct (which is `App` by default, but can be specified).

It also needs to know the operation types for any capabilities you are using that have non-unit structs as their request types (have the `Operation` trait implemented).

> The macro should really, by convention, be called `WithContext`, but it's possible that the name `Effect` more usefully describes the code that is generated.

> It is implemented as a derive macro, rather than an attribute macro, because it needs to be configured by non-macro attributes within the struct itself (which is not possible with attribute macros).

### Example usage

If you want to generate an enum called `Effect` and your app struct is called `App`, and the capabilities you are using only have unit operations, then you can simply just use the macro without additional configuration:

```rust
#[derive(Effect)]
pub struct Capabilities {
    pub render: Render<Event>,
}
```

If, however, you are using capabilities with non-unit operations (i.e. that take request data), you have to tell the macro what the operation struct is called (this is the struct that implements the `Operation` trait):

```rust
#[derive(Effect)]
pub struct Capabilities {
    #[effect(operation = "HttpRequest")]
    pub http: Http<Event>,
    pub render: Render<Event>,
}
```

If you want the generated Effect enum to be called something different, you can specify another name:

```rust
#[derive(Effect)]
#[effect(name = "MyEffect")]
pub struct Capabilities {
    pub render: Render<Event>,
}
```

If your app struct (that implements the `App` trait) is called something other than `App`, you can specify its name:

```rust
#[derive(Effect)]
#[effect(app = "MyApp")]
pub struct Capabilities {
    pub render: Render<Event>,
}
```

To specify both `app` _and_ `name` you can either use the attribute twice, like this:

```rust
#[derive(Effect)]
#[effect(name = "MyEffect")]
#[effect(app = "MyApp")]
pub struct Capabilities {
    pub render: Render<Event>,
}
```

Or, more idiomatically, combine them into one usage, like this:

```rust
#[derive(Effect)]
#[effect(name = "MyEffect", app = "MyApp")]
pub struct Capabilities {
    pub render: Render<Event>,
}
```

Full usage might look something like this:

```rust
#[derive(Effect)]
#[effect(name = "MyEffect", app = "MyApp")]
pub struct CatFactCapabilities {
    #[effect(operation = "HttpRequest")]
    pub http: Http<MyEvent>,
    #[effect(operation = "KeyValueOperation")]
    pub key_value: KeyValue<MyEvent>,
    pub platform: Platform<MyEvent>,
    pub render: Render<MyEvent>,
    pub time: Time<MyEvent>,
}
```


crux_http/tests/with_tester.rs
```diff
@@ -1,87 +1,77 @@
...
-
-    impl WithContext<MyApp, MyEffect> for MyCapabilities {
-        fn new_with_context(context: CapabilityContext<MyEffect, MyEvent>) -> MyCapabilities {
-            MyCapabilities {
-                http: Http::new(context.with_effect(MyEffect::Http)),
-            }
-        }
+    #[derive(Effect)]
+    pub(crate) struct Capabilities {
+        #[effect(operation = "HttpRequest")]
```
obmarg, line 55: Having to specify the operation here is a bit annoying.  Wonder if we could add an associated type to the `Capability` trait and just get it from there?

Suppose we'd need some way to differentiate between capabilities with data and those without, but reckon it might be do-able.

Suppose it could also make having a capability support >1 operation painful, although maybe we cross that bridge when we come to it?

crux_macros/src/effect.rs
```diff
@@ -0,0 +1,98 @@
...
+        .take_struct()
+        .expect("Should never be enum")
+        .fields;
+
+    let (variants, fields): (Vec<_>, Vec<_>) = fields
+        .into_iter()
+        .enumerate()
+        .map(|(i, f)| {
+            // This works with named or indexed fields, so we'll fall back to the index so we can
+            // write the output as a key-value pair.
```
obmarg, line 57: You've told darling you only support struct_named inputs, so I don't know if you'll ever not have a named field...

crux_macros/src/effect.rs
```diff
@@ -0,0 +1,98 @@
...
+    app: Option<Type>,
+    event: Option<Type>,
+    data: ast::Data<util::Ignored, EffectFieldReceiver>,
+}
+
+#[derive(Debug, FromField)]
+#[darling(attributes(effect))]
+pub struct EffectFieldReceiver {
+    ident: Option<Ident>,
+    operation: Option<Type>,
```
obmarg, line 21: re my "could we use an associated type for this" - you'd want to add a `ty: Type` field here and darling will use it to extract the type.

Then assuming the associated type was called `Operation` you can do a `quote! { <#ty as ::crux_core::Capability>::Operation }` to get at it.

crux_macros/src/effect.rs
```diff
@@ -0,0 +1,98 @@
...
+                (
+                    quote! {#pascal(#operation)},
+                    quote! {#snake: #pascal::new(context.with_effect(#name::#pascal))},
+                )
+            } else {
+                (
+                    quote! {#pascal},
+                    quote! {#snake: #pascal::new(context.with_effect(|_| #name::#pascal))},
+                )
+            }
```
obmarg, line 80: Doing this particular if statement will be tricky if we inferred the operation from the `Capability` impl...

Suppose one easy fix would be to make all of the capabilities take a payload, even if there's nothing of use in that payload?  So `Render` would become `Render(())` or whatever?  There's probably other ways to do it (with either a bunch of generics, more derive macros, or asking capability implementors to do more work), but that seems like the least complex option.

crux_macros/src/effect.rs
StuartHarris: yeah, true that! Borrowed code. Will fix that.

crux_http/tests/with_tester.rs
StuartHarris: Yeah, still not sure how we'd get it though, if it's not in the input token stream?

crux_macros/src/effect.rs
StuartHarris: Oh OK! Now I get it. Very clever :-) 🤯

crux_http/tests/with_tester.rs
StuartHarris: ahh just read the comment below

crux_macros/src/effect.rs
StuartHarris: interesting. Fell free to have a go at this if you can see it?

StuartHarris:
Oh, we can deduce the `Event` type by extracting the generic type param from (one of) the Capability types. Will look at that.

obmarg:
Just pushed up my suggested changes if you want to take a look @StuartHarris?  Did mean to do it on another branch, but accidentally pushed to this one without thinking 🤦 

Re-worked the unit tests a tiny bit w/ [insta](https://insta.rs) & prettyplease as well - was getting failures and it was fairly difficult to spot the probem in the `syn::to_string` output.

crux_macros/src/effect.rs
```diff
@@ -0,0 +1,235 @@
...
+                        let pascal =
+                            Ident::new(&snake.to_string().to_case(Case::Pascal), snake.span());
+                        (quote!(#snake), quote!(#pascal))
+                    })
+                    .expect("We already told darling we're on a struct with named fields");
+
+                let ty = &field.ty;
+                (
+                    quote! { #variant(<#ty as ::crux_core::capability::Capability<#event>>::Operation) },
+                    // TODO: Make this use the actual type, not #variant
```
obmarg, line 74: Spotted this - we probably want to take the name out of ty rather than using `#variant::new` here - have found in the past that you don't always get the correct output when you roundtrip through snakecase->pascalcase

obmarg:
Haven't had a chance to build & test the iOS/android apps - suspect they might need a couple of tweaks to compile.  Will try and do that later on this afternoon.

StuartHarris:
Amazing! Thanks @obmarg! Will tidy up later.

crux_macros/src/effect.rs
StuartHarris: I updated this to split the type, instead of converting from snake to pascal.

PR #36, opened 2023-01-07
Fix path support in crux_derive::Effect.

obmarg:
`split_event_type` was always looking for generic arguments in the first section of paths.  This isn't correct and meant if you provided a capability like `crux_http::Http<Event>` it'd look for generic arguments in `crux_http`, fail to find them and then complain that "all fields should be generic over the same event type"


StuartHarris (approved):
Great catch! 👍

PR #37, opened 2023-01-11
Create a "hello world" shared lib example

StuartHarris:
A simple "hello world" counter example, with tests. 
Note: this is only a shared lib (and does not include any native shells).

Also updates the `AppTester` to delegate a view method to the wrapped `App` (this is needed for the tests).


examples/hello_world/shared/src/counter.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn view(&self, model: &Self::Model) -> Self::ViewModel {
+        format!("Count is: {}", model.count)
+    }
+}
+
+#[derive(Effect)]
+#[effect(app = "Hello")]
+pub struct Capabilities {
+    render: Render<Event>,
+}
```
charypar, line 45: A really minor point, but should we declare this before it's used in the `App` impl?

charypar (approved):
Looks great!

examples/hello_world/shared/src/counter.rs
StuartHarris: yeah, good point. I've fixed the ordering of a few things

PR #38, opened 2023-01-12
Getting started docs for iOS

StuartHarris:
Flesh out the iOS part of the Getting Started section of the documentation.


docs/src/SUMMARY.md
```diff
@@ -25,5 +26,5 @@
 1. [FFI interface](./internals/uniffi.md)
    1. [Core API](./internals/core_api.md)
    1. [Type generation](./internals/typegen.md)
-1. [Serialisation](./internals/serialisation.md)
+1. [Serialization](./internals/serialization.md)
```
charypar, line 29: Is that part of the FFI, really?

StuartHarris: probably. Maybe we need a name like "bridge" although that might be a bit React Native-ish :-)

PR #41, opened 2023-01-17
Capabilities section of the guide

charypar:
- [x] Capabilities intro
- [x] Capabillity facing APIs


docs/src/guide/capabilities.md
```diff
@@ -86,16 +86,16 @@ That's it for linking the capability into our app, now we can use it in the `upd
...
 
-The other nuance to be aware of is that the capability calls return immediately. This should hopefully be relatively obvious by now, but all that's happening is effects are getting queued up to be requested from the Shell. In a way, capability calls are implicitly asynchronous (but you can't await them). It's perhaps easier to think about the capability calls as a kind of `return` statements which you can call within your update function and add effects to the batch to request.
+The other nuance to be aware of is that the capability calls return immediately. This should hopefully be relatively obvious by now, but all that's happening is effects are getting queued up to be requested from the Shell. In a way, capability calls are implicitly asynchronous (but you can't await them). It's perhaps easier to think about a capability call as a kind of `return` statement (or even better, a `yield` statement), which you can call within your update function, and which adds an effect to the batch that is sent to the shell for execution.
```
charypar, line 97: I thought about mentioning a `yield`, but both analogies have the same kind of problem that they also imply control flow, they either interrupt the code, or terminate it ealry, neither of which happens with the capability calls. Wonder if the whole comparison confuses matters, rather than clarify them...

StuartHarris: Yeah good point. I think the comparison is valid, but probably needs that clarification.

StuartHarris: or maybe, you're right and save it for when talking about building the capability itself

docs/src/guide/capability_apis.md
```diff
@@ -1,3 +1,215 @@
...
+    GetRandom(usize, usize),
+    Delay(usize),
+}
+```
+
+We now also need an output type:
+
+```rust
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct DelayOutput(Option<usize>)
```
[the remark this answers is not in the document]
charypar: You know what, you're right, I'll change it.

docs/src/guide/capability_apis.md
```diff
@@ -1,3 +1,215 @@
...
+    fn map_event<F, NewEvent>(&self, f: F) -> Self::MappedSelf<NewEvent>
+    where
+        F: Fn(NewEvent) -> Ef + Send + Sync + Copy + 'static,
+        Ef: 'static,
+        NewEvent: 'static,
+    {
+        Delay::new(self.context.map_event(f))
+    }
+}
+```
```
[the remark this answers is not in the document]
charypar: I was debating this too, I'll move it up

PR #43, opened 2023-01-20
Allow capabilities to stream data from the shell.

obmarg:
This PR contains a first pass of the shell->core streaming mechanism for capabilities.  The API that's exposed is broadly similar to what we have now, except:

1. The capability calls `stream_from_shell` instead.
2. They get a `futures::Stream` instead of a `Future`

The shell side should be similar: they get a request with a UUID, and should send any responses to that UUID.

When the capability task ends it'll drop the underlying channel which should free up most of the core-side resources involved.  An entry will remain in the `StepRegistry`, which is not ideal but it shouldn't occupy much memory.  If another message comes in for this UUID it'll be cleaned out of the StepRegistry as well.

Stuff we'll probably need to do on top of this:

1. Actually use it somewhere, it's all theoretical right now.
2. Give the shell some way to tell if a task has ended, and maybe even end it themselves (although it's possible we can handle one or both of those with the individual capability protocols).

### DIFF
--- a/crux_core/src/capability.rs
+++ b/crux_core/src/capability.rs
@@ -316,7 +316,7 @@ where
         // it's important that it is.  It forces all capabilities to
         // spawn onto the executor which keeps the ordering of effects
         // consistent with their function calls.
-        self.inner.steps.send(Step::once(operation));
+        self.inner.steps.send(Step::resolves_never(operation));
     }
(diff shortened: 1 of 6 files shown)


StuartHarris (approved):
❤️

PR #47, opened 2023-01-23
Re-order Event variants to prevent serialization errors

StuartHarris:
This is a workaround for a bug that affects clients that compile the shared library into their binary (e.g. the Yew examples).

The Events that are sent by the shell need to be ahead of any that are marked `#[serde(skip)]` otherwise the serialization won't round-trip.

This does not affect clients that use the static library (iOS), dynamic library (Android), or Wasm (e.g. NextJS), only those that compile and link a Rust library directly (e.g. Yew).

We should, at some point, provide an API for these clients so that they can skip the serialization completely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,9 +127,9 @@ checksum = "7a40729d2133846d9ed0ea60a8b9541bccddab49cd30f0715a1da672fe9a2524"
 
 [[package]]
 name = "async-trait"
-version = "0.1.61"
+version = "0.1.63"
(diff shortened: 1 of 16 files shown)


StuartHarris:
* master
  * **PR #47** <a href="https://app.graphite.dev/github/pr/redbadger/crux/47" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/redbadger/crux/47?utm_source=stack-comment).

PR #51, opened 2023-01-26
Improve termination detection in the counter CLI

charypar:
I've attempted to improve the way the CLI detects whether the work is done or whether there are still some side-effects pending. This works, but I don't particularly love the solution. I'll have a go at a slitghtly different design in a separate PR, but I think this is good to go.

I also took any counter specific bits out of the implementation of the side-effects (http and sse), so they are theoretically reusable. `futures::stream::try_unfold` is pretty awesome.

### DIFF
--- a/examples/counter/cli/src/main.rs
+++ b/examples/counter/cli/src/main.rs
@@ -1,14 +1,20 @@
 use async_std::io::ReadExt;
+use crossbeam_channel::Sender;
+use futures::{stream, TryStreamExt};
+
 use bcs::{from_bytes, to_bytes};
 use clap::Parser;
-use crossbeam_channel::Sender;
 use eyre::{bail, eyre, Result};
 use shared::{
     http::protocol::{HttpRequest, HttpResponse},
     sse::{SseRequest, SseResponse},
     Effect, Event, Request, ViewModel,
 };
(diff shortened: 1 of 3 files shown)


charypar:
* master
  * **PR #51** <a href="https://app.graphite.dev/github/pr/redbadger/crux/51" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/redbadger/crux/51?utm_source=stack-comment).

examples/counter/cli/src/main.rs
```diff
@@ -125,20 +146,14 @@ async fn http(uuid: Vec<u8>, method: Method, url: Url, tx: Sender<CoreMessage>)
 
     match status {
         200..=299 => match response.body_bytes().await {
```
StuartHarris, line 148: thinking about this, we probs shouldn't care about status here and just return it with the response regardless. Let the core decide what to do

StuartHarris (approved):
Awesome!
Also, unrelated to your changes, but just one comment.

examples/counter/cli/src/main.rs
StuartHarris: Gonna fix this on https://github.com/redbadger/crux/pull/50

PR #70, opened 2023-02-14
Handle unicode characters properly in the core and in the Web UI

charypar:
Handle unicode characters properly in the core and in the Web UI

Automerge edits operate on the level of unicode characters (above bytes,
below grapheme clusters). The browser UI works on a byte level.

This introduces a fix for the core to operate on characters, and the React
UI to convert between the native byte offsets  and unicode offsets.

The method I currently use for this is not ideal and introduces a somewhat
noticeable lag on editing long documents with emojis in them, but I'm fairly
confident we can improve this by moving the logic into the core and exposing
both types of offsets to the shells.

### DIFF
--- a/examples/notes/shared/src/app.rs
+++ b/examples/notes/shared/src/app.rs
@@ -90,7 +90,7 @@ impl App for NoteEditor {
 
                 caps.pub_sub.publish(change.bytes().to_vec());
 
-                let len = text.len();
+                let len = text.chars().count();
                 let idx = match &model.cursor {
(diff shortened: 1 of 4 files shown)


charypar:
* master
  * **PR #70** <a href="https://app.graphite.dev/github/pr/redbadger/crux/70" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/redbadger/crux/70?utm_source=stack-comment).

PR #75, opened 2023-02-16
Book section: Message interface between core and shell

StuartHarris:


charypar (approved):
Great! I've added a bunch of text and also some readme instructions for running the book locally with `admonish`.

Issue #78, opened 2023-03-04
Guide to using Crux with Flutter


duttaoindril:
There are no docs for using Crux with Flutter. Based on Crux's architecture, this should be relatively easy, since Flutter can take over UI, replacing 3 platform languages and their respective UI frameworks, allowing Crux to handle the core of the logic in the app. All Flutter would have to do is react to and trigger events. Would you have any thoughts on how to do this / get started?

StuartHarris:
This sounds like a great idea! I've not done any Flutter but would be well up for trying. If you would like to contribute an example, or add Flutter to one of the existing examples then please feel free to. Otherwise we'll definitely do this soon.

patmuk:
@StuartHarris I highly appreciate your efforts to support flutter!
I created a flutter-rust-bridge project (https://github.com/patmuk/flutter-rust-bridge_crux_style), for which I followed their guide, but made tiny changes to implement it in the crux style. It is quite fresh, so maybe this can give you a head start in exploring how crux can communicate with Flutter.

I am not deep enough into either project, but happy to help!

StuartHarris:
@patmuk thank you so much! 👍  I'll have a play with it over the weekend and let's see if we can get it in :-)

patmuk:
@StuartHarris I am glad I can help :) I had a try myself today to combine both. See the three new commits in the linked project.
But I didn't succeeded: The flutter-rust-bridge codegen (which is started with "just all" in the main directory conflicts with the crux codegen, with this error message:
> Failed to parse function argument type `&[u8]`

So, the initial commit is flutter-rust-bridge only, and works.
In the other commits I added a "cli shell" from a crux setup (which works) and the rest of the crux configuration.
While the crux codegen works, the flutter-rust-bridge doesn't.

I did not touch the flutter app - did not try to use the crux generated files here. Maybe this is the next step, and not the flutter-rust-bridge codegen.

Got too tired now to think straight though - wish you success!!

StuartHarris:
:-) Thanks @patmuk , yeah I got to the same point earlier in the week. I'm getting further with owned types (i.e. `Vec<u8>` instead of `&[u8]`). Will keep you posted

patmuk:
Hi @StuartHarris,
a small update: I refactored the git tree and made a branch for the work on integrating crux, see: https://github.com/patmuk/flutter-rust-bridge_crux_style/branches
Apologies for messing around with the history!

StuartHarris:
[...]

PR #81, opened 2023-03-12
Add Flutter to the Counter example

StuartHarris:
As per https://github.com/redbadger/crux/issues/78, this is an experiment to see what crux looks like in Flutter apps.

`serde-generate` has 'unfinished' support for Dart, as it does for TypeScript (which we're using already). The Dart types look OK, but I haven't tried them yet, so we'll see.  Still need to find a good solution for the bindings (looking at https://github.com/fzyzcjy/flutter_rust_bridge). I've also built the UI, but will push that last once I get it working.

- [ ] Add FFI for Dart
- [x] Add shared types generation for Dart
- [ ] Add Flutter UI
- [ ] Add relevant pages to the [book](https://redbadger.github.io/crux/)

### DIFF
--- a/crux_core/src/typegen.rs
+++ b/crux_core/src/typegen.rs
@@ -43,6 +43,9 @@
 //!
 //!     gen.typescript("shared_types", output_root.join("typescript"))
 //!         .expect("typescript type gen failed");
+//!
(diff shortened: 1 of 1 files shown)


StuartHarris:
Closing this for now. But will re-open and visit it again when time permits

PR #87, opened 2023-03-23
Separate bridge code from the core API

charypar:
So far, the Crux core was fused together with the serialisation code supporting the FFI. This PR makes changes to separate the serialisation code into a self-contained Bridge, which will handle

* Serialising and deserialising data for the FFI
* Parking the effects emitted by the app under a UUID so they can be resolved later once the Shell is done.

The key changes in this are in how `Effects`, `Steps`,  and `Operations` stack up. Previously, Steps contained Effects, which contained Operations. That meant that Steps did not have access to Operation type information and had to work on serialised bytes. By swapping the order, Steps are now generic over Operations and Effect bundle them into an enum for the purposed of dispatch. Each of the types has a specific job:

* Effect allows a type-checked dispatch in the Shell to the right capability implementation
* Step (now an enum too) captures the semantics of the interaction - is it a notification, a request with a single response, or a subscription with a stream of responses
* Operation captures the data needed to fulfil the interaction, and with the `Operation::Output` type, also the type of the result

There are a few consequences:
* Rust clients get a strongly typed API for the core with no serialisation in site 🎉 
* Capability context construction had to change because of the assymetry between a Context of Effect and Contex of Step<Operation> (not a big deal, just no introduce a different type for the initial construction and updated the macro)
* Sadly, this changes the Shell API once more. It now needs to be aware of steps. 
* This should, however, allow us to build foreign language interfaces to the core which hide the serialisation entirely, making Crux look native to them, and that API can now be the stable managed interface. This means that type generation and serialisation changes should no longer impact Shells. Or at least not as often.
* Not sure if AppTester is still necessary, it might be entirely possible to simply work with the core directly now

@obmarg I'd love to get your feedback on the work so far. I'm fairly confident this will work, but I may have missed something major.

## Updated on April 4th

There were more consequences in the end. 

* We ended up with two type sets for the bridge and the direct API. Both have an `Effect` and a `Request`, but they contain each other in a different order. Bridged effect is Request(Effect(Operation)), direct effect is Effect(Request(Operation)). This _may_ change later down to line, when we hide the UUID mechanics in the foreign language shells.
* The second Effect (called `EffectFfi` by default) is also macro generated. Original Effect now implements a trait which the bridge uses to serialize it. That turns the effect into its Ffi counterpart and a `ResolveBytes` which is an equivalent of Resolve that takes serialized output of the Operation. 
* A benefit of all this is that the foreign shell interface did not need to change 🎉 

# To do
- [x] re-stack Effect and Step
- [x] remove serialisation code
- [x] revisit construction of CapabilityContext
- [x] implement Bridge and StepRegistry for the new world
- [x] rename things to make more sense (hopefully)
- [x] organise modules in `crux_core` better
- [x] revisit AppTester and see how to make its API better for the new Effect/Request/Operation structure
- [x] update all Rust clients in examples, which now need to be ~step~ `Request` aware


charypar:
* master
  * **PR #87** <a href="https://app.graphite.dev/github/pr/redbadger/crux/87" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/redbadger/crux/87?utm_source=stack-comment).

crux_core/src/steps.rs
```diff
@@ -1,143 +1,88 @@
...
-    pub fn map_effect<Ef, F>(self, f: F) -> Step<Ef>
-    where
-        F: Fn(T) -> Ef + Sync + Send + Copy + 'static,
-        T: 'static,
-        Ef: 'static,
-    {
-        Step {
-            payload: f(self.payload),
-            resolve: self.resolve,
+    pub fn resolve(&mut self, output: Op::Output) -> Result<(), ResolveError> {
```
charypar, line 43: Actually needs to be `pub(crate)`, because the capability runtime needs to execute after each step is resolved - we can't let this be called from outside Crux unless we find a way to wake the runtime app right after.

There may be a way to do that with the wakers – `resolve` is basically a waker on the future that produced the step.

crux_core/src/testing.rs
```diff
@@ -115,51 +113,7 @@ impl<Ef, Ev> AppContext<Ef, Ev> {
...
-    where
-        T: serde::ser::Serialize,
-    {
-        self.context.steps.resume(
-            self.request.uuid.as_slice(),
-            &bcs::to_bytes(result).unwrap(),
-        );
-        self.context.updates()
-    }
-}
```
charypar, line 139: `TestEffect` is not really needed any more. And with that change the AppTester is starting to look _a lot_ like the Core, to the point that I wonder whether its still useful 🤔

wasnotrice:
@charypar thanks for the details. I won't say I have it 100% clearly yet, but your explanation helps a lot. This bit seems like the "crux" of the library—the tricky part—passing work to capabilities and back in a way that works across any platform.

crux_core/src/bridge/mod.rs
```diff
@@ -0,0 +1,83 @@
+mod registry;
+pub mod step_serde;
```
obmarg, line 2: Don't think this module needs to be public, since it only adds a public method onto an existing type.

```suggestion
mod step_serde;
```

crux_core/src/capability.rs
```diff
@@ -266,37 +266,92 @@ where
...
+{
+    shell_channel: Sender<Step<Op>>,
+    app_channel: Sender<Event>,
+    spawner: crate::executor::Spawner,
 }
 
-struct ContextInner<Ef, Event> {
-    steps: Sender<Step<Ef>>,
-    events: Sender<Event>,
+pub struct ProtoContext<E, Event> {
```
obmarg, line 285: Small thing, but `E` is quite ambiguous (particularly when there's already `Event`).  `Eff`, or `Effect` maybe?

crux_core/src/capability.rs
```diff
@@ -266,37 +266,92 @@ where
...
+    {
         let inner = Arc::new(ContextInner {
-            steps,
-            events,
+            shell_channel: self.shell_channel.map_input(func),
+            app_channel: self.app_channel.clone(),
+            spawner: self.spawner.clone(),
+        });
+
+        CapabilityContext { inner }
```
obmarg, line 337: Could get rid of the `#[cfg(test)]` on line 346 and just call `CapabilityContext::new(...)` here.

crux_core/src/capability.rs
```diff
@@ -401,5 +437,5 @@ mod tests {
     #[allow(dead_code)]
     enum Event {}
 
-    assert_impl_all!(CapabilityContext<Effect, Event>: Send, Sync);
+    assert_impl_all!(ProtoContext<Effect, Event>: Send, Sync);
```
obmarg, line 440: Probably still want an `assert_impl_all` test for `CapabilityContext`

crux_core/src/capability.rs
```diff
@@ -180,7 +180,7 @@ use crate::{channels::Sender, Step};
 ///     type Output = HttpResponse;
 /// }
 /// ```
-pub trait Operation: serde::Serialize + Send + 'static {
+pub trait Operation: serde::Serialize + PartialEq + Send + 'static {
```
obmarg, line 183: Why do operations require `PartialEq`?  Seems odd

crux_core/src/steps.rs
```diff
@@ -1,143 +1,107 @@
...
-
-impl Default for StepRegistry {
-    fn default() -> Self {
-        Self(Mutex::new(Store::new()))
+impl<Op> PartialEq for Step<Op>
+where
+    Op: Operation,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
```
obmarg, line 51: Guessing this is why you've made `Operation: PartialEq` elsewhere, but still wondering why `Step<Op>` needs PartialEq?

crux_core/src/steps.rs
```diff
@@ -1,143 +1,107 @@
...
-}
-
-type OnceFn = dyn FnOnce(&[u8]) + Send;
-type ManyFn = dyn Fn(&[u8]) -> Result<(), ()> + Send;
-
-pub(crate) enum Resolve {
-    Once(Box<OnceFn>),
-    Many(Box<ManyFn>),
-}
+pub struct Step<Op>(pub Op, pub Resolve<Op::Output>)
```
obmarg, line 6: Not sure about `pub Op` & `pub Resolve`.  It does destructure nice, but means adding another field is a breaking change, and when you're not destructuring `step.0` & `step.1` are not very clear.

Wondering if we'd be better making them private and exposing via functions?  Or making it a struct w/ public fields?  Not sure.

crux_macros/src/effect.rs
```diff
@@ -23,11 +23,18 @@ impl ToTokens for EffectStructReceiver {
...
+
+                (quote!(#ef_name), quote!(#ffi_ef_name))
+            }
             None => {
-                let x = Type::from_string("Effect").unwrap();
-                quote!(#x)
+                let ef_name = Type::from_string("Effect").unwrap();
+                let ffi_ef_name = Type::from_string("EffectFfi").unwrap();
+
+                (quote!(#ef_name), quote!(#ffi_ef_name))
```
obmarg, line 37: I may be missing something, but couldn't this be 
```suggestion
                (quote!(Effect), quote!(EffectFfi))
```

obmarg (commented):
nice, this has turned out pretty good.  left a handful of comments though

crux_core/src/steps.rs
```diff
@@ -1,143 +1,107 @@
...
-                match resolve(body) {
-                    Ok(_) => {}
-                    Err(_) => {
-                        // The associated task has ended so clean up our state.
-                        // We _probably_ want a way to let the shell know about this...
-                        // But I'm going to postpone figuring out how to do that.
-                        entry.remove();
-                    }
+                // The resolve has been used, turn it into a Never
+                if let Resolve::Once(f) = std::mem::replace(self, Resolve::Never) {
```
obmarg, line 73: this is quite neat

crux_core/src/bridge/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Request<Eff>
+where
+    Eff: Effect,
+{
+    pub uuid: Vec<u8>,
+    pub effect: Eff::Ffi,
+}
+
+pub struct Bridge<'de, Eff, A>
```
obmarg, line 21: I suspect you're going to have problems with this lifetime when you come to try and use the `Bridge`, but probably solvable.

charypar: Yes, I suspect so, I'm getting to that now, so I'll see what happens.

crux_core/src/steps.rs
charypar: I'll have another look at the Rust API surface in another PR, once the mechanics all work. 

I think there's a few renames and ergonomic improvements to be made before calling it a public API.

crux_core/src/steps.rs
charypar: 🙏🏻

crux_core/src/bridge/mod.rs
wasnotrice: I spent a little bit of time trying to use the `Bridge` in `hello_world` and this is true—this lifetime conflicts with the `static` lifetime we use on `Core` (or now `Bridge`) in the shell.

charypar: @wasnotrice Try again with the latest. I have the Counter example working on iOS now. Will review the new Rust public API for sensible naming and ergonomics, and then migrate all the Rust examples across.

wasnotrice: @charypar nice! I was able to get `hello_world` to compile, though the tests still need some work

crux_core/src/bridge/request_serde.rs
```diff
@@ -35,15 +35,15 @@ impl ResolveBytes {
     }
 }
 
-impl<Op> Step<Op>
+impl<Op> Request<Op>
```
wasnotrice, line 38: I really like this change, `Request` is much more descriptive.

crux_core/src/capability/mod.rs
```diff
@@ -305,13 +314,13 @@ where
     Eff: 'static,
 {
     pub(crate) fn new(
-        steps: Sender<Eff>,
-        events: Sender<Ev>,
-        spawner: crate::executor::Spawner,
+        shell_channel: Sender<Eff>,
+        app_channel: Sender<Ev>,
+        spawner: executor::Spawner,
```
wasnotrice, line 319: Channel naming is also much more descriptive

wasnotrice (commented):
@charypar the naming changes really help to clarify the code. They will make it much easier for new contributors. Great changes!

examples/counter/shared/src/app.rs
```diff
@@ -139,24 +141,27 @@ mod tests {
...
+        let response = HttpResponse {
             status: 200,
             body: serde_json::to_vec(&Counter {
                 value: 1,
                 updated_at: 1,
             })
             .unwrap(),
-        });
+        };
+        let update = app.resolve(request, response).expect("an update");
```
charypar, line 164: That pattern above in test is pretty neat!

charypar: (with `assert_let!`)

StuartHarris (approved):
I made a couple of commits with tweaks, but other than that looks f'in awesome to me!

wasnotrice (approved):
@charypar this looks really good, with a green build! 

I think this is a very strong change architecturally, separating out the serialization. And I really like the renaming changes that you made along the way. I think they will help to make crux even easier to understand.

The minimal tweaks needed to update the "notes" example highlights how well this change fits into the existing framework, and the rust shells are a lot simpler now, without having to worry about serialization.

Super work on this!

examples/hello_world/shared_types/build.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::Result;
-use crux_core::{typegen::TypeGen, Request};
-use shared::{Effect, Event, ViewModel};
+use crux_core::{bridge::Request, typegen::TypeGen};
+use shared::{EffectFfi, Event, ViewModel};
```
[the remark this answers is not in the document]
charypar: Turns out the type alias does not matter for serde-generate anyhow, but it will respect a `#[serde(rename = ...)]`, which is pretty neat!

Issue #88, opened 2023-03-23
[iOS] building to device issues - Library not loaded


romerorickyio:
I'm getting an error when building to the device. The build is successful but on boot, I get this error. 

error message:
```text
dyld[669]: Library not loaded: /Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib
  Referenced from: <_ID_> /private/var/containers/Bundle/Application/_ID_/iOS.app/iOS
  Reason: tried: '/usr/lib/system/introspection/libshared.dylib' (errno=2, not in dyld cache), '/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/private/preboot/Cryptexes/OS/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/usr/local/lib/libshared.dylib' (errno=2), '/usr/lib/libshared.dylib' (errno=2, not in dyld cache)
Library not loaded: /Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib
  Referenced from: <_ID_> /private/var/containers/Bundle/Application/_ID_/iOS.app/iOS
  Reason: tried: '/usr/lib/system/introspection/libshared.dylib' (errno=2, not in dyld cache), '/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/private/preboot/Cryptexes/OS/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/Users/romerorickyio/Developer/crux/apps/crux-submodule-app/target/aarch64-apple-ios/debug/deps/libshared.dylib' (errno=2), '/usr/local/lib/libshared.dylib' (errno=2), '/usr/lib/libshared.dylib' (errno=2, not in dyld cache)
[...]

PR #90, opened 2023-03-30
use cargo-xcode in Xcode setup and examples

StuartHarris:
We had a [problem](https://github.com/redbadger/crux/issues/88) with the iOS setup in Xcode, which meant that apps tried to load a dynamic library (rather than a static library) on physical devices.

This PR adjusts the documentation in the book, and the example iOS projects to use [`cargo-xcode`](https://crates.io/crates/cargo-xcode) to generate an Xcode project for our shared library that can be added as a sub-project in Xcode. This results in simpler setup and configuration, and links the static library into the app when deployed on an iPhone.


examples/notes/ios/Notes/ContentView.swift
```diff
@@ -2,7 +2,7 @@
 //  ContentView.swift
 //  Notes
 //
-//  Created by Viktor Charypar on 31/01/2023.
+//  Created by Stuart Harris on 30/03/2023.
```
charypar, line 5: Rude 😂

charypar (approved):
Looks good to me, but haven't tried following the instructions.

examples/notes/ios/Notes/ContentView.swift
StuartHarris: haha!

PR #92, opened 2023-04-04
Update hello_world for Bridge

wasnotrice:
@charypar here's an update to your `direct-core-api` branch that gets the hello_world example and tests building. Might save you some busy work :)

### DIFF
--- a/examples/hello_world/shared/src/counter.rs
+++ b/examples/hello_world/shared/src/counter.rs
@@ -54,7 +54,7 @@ impl App for Hello {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crux_core::{render::RenderOperation, testing::AppTester};
+    use crux_core::{render::RenderOperation, testing::AppTester, Request};
 
     #[test]
     fn renders() {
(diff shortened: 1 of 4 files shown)


examples/hello_world/shared/src/counter.rs
```diff
@@ -65,8 +65,10 @@ mod test {
 
         // Check update asked us to `Render`
         let actual_effect = &update.effects[0];
-        let expected_effect = &Effect::Render(RenderOperation);
-        assert_eq!(actual_effect, expected_effect);
+        match actual_effect {
+            Effect::Render(_) => assert!(true),
+            _ => assert!(false, "expected Render, got {:?}", actual_effect),
+        }
```
wasnotrice, line 71: This pattern feels a bit wordier then the original test, but not too bad?

charypar: I agree, it's a little annyoing. I was looking at using something like https://github.com/poignardazur/assert_let to make that a bit more succinct.

charypar (approved):
This is awesome, will definitely save some time. Thank you for the help!

Issue #93, opened 2023-04-06
Effect testing ergonomic improvements


charypar:
It would be good to provide nicer testing API for effects. `AppTester` goes quite far in making this convenient, but there are a few tasks that come up a lot in tests that are still a bit cumbersome:

* Asserting that a specific kind of effect was requested
* Finding a request for a particular effect, and returning it (failing if none is present)
* Filtering requested effects down to a particular effect type(s)

@obmarg suggested a nice solution for all that by implementing `TryInto<[operation type]>` on `Effect` in the effect derive macro, allowing convenient conversions between Effect and the expected operation with error handling.

PR #94, opened 2023-04-07
Assert effect macro

StuartHarris:
Adds a macro to assert that an update emitted an `Effect` that matches the supplied pattern.

e.g.
```rust
assert_effect!(update, Effect::Render(_));
```

Partial implementation of #93. 

**Don't merge before  #87**


obmarg (approved):
Nice

PR #96, opened 2023-04-11
Add tauri version of examples/counter app

wasnotrice:
This is a working version of the `counter` example app, using [Tauri 2.0.0 alpha](https://next--tauri.netlify.app/next/api/config/) with React. I'm opening this PR mainly to show that this combination works, and seems as clean as any of the other examples. This is all possible because of @charypar 's recent changes in #87. I'm not sure if we want to add this example to the repo. I realize that every example is a maintenance commitment, and we can't include everything.

Anyway, some notes:

- much of the code is lifted from other `counter` examples (e.g. cli), so it should be pretty familiar
- the UI is vanilla React, but the React component is copied from `examples/web-nextjs`
- so far I've only built this in the desktop dev environment
    - I haven't packaged the app yet
    - I've tried building iOS unsuccessfully. Tauri mobile is still in alpha, and it's pretty unstable at the moment. I just haven't had time to slog through the errors. This should be easy to update once the Tauri dependency settles down.

## Getting started

```
cd examples/counter
cargo build --package shared
cargo build --package shared_types
cd tauri
pnpm install
pnpm tauri dev
```

Edit: Corrected the `pnpm tauri dev` command that was causing an error with the missing `dist` folder


examples/counter/tauri/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,115 @@
+use futures::TryStreamExt;
```
wasnotrice, line 1: This is the basic shell. I extracted `http`, `sse`, and `error` modules so that it was easier to see the "essential" shell bits

examples/counter/tauri/src/App.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <section className="section has-text-centered">
+        <p className="is-size-5">Rust Core, Rust Shell (Tauri + React.js)</p>
+      </section>
+      <section className="container has-text-centered">
+        <p className="is-size-5">{state.text}</p>
+        <div className="buttons section is-centered">
+          <button
+            className="button is-primary is-warning"
+            onClick={() => {
+              invoke("decrement")
```
wasnotrice, line 46: Tauri doesn't yet have strong typing for events/commands. This string corresponds to the `decrement` function in `src/lib.rs`

examples/counter/tauri/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+            Ok::<(), Error>(())?
+        }
+    };
+
+    Ok(())
+}
+
+#[tauri::command]
+async fn increment(app_handle: tauri::AppHandle) -> () {
```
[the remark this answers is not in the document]
charypar: This was a clippy suggestion, btw. Highly recommend! <https://github.com/rust-lang/rust-clippy>

examples/counter/tauri/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+mod error;
+use error::Error;
+
+mod sse;
+use sse::sse;
+
+#[derive(Debug)]
+enum Task {
+    Event(Event),
+}
```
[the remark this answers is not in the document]
charypar: And it would simplify `handle_task` a great deal.

examples/counter/tauri/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        Effect::Render(_) => {
+            let view = core.view();
+            tauri_app.emit_all("render", view)?
+        }
+        Effect::Http(mut request) => {
+            tauri::async_runtime::spawn({
+                let core = core.clone();
+
+                async move {
+                    let response = http(&request.operation).await.unwrap();
```
charypar, line 59: Not a big one, but would be nicer to use `expect` instead of `unwrap` in these places, to aid diagnosing if they do fail for some reason.

examples/counter/tauri/src/App.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { invoke } from "@tauri-apps/api"
+import { listen, UnlistenFn } from "@tauri-apps/api/event"
+import "./App.css";
+
+type State = {
+  text: string;
+};
+
+const initialState: State = {
+  text: "",
```
charypar, line 11: Technically, this is missing the `confirmed` field, which allows the UI to style optimistic updates differently (e.g. iOS makes the text gray). Not a big deal though, not all of the examples use that field.

charypar (changes requested):
This is awesome, thanks for doing all this work @wasnotrice! I'm happy for us to take that example on, as working with Tauri is quite a common question. If the number of examples ever gets complicated, we can always drop some of the less interesting ones. Cat Facts for instance is a little extra now (and the API it uses seems to have stopped working).

The code broadly looks good to me, bar some minor comments. The main issue is it doesn't seem to run as is – there's a missing CSS import in the React part, and a `dist` folder is needed for the tauri macros to work properly (it's gitignored).

It would be nice to add a few instructions on how to run the example in the readme too, if you don't mind.

wasnotrice:
@charypar cool, I'll fix this up today!

examples/counter/tauri/src/App.tsx
wasnotrice: This is a funny one. The confirmed field does come in with updates, regardless of this type, so even though it's not here, this version still does the "pending" thing. But I agree it should be included for clarity.

wasnotrice: Funny, looks like I copied this initial state type/value from the `web-nextjs`. Should we update that example as well?I think it makes sense to keep them as similar as possible for maintenance and learning.

See https://github.com/redbadger/crux/blob/master/examples/counter/web-nextjs/pages/index.tsx#L27

examples/counter/tauri/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+/// The main entry point for Tauri
+#[cfg_attr(mobile, tauri::mobile_entry_point)]
+pub fn run() {
+    tauri::Builder::default()
+        .setup(|app| {
+            handle_task(Task::Event(Event::StartWatch), &CORE, app.handle()).unwrap();
+            Ok(())
+        })
+        .invoke_handler(tauri::generate_handler![increment, decrement])
+        .run(tauri::generate_context!())
```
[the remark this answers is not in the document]
wasnotrice: This is because I mistyped the build command. I've fixed it now in the PR description. Sorry about that!

examples/counter/tauri/src/App.tsx
wasnotrice: I haven't updated this type for now, for consistency with the other example, but it's an easy change in both places if we want to make it.

wasnotrice:
@charypar thanks for the review. I've updated the PR and fixed those issues. And now that the build command in the instructions is correct, you should actually be able to run this example 😄

examples/counter/tauri/src/App.tsx
charypar: Yea I noticed the web example has the same issue. Let's keep this as is for now, we can always fix it later on.

examples/counter/tauri/src-tauri/src/lib.rs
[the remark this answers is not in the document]
charypar: Ok worked out what's going on. This error does come up with `cargo check --features custom-protocol`. It was showing up in VSCode for me, because my rust-analyzer is set up with `--all-features`.

Seems this is somewhat common problem with the tauri template, some clarification here: <https://github.com/tauri-apps/tauri/issues/3142#issuecomment-1411918604>.

I'd be in favour of adding an empty `counter/tauri/dist/.gitkeep` (Git will warn, but it will do it).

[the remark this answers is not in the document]
wasnotrice: I agree we should figure out a way to make folks not run into this. I'll add that `.gitkeep` file.

wasnotrice:
Apologies for all the CI failures on this PR, I think we should be back in business now!

charypar:
@wasnotrice no worries at all, thanks for figuring that stuff out!

charypar (approved):

PR #97, opened 2023-04-12
Run CI on PRs as well

charypar:
Add a trigger for the build workflow to also run on PRs (including forks). 

This should be safe to do, based on <https://securitylab.github.com/research/github-actions-preventing-pwn-requests/>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,9 @@ name: build
 
 on:
   push:
+  pull_request:
+    branches:
+      - master
 
 env:
   CARGO_INCREMENTAL: 0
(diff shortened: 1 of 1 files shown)


StuartHarris (approved):

.github/workflows/build.yaml
```diff
@@ -13,6 +16,7 @@ env:
 jobs:
   build-crux:
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
```
obmarg, line 19: Never seen this before 🤔

charypar: It avoids both `push` and `pull_request` triggering on PRs local to `redbadger/crux` and building twice. I didn't want to just disable `push` (or restrict it to just master), but that is the other option.

obmarg: Interesting, nice

PR #99, opened 2023-04-13
Tidy up public API, make sure only intended things are public

charypar:


crux_core/src/bridge/mod.rs
```diff
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 use crate::Effect;
 use crate::{App, Core};
 use registry::ResolveRegistry;
+// ResolveByte is public to be accessible from crux_macros
 pub use request_serde::ResolveBytes;
```
obmarg, line 10: Is this ever intended for consumption by users, or is this private API that just has to be public for macro reasons?

Might be nice to `#[doc(hidden)]` it if the latter

charypar: I think it's _just_ macro reasons. I'll add the tag.

StuartHarris (approved):

Issue #82, opened 2023-03-13
Are there plans to support other tool chains? 


songlinshu:
Are there plans to support other tool chains? Example:x86_64-pc-windows-msvc

PR #103, opened 2023-04-21
enable windows users to build the android project

DominicD:
Hi I played around a bit with crux and I was able to build the android example with these changes.
Maybe related to #82 altough I am not even sure what effect the toolchain file actually has?

The cli example also works. The others I have not tried yet. 
I thought I will provide this and maybe a nice discussion could emerge

### DIFF
--- a/examples/counter/Android/shared/build.gradle
+++ b/examples/counter/Android/shared/build.gradle
@@ -74,24 +74,40 @@ afterEvaluate {
 task bindGen(type: Exec) {
     def outDir = "${projectDir}/src/main/java"
     workingDir "../../"
-    commandLine(
-            "sh", "-c",
-            """\
+    if (System.getProperty('os.name').toLowerCase().contains('windows')) {
+        commandLine(
+                "cmd", "/c",
+                "target\\debug\\uniffi-bindgen generate shared\\src\\shared.udl " +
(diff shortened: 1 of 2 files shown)


StuartHarris:
Thanks @DominicD ! Much appreciated. Let us know how you get on with the other examples on Windows, and if there's any more to add to this PR. 👍

DominicD:
For the android project it would maybe make sense to add a hint that the ndkVersion needs to match the one in shared/build.gradle this took me some time to figure out.

After installing wasm-pack manually (see modified Readme.md) web-nextjs worked.

I was also able to run web-yew after I found this: https://github.com/thedodd/trunk/issues/534 
I had to uninstall trunk and then install with `cargo install --locked trunk`
As this might be fixed I don't know if you want to do something at all about that?

Tauri works as well.

So except for iOS which obviously won't work everything works fine.

DominicD:
@StuartHarris Hi Stuart I don't want to annoy you. I was just wondering if there is something else you want from me or is it fine to merge it?

StuartHarris (approved):
@DominicD my sincere apologies, and thanks doing this. None of us use windows regularly, hence our inability to test and slow response! 😓 But happy to merge.

PR #104, opened 2023-04-22
Add a typegen interface for registering a type with samples

### LINKED ISSUE #102: Typegen fails when a Uuid is included
I was trying to pass a `Uuid` value in an `Event` variant, but this breaks typegen. It seems like the typegen tries to create an "empty" Uuid, which doesn't work.

I've created a minimal reproduction in the counter example app in [this commit](https://github.com/wasnotrice/crux/commit/765b90dbd44ab139d3cfb9b3919a2629c52df364) on [this branch](https://github.com/wasnotrice/crux/tree/shared_types-uuid). I don't think it matters where the Uuid appears in the types registered with typegen.

Any ideas about how to tweak the typegen to handle Uuids?

wasnotrice:
Adds a function that exposes a slightly lower-level interface than the existing `register_type()`. Fixes #102 

Before merge:

- [ ] remove `SendUuid` from the Counter app. I left that in for now as a sort of integration test, but we should remove it before merging.


crux_core/Cargo.toml
```diff
@@ -24,7 +24,7 @@ serde_json = "1.0.96"
 serde-generate = { version = "0.25.1", optional = true }
 serde-reflection = { version = "0.3.6", optional = true }
 thiserror = "1.0.40"
-uuid = { version = "1.3.1", features = ["v4", "wasm-bindgen", "js"] }
+uuid = { version = "1.3.1", features = ["v4", "wasm-bindgen", "js", "serde"] }
```
wasnotrice, line 27: This was necessary to use the `Uuid` type in the typegen tests. I was hoping not to have to change any dependencies, but maybe with this addition, we'll want to use `Uuid` directly in generated types, and we'll need it anyway.

crux_core/src/typegen.rs
```diff
@@ -8,15 +8,15 @@
 //!
 //! Ensure that you have the following line in the `Cargo.toml` of your `shared_types` library.
 //!
-//! ```rust
+//! ```rust,ignore
```
wasnotrice, line 11: Doctests in this file started failing when I added some regular tests, so they had to be ignored.

crux_core/src/typegen.rs
```diff
@@ -69,6 +69,9 @@ use std::{
     path::{Path, PathBuf},
 };
 
+// Expose from `serde_reflection` for `register_type_with_samples()`
+pub use serde_reflection::Samples;
```
wasnotrice, line 73: Users have to manage the lifetime of a `Samples` object, so we need to re-export it.

crux_core/src/typegen.rs
```diff
@@ -298,3 +343,35 @@ fn copy(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
...
+mod tests {
+    use crate::typegen::{Samples, TypeGen};
+    use serde::{Deserialize, Serialize};
+    use uuid::Uuid;
+
+    #[derive(Serialize, Deserialize)]
+    struct MyUuid(Uuid);
+
+    #[test]
+    fn test_typegen_for_uuid_without_samples() {
```
wasnotrice, line 358: I don't know if this test is ultimately useful, but I found it helpful for validating the change, and thought it might be useful in review. We can remove it before merging.

examples/counter/shared/src/app.rs
```diff
@@ -27,6 +28,7 @@ pub enum Event {
     Increment,
     Decrement,
     StartWatch,
+    SendUuid(Uuid),
```
wasnotrice, line 31: This is included as a sort of integration test, to verify that typegen works for Uuid in an example app. We can remove this before merging.

crux_core/src/typegen.rs
charypar: Do we need to expose `Samples` at all? My (possibly wrong) understanding is it is essentially an internal state of the tracer that may be useful in future type tracing – could we not hold it in our internal state, the same way we hold `Tracer`? The `State` type should probably have its variants named after the stages of processing, rather than what it holds (e.g. `Tracing`, `Generating`), to be a bit clearer.

charypar (changes requested):
Sorry about taking a while to look at this. It's looking good, but I'd like to see if we can expose fewer details to the end user. More comments at specific places in the code.

I'm not sure if that's also the reason why the doctests are failing, but it would be good to work out why that is and get them to pass, even if it means providing hidden context to them.

crux_core/src/typegen.rs
wasnotrice: Good question. I agree it would be nice to hide `Samples`. I only exposed it to satisfy the borrow checker, because it wanted the `Samples` struct to have the same lifetime as the sample data.

wasnotrice (commented):
I think I wasn't clear about the doctests. I don't think they were running before. They would need imports to be runnable. But it would be good to get them all running!

StuartHarris (approved):
Awesome! Thanks. 
I kinda want to leave something in the Counter example to demo how to use types like this. We can work it into a more realistic use case later. All good for now though.

wasnotrice:
Agreed, examples are nice. There are lots of them already in crux, so keeping that bar high is good!

PR #105, opened 2023-05-05
Add autodiscovery for type generation

charypar:
It is possible for us to find all the types used by a Crux app by starting with the top-level type implementing the `App` trait and registering the `Event`, `ViewModel` and all the capability operations and their outputs. That is the total set of types which we will be sending across the FFI boundary (including all the types used within those types).

In this way we can avoid listing all the types manually and reduce the build.rs of the `shared_type` crates to a single line (and eventually hopefully remove it altogether).

# To do
- [x] Basic scanning of the App type
- [x] Macro generated walking of the Capabilities struct
- [x] Bridge machinery types (Request, EffectFfi, etc.)
- [x] Use in all examples

### DIFF
--- a/crux_core/src/lib.rs
+++ b/crux_core/src/lib.rs
@@ -158,7 +158,7 @@ pub mod typegen;
 mod capabilities;
 mod core;
 
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 
 pub use self::{
     capabilities::*,
@@ -170,12 +170,12 @@ pub use self::{
 /// as the type argument to [`Core`] or [`Bridge`](bridge::Bridge).
 pub trait App: Default {
     /// Event, typically an `enum`, defines the actions that can be taken to update the application state.
-    type Event: Send + Deserialize<'static> + 'static;
+    type Event: Send + 'static;
     /// Model, typically a `struct` defines the internal state of the application
     type Model: Default;
     /// ViewModel, typically a `struct` describes the user interface that should be
     /// displayed to the user
-    type ViewModel: Serialize + Deserialize<'static> + 'static;
+    type ViewModel: Serialize;
     /// Capabilities, typically a `struct`, lists the capabilities used by this application
     /// Typically, Capabilities should contain at least an instance of the built-in [`Render`](crate::render::Render) capability.
     type Capabilities;
--- a/crux_core/src/lib.rs
+++ b/crux_core/src/lib.rs
@@ -158,7 +158,7 @@ pub mod typegen;
 mod capabilities;
 mod core;
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
 pub use self::{
     capabilities::*,
@@ -170,12 +170,12 @@ pub use self::{
 /// as the type argument to [`Core`] or [`Bridge`](bridge::Bridge).
 pub trait App: Default {
     /// Event, typically an `enum`, defines the actions that can be taken to update the application state.
-    type Event: Send + 'static;
+    type Event: Send + Deserialize<'static> + 'static;
     /// Model, typically a `struct` defines the internal state of the application
     type Model: Default;
     /// ViewModel, typically a `struct` describes the user interface that should be
     /// displayed to the user
-    type ViewModel: Serialize;
+    type ViewModel: Serialize + Deserialize<'static> + 'static;
     /// Capabilities, typically a `struct`, lists the capabilities used by this application
     /// Typically, Capabilities should contain at least an instance of the built-in [`Render`](crate::render::Render) capability.
     type Capabilities;
--- a/crux_core/src/typegen.rs
+++ b/crux_core/src/typegen.rs
@@ -89,6 +89,7 @@
 //! }
 //! ```
 
+use serde::Deserialize;
 use serde_generate::Encoding;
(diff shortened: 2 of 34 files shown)


charypar:
* master
  * **PR #105** <a href="https://app.graphite.dev/github/pr/redbadger/crux/105" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/redbadger/crux/105?utm_source=stack-comment).

crux_core/src/lib.rs
```diff
@@ -170,12 +170,12 @@ pub use self::{
 /// as the type argument to [`Core`] or [`Bridge`](bridge::Bridge).
 pub trait App: Default {
     /// Event, typically an `enum`, defines the actions that can be taken to update the application state.
-    type Event: Send + 'static;
+    type Event: Send + Deserialize<'static> + 'static;
```
obmarg, line 173: Might be nice to avoid adding the `Deserialize` bound here.  If you're doing a pure-rust `App` you wouldn't need to go through serde and you wouldn't need the typegen either.  This requirement could be lightly annoying for those cases.  

Could maybe put these bounds on `TypeGenregister::app` instead?

StuartHarris: Great spot @obmarg ! Fixed

crux_core/src/typegen.rs
```diff
@@ -42,54 +42,39 @@
...
+//!use uuid::Uuid;
 //!
-//! fn main() {
-//!     println!("cargo:rerun-if-changed=../shared");
+//!#[test]
+//!fn generate_types() {
+//!    let mut gen = TypeGen::new();
 //!
-//!     let mut gen = TypeGen::new();
+//!    gen.register_type::<Request<EffectFfi>>().unwrap();
```
charypar, line 56: Is this now unnecessary?

crux_macros/src/effect.rs
```diff
@@ -199,6 +224,41 @@ mod tests {
...
+                        > as ::crux_core::capability::Capability<
+                            Event,
+                        >>::Operation as ::crux_core::capability::Operation>::Output,
+                    >()?;
+                generator.register_type::<EffectFfi>()?;
+                generator.register_type::<::crux_core::bridge::Request<EffectFfi>>()?;
+                Ok(())
+            }
+        }
+        #[cfg(not(feature = "typegen"))]
```
obmarg, line 249: This looks sus to me - this code is going to be in the _users_ crate right?  Are we expecting them to declare a `typegen` feature?

obmarg: (I've only skimmed the latest iteration of the code, so apologies if I missed something)

obmarg: Never mind, I see from the examples that we are.  Seems a bit odd though tbh.  I'd have thought users would either want typegen or not want typegen, rather than wanting to enable/disable it?

charypar: Annoyingly, yes, because we need that feature in the macro generated code. Unless there's a way to make the macro itself conditional, which only just occured to me.

obmarg: > We need the macro to generate the typegen code when used from shared_types but not when used in the shared crate.

Why is that?

obmarg: I am still curious as to the why (maybe it invalidates my thinking), but :

Would a separate derive for `Export` or a parameter to the existing derive that controls whether it generates an `Export` not work?  If users are fine just having typegen everywhere they don't need to worry about feature flags, if they want to use feature flags they can `#cfg_attr` out the parameter/derive as appropriate. 

It is a tiny bit more code for users, but a macro that just doesn't work if your crate doesn't declare a particular feature flag seems a bit off to me...

StuartHarris: I like your idea to have a separate derive for `Export` (will give that a go) and to `#cfg_attr` it out. But still struggling to see how you'd avoid typegen code being deployed onto a phone without a feature.

StuartHarris: @obmarg I've split out the derive macro for implementing `Export`. Have a look and see if you think that's better (I do). The feature is still there but using `#[cfg_attr(feature = "typegen", derive(Export))]`, which I think is also better.

obmarg: Looks kind of like the GitHub outage was making us talk over each other yesterday 😂.  I definitely did not see your explanation prior to some of my comments.  But cool, I'll take a look

obmarg: Had a look - nice, definitely think it's an improvement.  Lets people optimise for this if they want, but doesn't _force_ people to.

> But still struggling to see how you'd avoid typegen code being deployed onto a phone without a feature.

If you don't call `Export` in your runtime code it'll get dropped well before it makes it to a phone, won't it?.  

Suppose there's a compile time argument for feature flags (why pay for compiling `Export` when compiling `shared` if it's not going to be used) but I'm not sure that's clear cut either: having a feature flag like this probably means you end up compiling `shared` (and maybe some of `crux`?) twice: once with the `build.rs` set of features and once without...

StuartHarris: Huh interesting. OK cool, thanks for your help with this. It's now, quite rightly, up to users to decide how they manage that, we just provide a derive macro to help. (I updated the book to explain this).

PR #106, opened 2023-05-16
use bincode instead of bcs

StuartHarris:
Use [bincode](https://github.com/bincode-org/bincode) instead of [bcs](https://github.com/zefchain/bcs).

It seems to us that bincode is faster, more widely used, better maintained, supported by `serde_generate`, supports `f32` & `f64`, and probably a safer bet for now.

- [x] use the installer from `serde_generate` to output the serialization and deserialization code for Swift and Kotlin
- [x] wrap `SharedTypes` (including `serde` and `bincode`) in a Swift package for easier use on iOS
- [x] update all examples
- [x] update docs
- [x] remove redundant serialization dependencies from capability crates


obmarg (approved):
Looks good to me.

Wonder if it would be worth marking a lot of the `novi` files as generated: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

Would make it a fair bit easier to find the important changes in giganto PRs like this.

PR #111, opened 2023-06-06
Effect testing ergonomic improvements

StuartHarris:
Fixes https://github.com/redbadger/crux/issues/93

### DIFF
--- a/crux_core/src/testing.rs
+++ b/crux_core/src/testing.rs
@@ -146,7 +146,6 @@ impl<Ef, Ev> Update<Ef, Ev> {
 ///
 /// ```
 /// # use crux_core::testing::Update;
-/// # use std::collections::VecDeque;
 /// # enum Effect { Render(String) };
 /// # enum Event { None };
 /// # let effects = vec![Effect::Render("test".to_string())].into_iter().collect();
--- a/crux_core/src/testing.rs
+++ b/crux_core/src/testing.rs
@@ -1,5 +1,5 @@
 //! Testing support for unit testing Crux apps.
-use std::{collections::VecDeque, rc::Rc};
+use std::rc::Rc;
 
 use anyhow::Result;
 
@@ -118,31 +118,22 @@ impl<Ef, Ev> AppContext<Ef, Ev> {
 #[derive(Debug)]
 pub struct Update<Ef, Ev> {
     /// Effects requested from the update run
-    pub effects: Effects<Ef>,
+    pub effects: Vec<Ef>,
     /// Events dispatched from the update run
     pub events: Vec<Ev>,
 }
 
-#[derive(Debug)]
-pub struct Effects<Ef>(VecDeque<Ef>);
-
-impl<'a, Ef> Default for Effects<&'a Ef> {
-    fn default() -> Self {
-        Self(Default::default())
+impl<Ef, Ev> Update<Ef, Ev> {
+    pub fn into_effects(self) -> impl Iterator<Item = Ef> {
+        self.effects.into_iter()
     }
-}
 
-impl<Ef> Iterator for Effects<Ef> {
-    type Item = Ef;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        self.0.pop_front()
+    pub fn effects(&self) -> impl Iterator<Item = &Ef> {
+        self.effects.iter()
     }
-}
 
-impl<Ef> std::iter::FromIterator<Ef> for Effects<Ef> {
-    fn from_iter<T: IntoIterator<Item = Ef>>(iter: T) -> Self {
-        Self(iter.into_iter().collect())
+    pub fn effects_mut(&mut self) -> impl Iterator<Item = &mut Ef> {
+        self.effects.iter_mut()
     }
 }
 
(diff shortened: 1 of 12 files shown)


obmarg:
Nice. This is a great start. 

I’m not sure exactly how best to structure it (on my phone on a train rn) but somehow generating a function that returns an Option<Effect> would be good, maybe as well as the TryFrom? It’ll pair reallly well with filter_map & find_map - for when you want to both filter and extract the inner type

obmarg:
(I’ll try and elaborate on that comment a bit tomorrow)

StuartHarris:
Thanks @obmarg. Could we just use `.ok()` to get an option from the `try_into()` result, for `filter_map` etc?
I'm playing with having the effects in the `Update` implement `Iterator`, so we can be more flexible all round, but not sure how much of an improvement it would end up being. It would be something like `update(..).effects.filter(..)` instead of `update(..).filter_effects(..)`.

StuartHarris:
@obmarg have a look when you get a moment. `Update.effects` is now a newtype that is an Iterator. I think the testing is much more intuitive now, but interested in your opinion (obvs). :-)

obmarg:
>  Could we just use .ok() to get an option from the try_into() result, for filter_map etc?

Yeah, that's definitely possible - I'm just sure there must be a way to avoid the closure altogether:

```
// This
let effect = update().effects.into_iter().find_map(Something::something)

// rather than this
let effect = update().effects.into_iter().find_map(|effect| MyEffect::try_from(effect).ok())
```

StuartHarris:
OK, had another go. Went back to `Vec<Ef>`, but with `Update.into_effects()`, `Update.effects()` and `Update.effects_mut()` for move, shared borrow and exclusive borrow iterators. Seems simpler and means we don't need to consume `Update` unless we want to.

StuartHarris:
> > Could we just use .ok() to get an option from the try_into() result, for filter_map etc?
> 
> Yeah, that's definitely possible - I'm just sure there must be a way to avoid the closure altogether:
> 
> ```
> // This
> let effect = update().effects.into_iter().find_map(Something::something)
> 
> // rather than this
> let effect = update().effects.into_iter().find_map(|effect| MyEffect::try_from(effect).ok())
> ```

@obmarg, @charypar  I added predicates for filtering and mapping to requests, so we can work with requests in the tests, rather than effects. Seems a lot easier now.

e.g.
```rust
        // An edit should trigger a timer
        let write_request = app
            .update(
                Event::EditTimer(TimerOutput::Finished { id: 1 }),
                &mut model,
            )
            .into_effects()
            .find_map(Effect::map_key_value)
            .unwrap();

        assert_let!(
            KeyValueOperation::Write(key, value),
            &write_request.operation
        );

        assert_eq!(key, "note");
        assert_eq!(value, &model.note.save());

```

crux_macros/src/effect.rs
```diff
@@ -66,12 +66,46 @@ impl ToTokens for EffectStructReceiver {
...
+                    }
+                    pub fn #map_predicate_ident(effect: #effect_name) -> Option<crux_core::Request<<#capability<#event> as ::crux_core::capability::Capability<#event>>::Operation>> {
+                        if let #effect_name::#variant(request) = effect {
+                            Some(request)
+                        } else {
+                            None
+                        }
+                    }
+                }
+            });
```
obmarg, line 108: Nice, yeah - this is the kind of thing I was thinking.

Wondering about the names though:

- `filter_{}` could maybe be `is_{}`?
- `map_x` is _almost_ a conversion method, so wonder if it should follow [the conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) and be named `into_x`?

obmarg (approved):
Awesome.  Got a couple of small comments but looks good either way

crux_core/src/testing.rs
```diff
@@ -131,15 +145,18 @@ pub struct Update<Ef, Ev> {
 /// # Example
 ///
 /// ```
-/// use crux_core::assert_effect;
+/// # use crux_core::testing::Update;
+/// # use std::collections::VecDeque;
```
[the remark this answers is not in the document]
StuartHarris: yep good catch, it was a hangover :-)

docs/src/guide/testing.md
```diff
@@ -0,0 +1,345 @@
...
+    },
+    request.operation.clone()
+);
+assert!(requests.next().is_none());
+```
+
+At this point the shell would start the timer — this is something the core can't
+do as it is a side effect — and so we need to tell the app that it was created.
+Note that resolving a request could call the app's `update()` resulting in more
+`Event`s being generated, which we need to feed back into the app.
```
charypar, line 262: This makes me wonder whether we should have an API for this common pattern - something like `resolve_and_update`, but it's not immediately clear to me what it should do, there are a few scenarios... The `resolve` can potentially trigger multiple events _or_ another effect, so it might be two separate APIs...?

Either way, we might want to look for these patterns and provide some convenience shorthands.

charypar (approved):
Lovely!

docs/src/guide/testing.md
StuartHarris: yep totally agree. For another PR though I think.

PR #113, opened 2023-06-22
Counter example in leptos

StuartHarris:
I was interested to see how Crux works with UI frameworks that are based on fine-grained reactivity. This is a basic example implemented in [Leptos](https://github.com/leptos-rs/leptos) that we can evolve/work with.


StuartHarris:
I tweaked the Yew example a little (e.g. extracting out client-side HTTP and SSE capability implementations — which are obvs identical anyway between Yew and Leptos) and what remained is remarkably similar (at least in terms of complexity and quantity of code).

I think we can probs merge this as a basic example. What would be interesting though is something a little more complex that effectively splits the view model down the component hierarchy. I can imagine another `Render` capability that could carry more granular view model updates with it (e.g. instead of just asking us to call `core.view()` and getting it all at once), but that is fun to save for later.

examples/counter/web-leptos/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Crux Counter Example - Yew</title>
```
obmarg, line 6:
```suggestion
    <title>Crux Counter Example - Leptos</title>
```

examples/counter/web-leptos/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+impl Task {
+    fn to_json(&self) -> String {
+        serde_json::to_string(self).unwrap()
+    }
+
+    fn from_json(json: &str) -> Result<Self> {
+        serde_json::from_str(json).map_err(Into::into)
+    }
+}
```
obmarg, line 27: This is a bit yuck.  Guessing you can't just use an `Rc` because you need to take it out of the `Rc` again at some point?

Also, don't we `serde(skip)` some variants - so those won't work with this example?

examples/counter/web-leptos/index.html
StuartHarris: OMG! ha. Thanks

examples/counter/web-leptos/src/main.rs
StuartHarris: Would love an alternative to this. So will have a think over the weekend. Unless you have immediate ideas. The `serde(skip)` is on events that stay inside the core, so shouldn't get involved. We just need a way to represent the `Event` that implements the traits leptos signals need (off the top of my head `Clone`, `PartialEq, `Eq`, maybe `Hash`) etc and we can't add those to `Event` right now because of the Capabilities' Events (which ironically aren't involved). Wonder if we should look at splitting the Event into internal and external, and adding all those trait impls to the external set?

obmarg: Right I see - I didn't really understand what was/wasn't going through this `Task` thing.

In that case you _could_ make this hack more elegant (but also more dangerous):

```
impl Clone for Task {
    fn clone(&self) -> Task {
        serde_json::from_str(serde_json::to_string(self).unwrap()).unwrap()
    }
}
```

obmarg: Although yeah it would be nice to find a more elegant solution.  

Splitting `Event` would work but would it be nice?  I'm not sure.

Maybe we could also look into forcing `Capabilities` events to impl the various traits?  I forget the reasons why it was difficult... some `http` type not being `Clone` was it?  can we just roll our own?  dunno

wasnotrice:
> I can imagine another `Render` capability that could carry more granular view model updates with it (e.g. instead of just asking us to call `core.view()` and getting it all at once), but that is fun to save for later.

@StuartHarris I've been thinking about this too. But I haven't run into a situation yet where `core.view()` is too slow 😄

examples/counter/web-leptos/src/main.rs
StuartHarris: Thanks @obmarg ! I implemented `Clone` on `Task` for now — you're right it's nicer. I'll have a play with implementing `Clone` on `Event` and report back.

StuartHarris: OK, my bad. It works fine implementing `Clone` on `Event`. I was also assuming it was the HTTP stopping us. But happily not.

StuartHarris:
I also took the opportunity to update the book to add a new page for the Leptos example, and update the Yew and NextJS pages to point to new minimal working code in the `hello_world` example. Apologies that it's flooded the PR with spam.

PR #119, opened 2023-07-03
Add response headers to HTTP Capability protocol

StuartHarris:
Adds ability to use HTTP headers returned from requests. 

- [x] Add headers to `HttpResponse`
- [x] Add builder pattern to avoid having to add `..Default::default()` everywhere


obmarg (approved):
Nice, this has been a long time coming.

PR #132, opened 2023-08-16
Refactor examples

StuartHarris:
A bit of a large PR, sorry. This is all around tidying up the examples ahead of Rust Global in Seattle. Covers all the web-based examples, iOS and Android to follow.

- [x] update `crux doctor` to take a `path` arg so that you can see diffs just for the specified directory
- [x] make `typegen` for `cores` in `Crux.toml` optional (caters for the situation where you have `shared` but not `shared_types` (like `hello_world`)
- [x] update the examples so that all the boilerplate code is in separate files (still to do in another PR — iOS, Android, Tauri)
- [x] make the platform sub-app in catfacts a little more self contained with ability to test at sub-app level


examples/cat_facts/web-nextjs/src/app/core.ts
```diff
@@ -0,0 +1,114 @@
...
+  BincodeDeserializer,
+} from "shared_types/bincode/mod";
+
+import { request as http } from "./http";
+
+type Response = PlatformResponse | TimeResponse | HttpResponse | KeyValueOutput;
+
+export function update(
+  event: Event,
+  callback: Dispatch<SetStateAction<ViewModel>>
```
charypar, line 35: Does this bind this core wrapper to only work with a `use_state` hook?

examples/cat_facts/web-nextjs/src/app/page.tsx
```diff
@@ -110,37 +43,37 @@ const Home: NextPage = () => {
...
         </section>
         <section className="section container has-text-centered">
-          <p>{state.fact}</p>
+          <p>{view.fact}</p>
         </section>
         <div className="buttons container is-centered">
           <button
             className="button is-primary is-danger"
-            onClick={() => dispatch(new EventVariantClear())}
+            onClick={() => update(new EventVariantClear(), setView)}
```
charypar, line 64: passing `setView` into these looks really strange to me, and easy to miss. Maybe we could add a level of indirection that links up the callback with the core wrapper?

examples/counter/cli/src/main.rs
```diff
@@ -44,85 +38,30 @@ struct Args {
...
+    core::update(&core, event, &Arc::new(tx))?;
 
-fn process_effect(
-    effect: Effect,
-    core: &Arc<Core<Effect, App>>,
-    tx: Weak<Sender<Message>>,
-) -> Result<(), ErrReport> {
-    match effect {
-        Effect::Render(_) => {
+    while let Ok(effect) = rx.recv() {
```
charypar, line 56: Oh I see why the Cat facts CLI is doing what it's doing... because there _could_ be multiple renders per effect loop settle

templates/simple_counter/web-leptos/src/main.rs
```diff
@@ -1,51 +1,41 @@
...
+                >
+                    {"Increment"}
+                </button>
+                <button class="button is-primary is-warning"
+                    on:click=move |_| set_event.update(|value| *value = Event::Decrement)
+                >
+                    {"Decrement"}
+                </button>
+            </div>
+        </section>
```
charypar, line 38: I feel like having the templates do the counter is going too far. I'd gut any counter-y things out of these and rename the `templates/simple_counter` to `templates/shells`. 

It's goin to get really confusing when you point the doctor at your app and it starts recommending some UI for the counter example 😅

charypar (changes requested):
A bunch of smaller questions, but one overriding one is - should the shell templates be counter based, or should they be _just_ the boilerplate with no actual logic in them?

templates/simple_counter/web-leptos/src/main.rs
StuartHarris: Yeah I agree! Makes much more sense. Will do that

examples/cat_facts/cli/src/core.rs
```diff
@@ -0,0 +1,140 @@
...
+use shared::{
+    key_value::{KeyValueOperation, KeyValueOutput},
+    platform::PlatformResponse,
+    time::TimeResponse,
+    CatFactCapabilities, CatFacts, Effect, Event,
+};
+
+use crate::http;
+
+pub type Core = Arc<shared::Core<Effect, CatFacts>>;
```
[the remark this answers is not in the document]
StuartHarris: I tried that, but when we spawn, because of the recursion, we need the lifetime of the instance to outlive 'static. So having them as static functions (which we pass a ref to core into) makes this much easier. And static functions on a mod called `core` is not really different to static methods on a struct called `Core`.

examples/cat_facts/cli/src/core.rs
```diff
@@ -0,0 +1,140 @@
...
+    debug!("event: {:?}", event);
+
+    for effect in core.process_event(event) {
+        process_effect(core, effect, &tx)?;
+    }
+    Ok(())
+}
+
+pub fn process_effect(core: &Core, effect: Effect, tx: &Arc<Sender<Effect>>) -> Result<()> {
+    debug!("effect: {:?}", effect);
```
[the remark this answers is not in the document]
StuartHarris: I was trying to keep all the examples so they log events and effects (useful in demos). So this will write to the terminal when `RUST_LOG=debug`.

examples/cat_facts/web-nextjs/src/app/core.ts
```diff
@@ -0,0 +1,114 @@
...
+    processEffect(uuid, effect, callback);
+  }
+}
+
+async function processEffect(
+  uuid: number[],
+  effect: Effect,
+  callback: Dispatch<SetStateAction<ViewModel>>
+) {
+  console.log("effect", effect);
```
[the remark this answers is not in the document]
StuartHarris: as above

examples/cat_facts/web-nextjs/src/app/core.ts
StuartHarris: It _does_. Let me try a callback instead

examples/cat_facts/cli/src/core.rs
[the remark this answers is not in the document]
charypar: Ah ok, let's leave it then.

examples/cat_facts/web-nextjs/src/app/page.tsx
StuartHarris: To me it reads as "update the core with this event and then call setView with any resulting view model changes". It's not so bad. 😓

templates/simple_counter/web-leptos/src/main.rs
charypar: Ok sure!

examples/cat_facts/web-nextjs/src/app/page.tsx
charypar: Hmm, don't love it, but I'm okay with it.

charypar (approved):

PR #136, opened 2023-08-29
Add an example that tests the bridge performance

charypar:
... by sending a message into the core as often as possible and rendering the UI in response, across platforms.


charypar:
This is super basic, but demonstrates the idea.

In a next PR, I'll add in a sparkline graph and a stop/start button, probably

examples/bridge_echo/Crux.toml
```diff
@@ -0,0 +1,44 @@
...
+template = "Android"
+source = "Android"
+cores = ["shared"]
+
+[shells.iOS]
+template = "iOS"
+source = "iOS"
+cores = ["shared"]
+
+[shells.web_dioxus]
```
StuartHarris, line 21: we should probs remove the shells that don't exist

examples/bridge_echo/README.md
```diff
@@ -0,0 +1,19 @@
...
+Simple counter example, with tests.
+
+### Notes:
+
+1. This example currently depends on the `pnpm` package manager when generating
+   types for TypeScript. We are currently revisiting the type generation for
+   foreign types and so this requirement will probably go, but for now, please
+   [install `pnpm`](https://pnpm.io/installation).
+
+## Simple Counter
```
StuartHarris, line 12: update readme?

StuartHarris (changes requested):
Nice! I wasn't sure whether the `withContext` calls in the Android example are needed, but that's because I don't fully understand it. I'm gonna go and learn 👀

StuartHarris (approved):

PR #139, opened 2023-09-01
iOS examples — extract core wrapper

StuartHarris:
This updates all 6 iOS examples to extract a "core" wrapper (the generic code that only changes when new capabilities are used) and to be more consistent with each other.


examples/bridge_echo/iOS/BridgePerf/ContentView.swift
```diff
@@ -1,53 +1,27 @@
...
-    case event(Event)
-}
-
-@MainActor
-class Model: ObservableObject {
-    @Published var view = ViewModel(count: 0)
-
-    init() {
+struct ContentView: View {
+    @ObservedObject var model: Core
```
charypar-nm, line 5: maybe `var core`?

examples/bridge_echo/iOS/BridgePerf/core.swift
```diff
@@ -0,0 +1,27 @@
...
+
+@MainActor
+class Core: ObservableObject {
+    @Published var view: ViewModel
+    
+    init() {
+        self.view = try! .bincodeDeserialize(input: [UInt8](BridgePerf.view()))
+    }
+    
+    func update(event: Event) {
```
charypar-nm, line 12: Do we want to be consistent on whether we pass single arguments as `_` or not?

examples/bridge_echo/iOS/BridgePerf/ContentView.swift
charypar, line 5: Maybe `var core`?

examples/bridge_echo/iOS/BridgePerf/core.swift
charypar, line 12: Do we want to be consistent in whether we pass single aguments named or as `_`?

examples/notes/iOS/Notes/ContentView.swift
```diff
@@ -19,70 +12,33 @@ extension TextCursor {
...
-    init() {
-        view = ViewModel(text: "", cursor: .position(0)) // this will get replaced by the next call
+struct ContentView: View {
+    @ObservedObject var model: Core
 
+    init(model: Core) {
+        self.model = model
         // Insert initial document for testing
-        update(event: .event(.insert("Hello from the core!")))
+        model.update(event: .insert("Hello!"))
```
charypar, line 21: Is it worth putting update into environment, like in the tap to pay example?

charypar (approved):
Looks good. A few questions, but nothing major. 

Decide what you wanna do about those, but ship it either way :shipit:

examples/notes/iOS/Notes/ContentView.swift
StuartHarris: I think it's deffo worth doing when we have child components, but otherwise probs not

PR #143, opened 2023-09-17
Generated java types general housekeeping

Trapfether:
When building the core app with an android target, java files are generated that we reference from kotlin to help facilitate our interaction with the core app. During this process, any types that were no longer being used were left in place. This PR updates the Java type generation helper clean up the shared types directory before generating the updated set of types.

### DIFF
--- a/crux_core/src/typegen.rs
+++ b/crux_core/src/typegen.rs
@@ -355,6 +355,11 @@ impl TypeGen {
 
         fs::create_dir_all(&path)?;
 
+        let package_path = package_name.replace('.', "/");
+
+        // remove any existing generated shared types, this ensures that we remove no longer used types
+        fs::remove_dir_all(path.as_ref().to_path_buf().join(&package_path)).unwrap_or(());
+
         let config = serde_generate::CodeGeneratorConfig::new(package_name.to_string())
             .with_encodings(vec![Encoding::Bincode]);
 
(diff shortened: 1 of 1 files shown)


crux_core/src/typegen.rs
```diff
@@ -355,6 +355,11 @@ impl TypeGen {
 
         fs::create_dir_all(&path)?;
 
+        let package_path = package_name.replace('.', "/");
+
+        // remove any existing generated shared types, this ensures that we remove no longer used types
+        fs::remove_dir_all(path.as_ref().to_path_buf().join(&package_path)).unwrap_or(());
```
StuartHarris, line 361: I think we can remove the `to_path_buf()` call, as `join()` is on `&path`. This will avoid an unneccesary coercion.

StuartHarris (approved):
Thank you so much for this! Much appreciated :-) Will merge (unless you fancy adding a similar thing to the Swift and TypeScript gen :-)

crux_core/src/typegen.rs
Trapfether: I removed the unnecessary coercion. Thank you for the bringing that to my attention.

Trapfether:
As far as I can tell, Swift and Typescript do not need similar treatment. The java installer puts each type in it's own file within the Java shared_types directory. The swift and typescript installers instead collate all the shared types into a single shared types file which is overwritten with each build.

StuartHarris:
Ahh perfect. Thanks!

wasnotrice:
@Trapfether nice improvement. I've definitely run into this issue, thanks!

PR #148, opened 2023-10-07
Added Send trait to NewEv generic type inside Capability derive macro

### LINKED ISSUE #147: Capability derive macro not able to send NewEv between threads
Implementing the Delay capability as described by the guide (https://redbadger.github.io/crux/guide/capability_apis.html) the compiler showed the following error:
```
error[E0277]: `NewEv` cannot be sent between threads safely
  --> shared/src/app.rs:73:10
   |
73 | #[derive(Capability)]
   |          ^^^^^^^^^^ `NewEv` cannot be sent between threads safely
   |
note: required by a bound in `Delay::<Ev>::new`
  --> shared/src/app.rs:80:19
   |
80 |     Ev: 'static + Send
   |                   ^^^^ required by this bound in `Delay::<Ev>::new`
81 | {
82 |     pub fn new(context: CapabilityContext<DelayOperation, Ev>) -> Self {
   |            --- required by a bound in this associated function
```

The function map_events should require the Send trait for the NewEv generic, but it doesn't:
```rust
            fn map_event<F, NewEv>(&self, f: F) -> Self::MappedSelf<NewEv>
            where
                F: Fn(NewEv) -> Ev + Send + Sync + Copy + 'static,
                Ev: 'static,
                NewEv: 'static,
            {
              #name::new(self.context.map_event(f))
            }
```

mmannes:
Fixes #147 by adding the Send trait to the generic type NewEv inside map_event function of the Capability macro.

In the tap_to_pay example, there is a Delay capability in which the same was done:
```rust
impl<Ev> Capability<Ev> for Delay<Ev> {
    type Operation = DelayOperation;
    type MappedSelf<MappedEv> = Delay<MappedEv>;

    fn map_event<F, NewEvent>(&self, f: F) -> Self::MappedSelf<NewEvent>
    where
        F: Fn(NewEvent) -> Ev + Send + Sync + Copy + 'static,
        Ev: 'static,
        NewEvent: 'static + Send, // <------------- Send trait was added
    {
        Delay::new(self.context.map_event(f))
    }
}
```

### DIFF
--- a/crux_macros/src/capability.rs
+++ b/crux_macros/src/capability.rs
@@ -40,7 +40,7 @@ impl ToTokens for CapabilityStructReceiver {
             where
                 F: Fn(NewEv) -> Ev + Send + Sync + Copy + 'static,
                 Ev: 'static,
-                NewEv: 'static,
+                NewEv: 'static + Send,
             {
               #name::new(self.context.map_event(f))
             }
--- a/crux_macros/src/capability.rs
+++ b/crux_macros/src/capability.rs
@@ -40,7 +40,7 @@ impl ToTokens for CapabilityStructReceiver {
             where
                 F: Fn(NewEv) -> Ev + Send + Sync + Copy + 'static,
                 Ev: 'static,
-                NewEv: 'static + Send,
+                NewEv: 'static,
             {
               #name::new(self.context.map_event(f))
             }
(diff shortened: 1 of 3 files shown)


StuartHarris:
Hey @mmannes Thank you for this! Instead of adding the bound in the macro, I added it to at the function level. It seems to make more sense here to me, not sure? Anyway I updated the Delay capability in the tap to pay example and the docs on writing a capability, have a look and LMK if you're happy, and I'll merge.

PR #154, opened 2023-11-13
Svelte simple counter example

scholtzan:
I've been tinkering around with crux and Svelte a little bit. I figured implementing the simple counter example might be useful to others as well.


StuartHarris:
@scholtzan ❤️ Thank you so much for this! I added `@parcel/config-default` for pnpm users, but otherwise it looks great! Thank you for updating the docs too. This may be the thinnest shell implementation yet — I haven't used svelte before, but will deffo explore it.

Issue #152, opened 2023-11-06
swift tools version, Package.swift


anvlkv:
Would be great to be able to override or otherwise adjust the contents of the generated package. For example I need to use a different version of swift tool.

https://github.com/redbadger/crux/blob/ebfe6affb8aa3d588a75b6f496590a60733cacf4/crux_core/src/typegen.rs#L333

I'd be happy to work on a PR for this

StuartHarris:
Hey, apologies for slow response. You would be more than welcome to submit a PR :-) Thank you

anvlkv:
Thanks!

What approach should I take?

I thought it could use a template file. Or use a custom path provided by the user.

Op do 9 nov. , Stuart Harris ***@***.***(mailto:Op do 9 nov.  , Stuart Harris <<a href=)> schreef:

charypar:
Not having thought about this very deeply, I'd go the way of attempting to allow custom code to be prepended/injected in the generated code, via an API call from the `build.rs` script in the types crate.

Not sure it covers all the necessary options. Do let us know what you find if you attempt it!

anvlkv:
Hi Victor,

Please see my related PR. https://github.com/redbadger/crux/pull/157

Please let me know if I should change it to make an API call instead

Op wo 27 dec , Viktor Charypar ***@***.***(mailto:Op wo 27 dec  , Viktor Charypar <<a href=)> schreef:

PR #157, opened 2023-11-21
allow custom typegen extensions

anvlkv:
addresses #152


anvlkv:
let me know If I can do anything about the release-plz check failing

anvlkv:
Thanks @StuartHarris!

StuartHarris:
Thank you too! 👍  This looks great. I'll try it out later and then we can get it merged.

anvlkv:
I tested this a little more with typescript, where I needed some adjustments to tsconfig.json. While my solution works, the user has to copy whole typescript directory from original typegen_extensions to avoid missing files. 

I tried a more ergonomic solution, which would check if a file from the original typegen_extensions exists in user's custom typegen_extensions. Allowing the user to replace just one file and not whole directory.
And it almost worked, only this appeared `import * as util from "https://deno.land/std@0.85.0/node/util.ts";`, not sure where it comes from but I saw a related FIXME nearby. If you could explain what's that about I'd be happy to fix the fixme and finalize this typescript adjustment, if you think it makes sense.

anvlkv:
I see the deno part is coming from serde-reflection and crux has to maintain its own copy of novi. Never mind my comment about typescript, I guess.

StuartHarris:
Thanks for this. Sorry, I'm out of the country rn. Will get on it soon though ❤️

StuartHarris (approved):
Thanks for the contribution! I've tested and it looks good. I'll tidy up the documentation and add a section to the book after this is merged.

PR #169, opened 2023-12-31
Docs: Getting started: Add small precisions about registering app type

nmrshll:
In the docs, in the section getting started, while following it I got slightly stuck on registering the app type (the Counter example just had a different name for the App (`App` instead of `Counter` in the getting started))

I'm trying to make the docs even more obvious to follow for the next person to read them.


StuartHarris (approved):
Thank you! ❤️

PR #173, opened 2024-01-08
Add example using vite

### LINKED ISSUE #172: Add an example using vite as a JS/TS build tool
For a few reasons it was a bit harder than expected adapting the examples to work with Vite (+solid-js): I hit a few problems which weren't obvious to troubleshoot.

I'm planning on adding a web target to the counter example, using Vite and SolidJs.

nmrshll:
closes #172

### DIFF
--- a/examples/simple_counter/web-solid/index.html
+++ b/examples/simple_counter/web-solid/index.html
@@ -1,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + Solid + TS</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/index.tsx"></script>
+  </body>
+</html>
--- a/examples/simple_counter/web-solid/package.json
+++ b/examples/simple_counter/web-solid/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "web-solid",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "pnpm run wasm:build && pnpm install && vite",
+    "build": "pnpm run wasm:build && pnpm install && tsc && vite build",
+    "preview": "pnpm run wasm:build && pnpm install && vite preview",
+    "wasm:build": "cd ../shared && wasm-pack build --target web"
+  },
+  "dependencies": {
+    "bulma": "^0.9.4",
+    "shared": "file:../shared/pkg",
+    "shared_types": "file:../shared_types/generated/typescript",
+    "solid-js": "^1.8.7",
+    "svelte": "^4.2.7"
+  },
+  "devDependencies": {
+    "typescript": "^5.2.2",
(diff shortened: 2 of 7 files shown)


StuartHarris:
Thank you @nmrshll for contributing this. We'd be happy to merge it once a few things are sorted.
* it looks to me like this belongs to the `simple_counter` example rather than the `counter` example, so it should be moved
* The view model property to display the count should therefore be `.count` rather than `.text` — e.g. `<button>{view.getView().count}</button>` 
* I couldn't see where the shared library is being built (maybe I'm missing something). Maybe the `package.json` scripts should look something like this?
```json
  "scripts": {
    "dev": "pnpm run wasm:build && vite",
    "build": "pnpm run wasm:build && tsc && vite build",
    "preview": "vite preview",
    "wasm:build": "cd ../shared && wasm-pack build --target web"
  },
```

examples/counter/web-solid/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
StuartHarris, line 1: I think all this belongs in `simple_counter`, rather than `counter`

examples/counter/web-solid/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "web-solid",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
```
StuartHarris, line 6: Do we need to build the `shared` lib?
```json
  "scripts": {
    "dev": "pnpm run wasm:build && vite",
    "build": "pnpm run wasm:build && tsc && vite build",
    "preview": "vite preview",
    "wasm:build": "cd ../shared && wasm-pack build --target web"
  }
```

examples/counter/web-solid/src/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { render } from "solid-js/web";
+import view, { update } from "./core";
+import { EventVariantReset, EventVariantIncrement, EventVariantDecrement } from "shared_types/types/shared_types";
+
+function App() {
+  return (
+    <>
+      <h1>Vite + Solid</h1>
+      <div class="card">
+        <button>count is {view.getView().text}</button>
```
StuartHarris, line 11:
```tsx
<button>{view.getView().count}</button>
```

StuartHarris (changes requested):

StuartHarris:
When I tested it, there was no styling. Any chance you could bring that in too? Thank you!

StuartHarris:
Hey @nmrshll this looks great now! Thank you for contributing. There's just a couple of minor things. The first is that we'll need a `.gitignore` file to exclude `node_modules`.

The second is that the `update()` function initialises the Wasm module every time it is called. It looks like this is idempotent, and so probably safe, but it does mean that there could be a slight delay (especially on poor bandwidth connections) when a button is first pressed, while the wasm is downloaded and initialised. I think a better approach would be to initialise the wasm module in `index.tsx`, instead. It looks like top-level await is supported so this should work...

```ts
import initCore from "shared";
await initCore();
```

Other than that would be happy to merge. Thank you.

mhedgpeth:
@nmrshll thanks for the PR! I'm doing some house keeping on the repo right now. Are you able to make the final changes requested so we can get this merged, or have you moved on and we should close this PR.

PR #177, opened 2024-01-19
remove uuid unused wasm-bindgen feature flag

andrewweston:
`uuid` 1.7.0 has dropped the wasm-bindgen feature flag making it incompatible with previous versions (<= 1.6.1). 

The response from `uuid` maintainers:

> Thanks for pointing this out! The `wasm-bindgen` dependency was never meant to be public. It was only a Cargo limitation that made it so. Packages should instead use the js feature...


StuartHarris:
Thank you!

StuartHarris (approved):

Issue #91, opened 2023-04-03
Async combinators for capabilities


lvrg-acc:
Sometimes, it's necessary to request multiple APIs to display something on the screen. For instance:

`cap.http(STORIES, Event::Stories)`
`cap.http(FEED, Event::Feed)`

However, we need to ensure that nothing is displayed until both results are returned. There are two main reasons for this: data consistency and rendering efficiency. Data consistency ensures that we display a complete view of the screen, without jumping between different states. Rendering efficiency is important because we don't want to re-render the screen twice.

I don't know how to address this, but probably we need something like combinators that allow us to express concurrent capabilities. An example of this would be:

`cap.join!(cap.http(STORIES), cap.http(FEED), Event::StoriesAndFeed)`

Edit: As @charypar pointed, we can avoid emitting render in the "mid-state", until both tasks are completed. On the other hand, this may lead to conditional rendering inside n calls, so perhaps it's better should be phrased as an ergonomic problem, not rendering problem.

charypar:
Yes, some level of effect combinators would be useful and we need to look at how that might work. 

Until then you can solve your situation by keeping track of the state of the requests in the app, updating your model as they arrive, and only calling the render capability when both feed and stories have completed. (The other option is having a slightly more inteligent `view` implementation that knows not to display anything until both feed and stories are available.)

I do agree that both of those solutions feel a litte bit like workarounds, and will probably get out of hand in a larger app.

As a note to our future selves: it seems like one approach could be to allow layering of capabilities, such that one capability could call another to produce the other capability's effect. Then the parent capability could use the async runtime to orchestrate across multiple effects, even across different capabilities. The flip side is that it would move some of what is arguably application logic into the capabiliites.

charypar:
This has been released in [crux_core-0.7.2](https://github.com/redbadger/crux/releases/tag/crux_core-v0.7.2).

PR #179, opened 2024-01-23
Async capability combinators

charypar:
This PR is work in progress on #91.

It turns out, surprisingly, that async capability combinators were always possible in "user land", and could be implemented as an additional capability, but the implementation is not exactly obvious, and this type of use needs to be supported by the capabilities being orchestrated across, especially `crux_http`.

This PR aims to make the use of `async` across multiple capabilities more obvious. There are a few challenges to overcome

- [x] The Orchestrate capability should probably come with crux_core
- [ ] ~Automatically derive `Clone` for capabilities and the `Capabilities` struct, to avoid various lifetime errors when moving the capabilities or the context into the async blocks~  - decided against, it would conflict with custom `Clone` implementations
- [x] Provide a `Never` operation type
- [x] Avoid producing an `Effect` variant for capabilities who's ~operations are `Never` ()~ fields are `#[effect(skip)]`
- [x] The use needs thorough documentation and an example
- [ ] Existing capabilities need to expose an async version of their APIs (#180, #181, #182)

All feedback on the ergonomics of this is welcome at this stage, but I'd like to aim to ship this soon, ideally this week.

### DIFF
--- a/crux_core/src/capabilities/compose.rs
+++ b/crux_core/src/capabilities/compose.rs
@@ -52,7 +52,7 @@ impl<Ev> Compose<Ev> {
     /// fn update(&self, event: Self::Event, model: &mut Self::Model, caps: &Self::Capabilities) {
     ///     match event {
     ///        Event::Trigger => caps.compose.spawn(|context| {
-    ///             let caps = Clone::clone(caps);
+    ///             let caps = caps.clone();
     ///
     ///             async move {
     ///                 let (result_one, result_two) =
--- a/crux_core/src/capabilities/compose.rs
+++ b/crux_core/src/capabilities/compose.rs
@@ -1,0 +1,101 @@
+//! A capability which can spawn tasks which orchestrate across other capabilities. This
+//! is useful for orchestrating a number of different effects into a single transaction.
+
+use crate::capability::{CapabilityContext, Never};
+use crate::Capability;
+use futures::Future;
+
+/// Compose capability can be used to orchestrate effects into a single transaction.
+///
+/// Example include:
+/// * Running a number of HTTP requests in parallel and waiting for all to finish
+/// * Chaining effects together, where the output of one is the input of the next and the intermediate
+///   results are not useful to the app
+/// * Implementing request timeouts by selecting across a HTTP effect and a time effect
+/// * Any arbitrary graph of effects which depend on each other (or not).
+///
+/// Note that testing composed effects is more difficult, because it is not possible to enter the effect
+/// transaction "in the middle" - only from the beginning - or to ignore some of the effects with out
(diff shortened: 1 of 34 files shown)


wasnotrice:
@charypar thanks for this, it looks great! I'm so excited to try this in my app. I think it will help a lot with maintainability. I have some places where the flow of orchestrated events is really hard to follow.

In the example, there are two concurrent events, but it could easily be two sequential events, or any other orchestration, right?

charypar:
@wasnotrice Yea, absolutely, it's just an open ended async context with access to the capability runtime, I'm pretty sure it can orchestrate an arbitrary graph of dependent effects (and emit events alon the way meantime). 

One thing to be aware off is that it is likely to be quite hard to test "sub-steps" in isolation on long effect chains. The entire outer future is one atomic flow which can only be entered into at the top.

crux_core/src/capabilities/compose.rs
```diff
@@ -0,0 +1,101 @@
...
+    ///
+    /// The argument is a closure which receives a [`ComposeContext`] which can be used to send
+    /// events to the app.
+    ///
+    /// For example:
+    /// ```
+    /// fn update(&self, event: Self::Event, model: &mut Self::Model, caps: &Self::Capabilities) {
+    ///     match event {
+    ///        Event::Trigger => caps.compose.spawn(|context| {
+    ///             let caps = Clone::clone(caps);
```
obmarg, line 55: Not a big deal, but maybe: 

```suggestion
    ///             let caps = caps.clone();
```

Don't think I've ever seen someone do `Clone::clone`, looks a bit odd

obmarg: > @StuartHarris pointed out that Clone::clone is somewhat more explicit of the intent because it doesn't rely on automatic dereferencing.

True, though I'd usually express that as `WhateverTheActualTypeIsCalled::clone`.  Although maybe that's just me.

But whatever, does not matter.

charypar: Ah yes. I might still change it to `.clone()` because it's the least likely to trip the reader up and make them wonder what is going on.

crux_macros/src/effect.rs
```diff
@@ -15,9 +15,18 @@ struct EffectStructReceiver {
 }
 
 #[derive(FromField, Debug)]
+#[darling(attributes(effect))]
 pub struct EffectFieldReceiver {
     ident: Option<Ident>,
     ty: Type,
+    skip: Option<bool>,
```
obmarg, line 22:
```suggestion
    #[darling(default)]
    skip: bool,
```

StuartHarris (approved):
yay!

Issue #182, opened 2024-01-23
Implement async API in crux_kv


charypar:
See https://github.com/redbadger/crux/pull/179 for context.

This should hopefully be a backwards compatible refactoring

PR #187, opened 2024-01-26
Async KV and Time capability APIs

### LINKED ISSUE #181: Implement async API in crux_time
See https://github.com/redbadger/crux/pull/179 for context.

This should hopefully be a backwards compatible refactoring

charypar:
Resolves #181 and #182.

Updates both KV and Time capbilities adding an async version of their API. I also switched the Time capability to use Chrono for the underlying representation - it supports timezones and serialization, so it was quite an easy choice. Open to suggestions for a better choice, if you think Chrono is too heavy a dependency for this. 

I also made `Render` capability `Clone` so it can be used along async composition (otherwise typical `Capabilities` can't be `Clone`).

### DIFF
--- a/crux_kv/tests/kv_test.rs
+++ b/crux_kv/tests/kv_test.rs
@@ -11,6 +11,7 @@ mod shared {
     pub enum Event {
         Write,
         Read,
+        ReadThenWrite,
         Set(KeyValueOutput),
     }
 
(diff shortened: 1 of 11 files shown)


crux_kv/tests/kv_test.rs
```diff
@@ -63,10 +86,12 @@ mod shared {
         }
     }
 
-    #[derive(Effect)]
+    #[derive(Effect, Clone)]
     pub struct Capabilities {
         pub key_value: KeyValue<Event>,
         pub render: Render<Event>,
+        #[effect(skip)]
+        pub compose: crux_core::compose::Compose<Event>,
```
obmarg, line 94: Not strictly related to this change, just a thought I had while reading this: it's a bit wasteful to be cloning every single capability when we want to do async things.  Not a big deal, but should we consider making `App::Capabilities` an `Arc<Capabilities>` (at least in our examples, maybe not forcing users to do so)?

charypar: Yea, this is a fair point, we probably should do something with that. 

Maybe all the examples should also be cloning just the individual capabilities needed in the async block.

obmarg: Yeah, that could also be an approach

charypar: let me at least make those changes.

charypar: Right, way less cloning now

StuartHarris (approved):

PR #189, opened 2024-01-28
Capability runtime documentation

charypar:
This is the beginning of the internals section of the docs.

I'm aiming to cover the capability runtime, Operations, Requests, and the serialisation in the bridge. The Effect derive macro and typegen are probably a separate docs PR.

## To do 
- [x] make the included code references stable using [anchors](https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file)
- [x] re-read and spellcheck

### DIFF
--- a/docs/src/internals/runtime.md
+++ b/docs/src/internals/runtime.md
@@ -1,5 +1,158 @@
 # Capability Runtime
 
-```admonish info
-Coming soon.
+In the previous sections we focused on building applications in Crux and using
+it's public APIs to do so. In this and the following chapters, we'll look at how
+the internals of Crux work, starting with the capability runtime.
+
+The capability runtime is a set of components whose job it is to facilitate the
+processing of effects to present the two perspectives we previously mentioned:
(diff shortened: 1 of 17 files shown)


docs/src/internals/runtime.md
```diff
@@ -1,5 +1,158 @@
...
+This would work, and in our case wouldn't even be that inefficient, but the
+async system is set up to avoid unnecessary polling of futures with one
+additional concept - wakers. A waker is a mechanism which can be used to signal
+the executor that something a given task is waiting on has changed, and the
+task's future should be polled, because it will be able to proceed.
+
+In our case there's a single situation which causes such a change - a result has
+arrived from the shell, for a particular effect requested earlier.
+
+So, in broad strokes, our strategy for the capability runtime is as follows:
```
charypar, line 64: Ended up not so broad, will need simplifying...

StuartHarris (approved):
love it!

Issue #191, opened 2024-01-30
The bridge should allow other other serialisers than bincode


StuartHarris:
The bridge currently assumes and only supports bincode serialisation. This is not desirable in every situation, and it should be possible to make the bridge generic over the serializer, to allow, for example, JSON serialization over the boundary when needed.

StuartHarris:
released in `crux_core` v0.7.3

PR #194, opened 2024-01-30
Allow bridge to use different serializer

charypar:
Addresses #191. The current state is pretty awful and not ready to ship, but the basic mechanics should work, I hope.

# to do
- [x] Proof of concept implementation
- [x] Test examples across platforms for backwards compatibility
- [x] Add a test with JSON serialization
- [x] Choose better names for things
- [x] Document, and add warnings about typegen not supporting other serialization than bincode (at least for now)


crux_core/src/bridge/serde.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// A basic derserializer for data across the FFI boundary. This allows
+/// you to choose a different serialization format than bincode.
+pub trait Deserializer {
+    type Error: Error;
+
+    fn deserialize<'de, T>(&self, data: &'de [u8]) -> Result<T, Self::Error>
+    where
+        T: Deserialize<'de>;
+}
```
charypar, line 23: These feel like I'm reinventing the wheel somehow. Any better suggestions?

obmarg: Why not just use serde?

crux_core/src/bridge/mod.rs
```diff
@@ -29,8 +32,7 @@ where
     Eff: Effect,
     A: App,
 {
-    core: Core<Eff, A>,
-    registry: ResolveRegistry,
+    inner: BridgeInner<Eff, A, Bincode>,
```
obmarg, line 35: What's the benefit to having `BridgeInner` vs just doing this on the Bridge?

charypar: Basically just backwards compatibility. `Inner` is a terrible name - what it should be is a `BridgeWithSerializer`.

I think we still want `Bridge` to just default to bincode like it always did, because the typegen assume is. This just opens up the option for a different serialization, if you're willing to handwrite the foreign bindings, or generate them using something else.

I'm almost tempted to hide this behind a feature flag even.

crux_core/src/bridge/serde.rs
charypar: I couldn't quite find the right thing in there when I looked. Their `Serializer` and `Deserializer` are quite complicated and I'm not entirely sure they do the whole thing we need, i.e. "A thing that turns types into `Vec<u8>`", theirs goes into the data format, if I understand right. 

It's highly likely I'm missing something though, it's been a long day. If you could point me at the right trait from their end, I would love not to define our own.

obmarg: You'd probably need to update `BridgeWithSerializer::process_event`/`handle_response` to take a `Deserializer`  & `Serializer` pair instead of taking a byte buffer and outputting a byte buffer.

Then it's just `data.deserialize(deserializer)` for the input and `data.serialize(serializer)` on the other end.

The plain bridge would just wrap its byte buffer input in one of these: https://docs.rs/bincode/1.3.3/bincode/de/struct.Deserializer.html and pass in one of these: https://docs.rs/bincode/1.3.3/bincode/struct.Serializer.html wrapping another buffer for the return value.

Not sure how much better it'd be, the current approach isn't the worst

charypar: We tried this for a while, but the problem is the resolve callbacks now become generic over the type of deserializer, which bubbles all the way up to the bridge. That's fine for the generic bridge, but for the current, bridge, we need to be able to express that concrete type of deserializer, and unfortunately, in bincode's case, the type is dependent on the data source down to its lifetime. The type is:

```rust
pub struct Deserializer<R, O: Options> {
    pub(crate) reader: R,
    options: O,
}
```

where `R`, in our case is a `SliceReader` and that's defined as:

```rust
pub struct SliceReader<'storage> {
    slice: &'storage [u8],
}
```

charypar: Don't think this mild improvement is worth changing the `Bridge` type (and I'm not 100% sure what that would even look like)

crux_core/src/bridge/request_serde.rs
```diff
@@ -4,29 +4,35 @@ use crate::{
...
 
 /// A deserializing version of Resolve
 ///
-/// ResolveBytes is a separate type because lifetime elision doesn't work
+/// ResolveSerialized is a separate type because lifetime elision doesn't work
 /// through generic type arguments. We can't create a ResolveRegistry of
 /// Resolve<&[u8]> without specifying an explicit lifetime.
 /// If you see a better way around this, please open a PR.
-pub enum ResolveBytes {
+pub enum ResolveSerialized {
```
charypar, line 17: This rename may be a step too far, as it requires a change in `crux_macros` which would otherwise not be needed, so I'm tempted to undo that.

charypar: Unless there is a way for us to somehow specify compatibility/requirements between the two

charypar:
I've tested a few examples now in iOS and webassembly (with next-js). All seems to work just fine.

StuartHarris (changes requested):
I want to have your babies ❤️  @obmarg @charypar

crux_core/src/bridge/mod.rs
```diff
@@ -93,11 +203,16 @@ where
...
-        bincode::serialize(&requests).expect("Request serialization failed.")
+        requests
+            .erased_serialize(requests_out)
+            .expect("Request serialization failed.")
     }
 
     /// Get the current state of the app's view model (serialized).
-    pub fn view(&self) -> Vec<u8> {
-        bincode::serialize(&self.core.view()).expect("View should serialize")
+    pub fn view(&self, ser: &mut dyn erased_serde::Serializer) {
```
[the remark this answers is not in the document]
obmarg: Oops - yeah, didn't mean to expose that

crux_core/src/bridge/mod.rs
```diff
@@ -39,6 +40,97 @@ where
...
+
+    /// Receive an event from the shell.
+    ///
+    /// The `event` is serialized and will be deserialized by the core before it's passed
+    /// to your app.
+    pub fn process_event(&self, event: &[u8]) -> Vec<u8>
+    where
+        A::Event: for<'a> Deserialize<'a>,
+    {
+        let options = DefaultOptions::new()
```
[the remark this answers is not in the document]
obmarg: 🤦🏻‍♂️

charypar:
Okay, that should all be ready. Added the re-export of `crux_macros` as well.

StuartHarris (approved):
yay! Well done!

PR #195, opened 2024-01-31
`crux_http` to use form of `http-types`

StuartHarris:
The published `http-types` crate uses 128bit `trace_id`, which doesn't compile for `wasm32-unknown-emscripten`.

See https://github.com/http-rs/http-types/pull/522


obmarg (commented):
😔 etc (but yeah, what can we do)

obmarg (approved):

PR #196, opened 2024-01-31
fun with erased_serde

obmarg:


crux_core/src/bridge/mod.rs
```diff
@@ -44,35 +42,66 @@ where
...
+        let mut return_buffer = vec![];
+        let mut deser = bincode::Deserializer::from_slice(
+            event,
+            DefaultOptions::new()
+                .with_fixint_encoding()
+                .allow_trailing_bytes(),
+        );
+        self.inner.process_event(
+            &mut deser,
+            &mut bincode::Serializer::new(&mut return_buffer, DefaultOptions::new()),
```
obmarg, line 66: This probably shouldn't be default options actually, but the `with_fixint_encoding` I copied from above started complaining and I couldn't be bothered figuring it out

charypar: Should they both be default options instead...? Do we have opinions on the bincode serialization?

charypar (approved):
This is great. Only thing I'd change is a few names here and there, but looks exactly like what I was trying, but the compiler wouldn't let me have it without`erased_serde`.

Good Stuff! 👏🏻

crux_core/src/bridge/mod.rs
obmarg: These are the options that `bincode::deserialize` sets up, so I'm just keeping things consistent.  Bit weird of that they don't use default options by default, but... y'know

obmarg: Although looks like I didn't push up the change where I fixed this. hmm.

obmarg: Ok, options are now consistently the same as those used by `bincode::deserialize`

PR #197, opened 2024-02-01
update `crux_time` to not use generic response type

StuartHarris:
- [x] Update Time capability to not use `DateTime<Utc>` across the boundary as this doesn't work with reflection using `serde-reflect` and `serde-generate`
- [x] Improved error handling for typegen to make this more obvious if it happens again
- [x] update Rust deps
- [x] update pnpm deps


crux_time/src/lib.rs
```diff
@@ -12,8 +12,11 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TimeRequest;
 
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TimeResponse(pub String);
+
 impl Operation for TimeRequest {
-    type Output = String;
+    type Output = TimeResponse;
```
charypar, line 19: Interesting! What forced that?

examples/cat_facts/iOS/generated/shared.swift
charypar, line 1: Did you mean to commit these?

crux_time/src/lib.rs
StuartHarris: Yeah, was struggling without a wrapper. The typegen creates types we can call serialize and deserialize on

examples/cat_facts/iOS/generated/shared.swift
StuartHarris: we have been doing. Although I guess not strictly necessary. But if we remove them, it should be in another PR, I think

charypar (approved):
🚢

PR #201, opened 2024-02-05
Better error handling for crux_http

charypar:
This should allow apps to handle HTTP errors better and provide special error handling for bad requests, unauthorized requests and other common situations that can be handled gracefully.

There are two changes
* A nicer error type which is easier to work with
* When using the event returning API, response status in the 4xx and 5xx range is considered an `Err` not an `Ok`

### DIFF
--- a/crux_http/src/error.rs
+++ b/crux_http/src/error.rs
@@ -1,22 +1,26 @@
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
-pub struct Error {
-    message: String,
-    code: Option<crate::http::StatusCode>,
+use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


crux_http/src/error.rs
```diff
@@ -1,22 +1,26 @@
...
+pub enum Error {
+    #[error("HTTP error {0}")]
+    Http(HttpError),
+    #[error("JSON serialisation error: {0}")]
+    Json(String),
+    #[error("URL parse error: {0}")]
+    Url(String),
 }
 
-impl std::fmt::Display for Error {
```
StuartHarris, line 7: The shortest lived code in Crux history. Never even released 😂

StuartHarris (approved):

PR #215, opened 2024-03-03
Update examples for latest `crux_http`

StuartHarris:
- [x] update crate deps
- [x] expose `HttpErrorHttp` from `crux_http`
- [x] add `impl From<Result<HttpResponse, HttpError>> for HttpResult` to `crux_http`
- [x] update all variants of `counter` and `cat_facts` examples to use `HttpResult`
- [x] move `cli` and `tauri` examples from `surf` to `reqwest` (`surf` now requires silly conversions since we started using a fork of `http-types` in `crux_http`)
- [x] move from `async-std` to `tokio`


StuartHarris:
Apologies for the large PR — mostly deps and example updates. The only thing that really needs reviewing are the changes to `crux_http` before we publish a new version with some small changes to help with errors

crux_http/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub use http_types::{self as http};
 
 pub use self::{
     config::Config,
-    error::HttpError,
+    error::{HttpError, HttpErrorHttp},
```
obmarg, line 29: Doesn't look like it was introduced by this PR but `HttpErrorHttp` seems like a bad name for a type....

obmarg (approved):

crux_http/src/lib.rs
charypar: We could make that into a struct variant on the `HttpError` and move the impl into `HttpError` I suppose 🤔 That's why it's called it - it's a struct for the `Http` variant of the `HttpError`

obmarg: Hmm, yes, I see.  Well 🤷 I have no real preference, just saw the name and thought "what is going on". 

Will leave yous to decide whether or not it's fine.

StuartHarris: I made it a struct variant. Seems much cleaner

PR #217, opened 2024-03-09
rustdoc typegen

StuartHarris:
This is an experiment to see what it's like to use rustdoc's JSON output to generate foreign types (in TypeScript, Swift and Kotlin) for crux shared libs.

The JSON has all the information we need, so we use it to list all the types that implement `crux_core::App`, find their associated types and follow them to their leaves, even crossing crate boundaries (e.g. to `crux_http`). After we have gathered the relevant type information, we generate the foreign types, including any serialisation / deserialisation code).

- [x] add a `codegen` subcommand to `crux_cli`
- [x] invoke cargo doc and generate JSON output
- [x] deserialise the output into [`rustdoc-types`](https://github.com/rust-lang/rust/tree/master/src/rustdoc-json-types)
- [x] gather the relevant type information
- [x] generate foreign types (or rather, for now, hook into the same `serde_generate` backend)
- [x] generate serialisation and deserialisation code (also via `serde_generate`)

All the examples now use the `#[effect]` macro, which means that they can be used with both the current typegen and this new typegen (and both should produce the exact same generated code). 

To test an example, comment out or remove the code in `shared_types/build.rs` so that the existing typegen doesn't compete. Then use the `crux_cli` binary from this PR to generate the foreign types, e.g.:

```sh
cd examples/counter
# currently requires nightly rust (`rustup install nightly`), while we work out what version of Rustdoc JSON to use
../../target/debug/crux codegen \
  --lib shared \
  --output ./shared_types/generated # this is optional, but defaults to `./shared/generated'
```


crux_cli/src/codegen/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        bail!("Could not find workspace package with path {}", args.lib)
+    };
+
+    let lib_name = lib.name();
+
+    let registry = run(lib_name, |name| load_crate(name, &manifest_paths))?;
+
+    // switch from vendored types to `serde-reflection` types
+    let registry: serde_reflection::Registry =
+        serde_json::from_slice(&serde_json::to_vec(&registry)?)?;
```
charypar, line 50: What's the plan for this in the long run? If any incompatibility happens there, the CLI will just blow up, right?

crux_cli/src/codegen/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+fn run<F>(crate_name: &str, load: F) -> Result<Registry>
+where
+    F: Fn(&str) -> Result<Crate>,
+{
+    let mut previous: HashMap<String, Crate> = HashMap::new();
+
+    let shared_lib = load(crate_name)?;
+
+    let mut filter = Filter::default();
+    filter.process(crate_name, &shared_lib)?;
```
charypar, line 79: These things need better names, IMO. 

It's not clear what sort of filtering is happening, for what purpose... Not a huge priority now, but, though I'd point it out.

LICENSE-MIT
charypar, line 1: eh?

charypar (approved):
I have not meticulously read the code, because there's so much of it 😓 I scanned through the codegen CLI, which seems to make sense, and looked for any adverse changes anywhere else - there don't seem to be any.

In the interest of being able to iterate on this better, let's morge it, but I think it will need a pass to give things more descriptive names, organise the modules a bit, clarify what's ours and what's vendored, and split various pieces into smaller building blocks (e.g. the giant logic program in `Filter` which does two if nto three jobs at once). I can only just follow what is happening at the moment 😬

examples/bridge_echo/Android/app/src/main/java/com/example/bridge_echo/Core.kt
```diff
@@ -5,11 +5,11 @@ import androidx.compose.runtime.mutableStateOf
...
-import com.example.bridge_echo.shared_types.Effect
-import com.example.bridge_echo.shared_types.Event
-import com.example.bridge_echo.shared_types.Request
-import com.example.bridge_echo.shared_types.Requests
-import com.example.bridge_echo.shared_types.ViewModel
+import com.crux.shared.types.Effect
+import com.crux.shared.types.Event
+import com.crux.shared.types.Request
+import com.crux.shared.types.Requests
+import com.crux.shared.types.ViewModel
```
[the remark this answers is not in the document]
StuartHarris: Good point. They do. This was a hangover from before I added the ability to specify a namespace in the CLI. Will change.

crux_cli/src/codegen/mod.rs
StuartHarris: Yeah. I used the types from their lib as the intermediate representation, but had to vendor them to add things like `impl Hash` for `ascent`. Then wanted to use the published types for generation, hence the ser/de.
I think in the long run we want to use our own types throughout, and obvs control the generation ourselves. This is just an intermediate step (that also ensures compatibility with what we are doing today).

crux_cli/src/codegen/mod.rs
StuartHarris: Agreed. We can refactor later I think.

LICENSE-MIT
StuartHarris: So I added this because there was some condition in some dependency I was using (can't remember what it was). So I think I can remove it now. Will do that.

PR #220, opened 2024-03-19
Docs: Suggestions from following the Getting Started guide

AngusP:
I've added a few pointers for things I got stuck on or had to check when following the 'Getting Started' guide in the book.

Some of these will be helpful for Rust devs that aren't that familiar with app development, and others hopefully avoid getting stuck where there's sensitivity to exact naming or setup.

### DIFF
--- a/docs/src/getting_started/Android/android.md
+++ b/docs/src/getting_started/Android/android.md
@@ -16,9 +16,13 @@ We want to make setting up Android Studio to work with Crux really easy. As time
 The first thing we need to do is create a new Android app in Android Studio.
 
 Open Android Studio and create a new project, for "Phone and Tablet", of type
-"Empty Compose Activity (Material3)". In this walk-through, we'll call it
-"Counter", use a minimum SDK of API 34, and save it in a directory called
-`Android`.
+"Empty Activity". In this walk-through, we'll call it "Counter"
+
+* "Name": `Counter`
+* "Package name": `com.example.simple_counter`
+* "Save Location": a directory called `Android` at the root of our monorepo
+* "Minimum SDK" `API 34`
+* "Build configuration language": `Groovy DSL (build.gradle)`
 
 Your repo's directory structure might now look something like this (some files
 elided):
@@ -35,7 +39,7 @@ elided):
 │  │        └── java
 │  │           └── com
 │  │              └── example
-│  │                 └── counter
+│  │                 └── simple_counter
 │  │                    └── MainActivity.kt
 │  ├── build.gradle
 │  ├── gradle.properties
@@ -47,7 +51,7 @@ elided):
 │  ├── build.rs
 │  ├── Cargo.toml
 │  ├── src
-│  │  ├── counter.rs
+│  │  ├── app.rs
 │  │  ├── lib.rs
 │  │  └── shared.udl
 │  └── uniffi.toml
@@ -63,9 +67,11 @@ elided):
 
 This shared Android library (`aar`) is going to wrap our shared Rust library.
 
-Under `File -> New -> New Module`, choose "Android Library" and call it
-something like `shared`. Set the "Package name" to match the one from your
-`/shared/uniffi.toml`, e.g. `com.example.counter.shared`.
+Under `File -> New -> New Module`, choose "Android Library" and give it the "Module name"
+`shared`. Set the "Package name" to match the one from your
+`/shared/uniffi.toml`, which in this example is `com.example.simple_counter.shared`.
+
+Again set the "Build configuration language" to `Groovy DSL (build.gradle)`.
 
 For more information on how to add an Android library see
 <https://developer.android.com/studio/projects/android-library>.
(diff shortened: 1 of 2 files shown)


docs/src/getting_started/Android/android.md
```diff
@@ -114,6 +123,10 @@ like this:
 
 ```
 
+```admonish warning title="Sharp edge"
+You will need to set the `ndkVersion` to one you have installed, go to "**Tools, SDK Manager, SDK Tooks**" and check "**Show Package Details**" to get your installed version, or to install the version matching `build.gradle` above.
+```
```
AngusP, line 128: I'm not sure if there's an easier way around this, like some way to say `"latest"`? The error message if the version mismatches is an unhelpful `NDK is not installed`

docs/src/getting_started/Android/android.md
```diff
@@ -16,9 +16,13 @@ We want to make setting up Android Studio to work with Crux really easy. As time
...
-"Empty Compose Activity (Material3)". In this walk-through, we'll call it
-"Counter", use a minimum SDK of API 34, and save it in a directory called
-`Android`.
+"Empty Activity". In this walk-through, we'll call it "Counter"
+
+* "Name": `Counter`
+* "Package name": `com.example.simple_counter`
+* "Save Location": a directory called `Android` at the root of our monorepo
+* "Minimum SDK" `API 34`
+* "Build configuration language": `Groovy DSL (build.gradle)`
```
AngusP, line 25: For me Android Studio was defaulting to `build.gradle.kts` build files, which ofc won't then work with the copy-pastes later.

docs/src/getting_started/Android/android.md
```diff
@@ -16,9 +16,13 @@ We want to make setting up Android Studio to work with Crux really easy. As time
 The first thing we need to do is create a new Android app in Android Studio.
 
 Open Android Studio and create a new project, for "Phone and Tablet", of type
-"Empty Compose Activity (Material3)". In this walk-through, we'll call it
-"Counter", use a minimum SDK of API 34, and save it in a directory called
-`Android`.
+"Empty Activity". In this walk-through, we'll call it "Counter"
+
+* "Name": `Counter`
+* "Package name": `com.example.simple_counter`
```
AngusP, line 22: I think this is important, e.g. if I chose a different namespace like `com.random_tutorials.simple_counter` you'd then have a problem later as the `shared` library will be in the `com.example.` namespace? (Maybe I'm wrong, I'm not an Android/Java dev 😅)

docs/src/getting_started/Android/android.md
```diff
@@ -47,7 +51,7 @@ elided):
 │  ├── build.rs
 │  ├── Cargo.toml
 │  ├── src
-│  │  ├── counter.rs
+│  │  ├── app.rs
```
AngusP, line 54: I think this is just drift -- the book has this as `app.rs` and `simple_counter` earlier -- https://redbadger.github.io/crux/getting_started/core.html#the-app and https://redbadger.github.io/crux/getting_started/core.html#the-interface-definitions

AngusP:
I did also wonder, for iOS you add the targets to `rust-toolchain.toml`

```toml
[toolchain]
channel = "stable"
components = ["rustfmt", "rustc-dev"]
targets = [
  "aarch64-apple-darwin",
  "aarch64-apple-ios",
  "aarch64-apple-ios-sim",
  "aarch64-linux-android",
  "wasm32-unknown-unknown",
  "x86_64-apple-ios"
]
profile = "minimal"
```

Why dont' you also add the Android ones, that instead we installed by manually running `rustup` later?

```sh
$ rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
```

to have this:

```toml
[toolchain]
channel = "stable"
components = ["rustfmt", "rustc-dev"]
targets = [
  "aarch64-apple-darwin",
  "aarch64-apple-ios",
  "aarch64-apple-ios-sim",
  "aarch64-linux-android",
  "armv7-linux-androideabi",
  "i686-linux-android",
  "wasm32-unknown-unknown",
  "x86_64-apple-darwin",
  "x86_64-apple-ios",
  "x86_64-linux-android",
]
profile = "minimal"
```

AngusP:
Last point I'm not sure about is your `build.gradle` examples look like this:

```gradle
dependencies {
    // our shared library
    implementation project(path: ':shared')

    def composeBom = platform('androidx.compose:compose-bom:2022.10.00')
    implementation composeBom
    androidTestImplementation composeBom

    implementation 'androidx.compose.material3:material3:1.2.0-beta01'
    ...
}
```

whereas the ones Android Studio generated for me look like

```gradle
dependencies {
    // our shared library
    implementation project(path: ":shared")

    implementation libs.androidx.core.ktx
    implementation libs.androidx.lifecycle.runtime.ktx
    implementation libs.androidx.activity.compose
   ...
}
```

It wasn't that hard to mentally diff them and add the bits needed but it's a little confusing for a noob! I'd hazard a guess it's best-practice to give fixed versions but for a tutorial that might also be a bit brittle (i.e. might just stop working in a few months/years?)

(I did also have to click something that then added `implementation libs.androidx.lifecycle.viewmodel.compose` in lieu of `implementation("androidx.lifecycle:lifecycle-viewmodel-compose:2.6.2")` but that will be because I missed it when manually diff'ing.)

docs/src/getting_started/Android/android.md
StuartHarris: yeah it is! Good catch, Thanks

docs/src/getting_started/Android/android.md
StuartHarris: No this is good. Thanks!

StuartHarris (approved):
Thank you so much for this — much appreciated.

PR #221, opened 2024-04-01
allow export derive macro to name effect

StuartHarris:
Modifies the `Export` derive macro so that it respects the `name` set by the `effect` attribute for the `Effect` derive macro. This allows the following to work, creating `MyEffectFfi` to match the `MyEffect` type.

```rust
#[cfg_attr(feature = "typegen", derive(crux_core::macros::Export))]
#[derive(crux_core::macros::Effect)]
#[effect(name = "MyEffect", app = "NoteEditor")]
pub struct Capabilities {
    timer: Timer<Event>,
    render: Render<Event>,
    pub_sub: PubSub<Event>,
    key_value: KeyValue<Event>,
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,47 +28,48 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.13"
+version = "0.6.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


charypar (approved):
Huh, I would've sworn we already had this. 🤔

charypar:
Should we add a test using this as well?

charypar (approved):

PR #223, opened 2024-04-08
Clarifications to capability runtime docs

wasnotrice:
@charypar nice job on these docs! I think they are clear and comprehensive. Here are a few suggestions for clarification, from an end-user perspective.

### DIFF
--- a/docs/src/internals/runtime.md
+++ b/docs/src/internals/runtime.md
@@ -71,14 +71,11 @@ In our case there's a single situation which causes such a change - a result has
 arrived from the shell, for a particular effect requested earlier.
 
 ```admonish warning
-This is a strong assumption Crux makes, and can lead to unexpected behaviour
-when futures with different wakers are submitted onto the runtime. Trying to do
-so will either simply not work on some platforms where the given type of waker
-(e.g. network I/O) is not available, or the futures will eventually proceed,
-but only when one of the Crux core APIs are called, because that's when the
-executor runs, as we will see below. Don't implement async work in your core code.
-Instead, implement your async work in a separate capability, always using the
-capability API (see the [capabilities](../guide/capability_apis.md) chapter).
+Always use the capability APIs provided by Crux for async work (see the
+[capabilities](../guide/capability_apis.md) chapter). Using other async APIs can
+lead to unexpected behaviour, because the resulting futures are not tied to crux
+effects. Such futures will resolve, but only after the next shell request causes
+the crux executor to execute.
 ```
 
 ## One effect's life cycle
--- a/docs/src/internals/runtime.md
+++ b/docs/src/internals/runtime.md
@@ -4,8 +4,8 @@ In the previous sections we focused on building applications in Crux and using
 its public APIs to do so. In this and the following chapters, we'll look at how
 the internals of Crux work, starting with the capability runtime.
 
-The capability runtime is a set of components whose job it is to facilitate the
-processing of effects to present the two perspectives we previously mentioned:
+The capability runtime is a set of components that processes effects, presenting
+the two perspectives we previously mentioned:
 
 - For the core, the shell appears to be a platform with a message based system
   interface
@@ -13,9 +13,9 @@ processing of effects to present the two perspectives we previously mentioned:
   with request for side-effects
 
 There are a few challenges to solve in order to facilitate this interface.
-First, each run of the `update` function can call several capabilities, and the
-requested effects are expected to be emitted together and processed
-concurrently, so the calls can't be blocking. Second, each effect requested from
+First, each run of the `update` function can call several capabilities. The
+requested effects are expected to be emitted together, and each group of effects
+will be processed concurrently, so the calls can't be blocking. Second, each effect requested from
 a capability may require multiple round-trips between the core and shell to
 conclude and we don't want to require a call to `update` per round trip, so we
 need some ability to "suspend" execution in capabilities while waiting for an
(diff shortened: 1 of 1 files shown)


docs/src/internals/runtime.md
```diff
@@ -4,18 +4,18 @@ In the previous sections we focused on building applications in Crux and using
 its public APIs to do so. In this and the following chapters, we'll look at how
 the internals of Crux work, starting with the capability runtime.
 
-The capability runtime is a set of components whose job it is to facilitate the
-processing of effects to present the two perspectives we previously mentioned:
+The capability runtime is a set of components that processes effects, presenting
+the two perspectives we previously mentioned:
```
wasnotrice, line 8: I don't think we lose any important nuance here

docs/src/internals/runtime.md
```diff
@@ -4,18 +4,18 @@ In the previous sections we focused on building applications in Crux and using
...
 - For the shell, the core appears as a stateful library responding to events
   with request for side-effects
 
 There are a few challenges to solve in order to facilitate this interface.
-First, each run of the `update` function can call several capabilities, and the
-requested effects are expected to be emitted together and processed
-concurrently, so the calls can't be blocking. Second, each effect requested from
+First, each run of the `update` function can call several capabilities. The
+requested effects are expected to be emitted together, and each group of effects
+will be processed concurrently, so the calls can't be blocking. Second, each effect requested from
```
wasnotrice, line 18: I'm trying to capture here the fact that mutliple effects emitted from a single run of `update` happen sequentially, but each "batch" or "group" of effects is processed concurrently. Maybe "batch" is a better term.

docs/src/internals/runtime.md
```diff
@@ -76,8 +76,9 @@ when futures with different wakers are submitted onto the runtime. Trying to do
 so will either simply not work on some platforms where the given type of waker
 (e.g. network I/O) is not available, or the futures will eventually proceed,
 but only when one of the Crux core APIs are called, because that's when the
-executor runs, as we will see below. Instead, implement your async work in
-a separate capability (see the [capabilities](../guide/capability_apis.md) chapter).
+executor runs, as we will see below. Don't implement async work in your core code.
+Instead, implement your async work in a separate capability, always using the
+capability API (see the [capabilities](../guide/capability_apis.md) chapter).
```
wasnotrice, line 81: I think the concept of "a future with a different waker" might be kind of abstract to the crux user. Trying to add a more concrete hint about what to do (and not to do).

I think it's important to emphasize the need to use the capability API because simply using a capability isn't enough—you can still create a future with a different waker in your capability code. But I think if you always use the provided capability API, all will be well.

obmarg: > I think the concept of "a future with a different waker" might be kind of abstract to the crux user.

Also not sure it's correct from a rust perspective tbh.  Wakers are provided by executor (so crux in this case) - it's things other than crux calling the waker that causes problems.

Not sure how better to explain it though.

wasnotrice: Yes the connection between waker and executor is a good point. In this case it's the reverse problem though—a waker provided by "not crux".

I think maybe we need to clarify the situation better for this warning to make sense. Recently I created some async work without using crux's capability API, and it seemed to "stall". 

```rust
 self.context.spawn({
    async move {
        if let Some(millis) = delay_ms {
            Delay::new(Duration::from_millis(millis)).await;
        }

        let response = context
            .request_from_shell(BleOperation::Write(id, characteristic, data))
            .await;
        context.update_app(make_event(response));
    }
});
```

(see [Zulip discussion](https://crux-community.zulipchat.com/#narrow/stream/373912-general/topic/Stalling.20future.20in.20capabilities.20.28probably.20user.20error.29/near/431553246)). The problem here is that, even though the delay async work was created in a capability, it was not created on the crux executor (i.e. using the capability APIs), so the waker was not called when the delay was over, but instead whenever the next request came in.

obmarg: > Yes the connection between waker and executor is a good point. In this case it's the reverse problem though—a waker provided by "not crux".

My point was that the waker is _always_ provided by crux.  Even in the situation you're describing.

> The problem here is that, even though the delay async work was created in a capability, it was not created on the crux executor (i.e. using the capability APIs), so the waker was not called when the delay was over, but instead whenever the next request came in.

The waker was almost certainly called when the delay was over, it's just that calling the waker alone does nothing if the executor isn't running.  Which only happens when the next request comes in.

But anyway, don't mind me - I'm just nitpicking.

obmarg: I suppose a more accurate wording would be something like:

```
This is a strong assumption Crux makes, and can lead to unexpected behaviour
when using futures that are not tied to crux effects. Such futures will resolve,
but only after the next shell request causes the crux executor to execute.
```

charypar: @obmarg yea, it's not the waker mechanics specifically, it's the source of calling it, but I'm not sure how to say that in less than 800 words 😅

obmarg: > @obmarg yea, it's not the waker mechanics specifically, it's the source of calling it, but I'm not sure how to say that in less than 800 words 😅

That's why I suggested framing it as "the executor is only running while we process requests from the shell".  Seems easier to explain and also more accurate.

wasnotrice: How about this?

```
Always use the capability APIs provided by Crux for async work (see the 
[capabilities](../guide/capability_apis.md) chapter). Using other async APIs can 
lead to unexpected behaviour, because the resulting futures are not tied to crux 
effects. Such futures will resolve, but only after the next shell request causes 
the crux executor to execute.
```

charypar (approved):
Fantastic, thank you!

PR #226, opened 2024-04-25
Instant and Duration subscriptions for `crux_time`

StuartHarris:
Updates `crux_time` capability to be more closely aligned with [`wasi-clocks`](https://github.com/WebAssembly/wasi-clocks/).

- [x] update `now` to return an `Instant`, which represents a UTC point in time (number of seconds since the epoch and number of nanoseconds since the last second).
- [x] add `subscribe_duration` and its async counterpart. The shell side of the capability responds when the requested duration has elapsed.
- [x] add `subscribe_instant` and its async counterpart. The shell side of the capability responds when the requested instant has arrived.
- [x] Add a test that demonstrates a debounce based on `subscribe_duration`

### DIFF
--- a/crux_time/src/lib.rs
+++ b/crux_time/src/lib.rs
@@ -3,62 +3,29 @@
 //! Current time (on a wall clock) is considered a side-effect (although if we were to get pedantic, it's
 //! more of a side-cause) by Crux, and has to be obtained externally. This capability provides a simple
 //! interface to do so.
-//!
-//! This is still work in progress and as such very basic.
 
-use chrono::{DateTime, Utc};
+pub mod duration;
+pub mod error;
+pub mod instant;
+
+pub use duration::Duration;
+pub use error::TimeError;
+pub use instant::Instant;
+
 use serde::{Deserialize, Serialize};
-use thiserror::Error;
 
 use crux_core::capability::{CapabilityContext, Operation};
 
-/// Error type for time operations
-#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Error)]
-pub enum TimeError {
-    #[error("invalid time")]
-    InvalidTime,
-}
-
-/// Represents a point in time:
-///
-/// - seconds: number of seconds since the Unix epoch (1970-01-01T00:00:00Z)
-/// - sub_nanos: number of nanoseconds since the last second
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct Instant {
-    pub seconds: u64,
-    pub sub_nanos: u32,
-}
-
-impl TryFrom<Instant> for DateTime<Utc> {
-    type Error = TimeError;
-
-    fn try_from(time: Instant) -> Result<Self, Self::Error> {
-        let seconds = i64::try_from(time.seconds).map_err(|_| TimeError::InvalidTime)?;
-        DateTime::<Utc>::from_timestamp(seconds, time.sub_nanos).ok_or(TimeError::InvalidTime)
-    }
-}
-
-impl TryFrom<DateTime<Utc>> for Instant {
-    type Error = TimeError;
-
-    fn try_from(time: DateTime<Utc>) -> Result<Self, Self::Error> {
-        let seconds = time.timestamp();
-        let sub_nanos = time.timestamp_subsec_nanos();
-        Ok(Instant {
-            seconds: seconds.try_into().map_err(|_| TimeError::InvalidTime)?,
-            sub_nanos,
-        })
-    }
-}
-
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub enum TimeRequest {
     Now,
     SubscribeInstant(Instant),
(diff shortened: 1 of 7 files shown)


crux_time/src/lib.rs
```diff
@@ -5,24 +5,74 @@
...
+    #[error("invalid time")]
+    InvalidTime,
+}
+
+/// Represents a point in time:
+///
+/// - seconds: number of seconds since the Unix epoch (1970-01-01T00:00:00Z)
+/// - sub_nanos: number of nanoseconds since the last second
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Instant {
```
wasnotrice, line 27: I love the change to using `Instant` vs `chrono::DateTime`, since I'm using `time` in my app. It would be cool if `chrono` and `time` conversions were behind features, so you could opt-in to the ones you need and not pull in extra dependencies.

crux_time/src/lib.rs
```diff
@@ -44,33 +94,73 @@ where
...
+    /// Subscribe to receive a notification when the specified [`Instant`] has arrived.
+    /// This is an async call to use with [`crux_core::compose::Compose`].
+    pub async fn subscribe_instant_async(&self, instant: Instant) -> TimeResponse {
+        self.context
+            .request_from_shell(TimeRequest::SubscribeInstant(instant))
+            .await
+    }
+
+    /// Subscribe to receive a notification when the specified duration has elapsed.
+    pub fn subscribe_duration<F>(&self, duration: u64, callback: F)
```
wasnotrice, line 143: It might be nice to include a `Duration` type that supports seconds, milliseconds, etc., and/or some helpers to generate nanosecond values from other units, to avoid conversion mistakes.

wasnotrice (commented):
@StuartHarris great update to `time`!

crux_time/src/lib.rs
StuartHarris: yeah, good idea. Will do that

crux_time/src/lib.rs
```diff
@@ -44,33 +61,135 @@ where
...
-            .request_from_shell(TimeRequest)
+            .request_from_shell(TimeRequest::SubscribeInstant(instant))
             .await
-            .0
-            .parse::<DateTime<Utc>>()
-            .expect("invalid time format")
+    }
+
+    /// Subscribe to receive a notification when the specified duration has elapsed.
+    pub fn subscribe_duration<F>(&self, duration: Duration, callback: F)
```
obmarg, line 110: Wondering about the name subscribe here.  Maybe I'm just far too down the GraphQL rabbit hole (feel free to call me out if so 😅) but "subscribe" sort of suggests a recurring thing to me.

Seems less ambiguous for `subscribe_instant` because an instant can obviously only happen once - but with duration I'd have to read the docs to see whether I'm going to get an update every `duration` or not.

crux_time/src/lib.rs
```diff
@@ -44,33 +61,135 @@ where
...
+        });
+    }
+
+    /// Subscribe to receive a notification when the specified duration has elapsed.
+    /// This is an async call to use with [`crux_core::compose::Compose`].
+    pub async fn subscribe_duration_async(&self, duration: Duration) -> TimeResponse {
+        self.context
+            .request_from_shell(TimeRequest::SubscribeDuration(duration))
+            .await
+    }
```
obmarg, line 132: This function is often called sleep elsewhere - I wonder if borrowing that name here would be a good idea?

crux_time/src/lib.rs
obmarg: Maybe `notify_after`/`notify_after_duration` or something along those lines?  It's probably not a big deal though if you prefer the current names.

crux_time/src/lib.rs
```diff
@@ -44,33 +61,135 @@ where
...
+    /// This is an async call to use with [`crux_core::compose::Compose`].
+    pub async fn subscribe_instant_async(&self, instant: Instant) -> TimeResponse {
         self.context
-            .request_from_shell(TimeRequest)
+            .request_from_shell(TimeRequest::SubscribeInstant(instant))
             .await
-            .0
-            .parse::<DateTime<Utc>>()
-            .expect("invalid time format")
+    }
```
obmarg, line 107: If you're into the `sleep` suggestion I made elsewhere, maybe this is `sleep_until`?

obmarg (approved):
Nice, I just have a few naming nitpicks

obmarg:
Though I just skimmed that spec you linked, and have realised that's probably where you got the names.  Never mind then

crux_time/src/lib.rs
StuartHarris: Ha! I don't like those names either. I wondered if "wait_for" and "wait_until" work? Although not sure we're exactly waiting. Notify might be a better root

obmarg: The async ones are kind of waiting I guess?  Though yeah, maybe it makes less sense for the callback ones...

StuartHarris: `notify_`* wins!

PR #227, opened 2024-04-30
Evolve `crux_kv`

StuartHarris:
- [x] add delete
- [ ] improve error handling
- [ ] inch towards [`wasi-keyvalue`](https://github.com/WebAssembly/wasi-keyvalue/blob/main/wit/store.wit)
- [ ] improve tests

### DIFF
--- a/crux_kv/src/error.rs
+++ b/crux_kv/src/error.rs
@@ -1,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+pub type KeyValueResult<T> = Result<T, KeyValueError>;
+
+/// Error type for KeyValue operations
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Error)]
+#[serde(rename_all = "camelCase")]
+pub enum KeyValueError {
+    #[error("IO error: {message}")]
+    Io { message: String },
+    #[error("timeout")]
+    Timeout,
+    #[error("other error: {message}")]
+    OtherError { message: String },
+}
--- a/crux_kv/src/lib.rs
+++ b/crux_kv/src/lib.rs
@@ -4,25 +4,35 @@
 //! persist the data using platform native capabilities (e.g. disk or web localStorage)
 //!
 //! This is still work in progress and extremely basic.
+pub mod error;
+
 use crux_core::capability::{CapabilityContext, Operation};
 use crux_core::macros::Capability;
+use error::KeyValueResult;
 use serde::{Deserialize, Serialize};
 
 /// Supported operations
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
 pub enum KeyValueOperation {
     /// Read bytes stored under a key
-    Read(String),
+    Get { key: String },
     /// Write bytes under a key
-    Write(String, Vec<u8>),
+    Set { key: String, value: Vec<u8> },
+    /// Remove a key and its value
+    Delete { key: String },
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
 pub enum KeyValueOutput {
-    // TODO: Add support for errors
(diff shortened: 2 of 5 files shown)


crux_kv/src/error.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Error type for KeyValue operations
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Error)]
+#[serde(rename_all = "camelCase")]
+pub enum KeyValueError {
+    #[error("IO error: {message}")]
+    Io { message: String },
+    #[error("timeout")]
+    Timeout,
+    #[error("other error: {message}")]
+    OtherError { message: String },
```
charypar, line 15:
```suggestion
    Other { message: String },
```

Maybe a better name?

crux_kv/src/lib.rs
```diff
@@ -2,27 +2,43 @@
...
+    },
+    Set {
+        result: KeyValueResult<()>,
+    },
+    Delete {
+        result: KeyValueResult<()>,
+    },
+    Exists {
+        result: KeyValueResult<bool>,
+    },
```
charypar, line 41: Does the codegen handle that type correctly despite the generics...?

crux_kv/src/lib.rs
```diff
@@ -51,33 +67,34 @@ where
...
             .await
     }
 
     /// Set `key` to be the provided `value`. Typically the bytes would be
     /// a value serialized/deserialized by the app.
     ///
-    /// Will dispatch the event with a `KeyValueOutput::Write(bool)` as payload
-    pub fn write<F>(&self, key: &str, value: Vec<u8>, make_event: F)
+    /// Will dispatch the event with a `KeyValueOutput::Set { result: KeyValueResult<()> }` as payload
+    pub fn set<F>(&self, key: &str, value: Vec<u8>, make_event: F)
```
charypar, line 97: Can this be written in terms of `set_async` or am I missing some important detail?

charypar: Same for the other sync APIs

crux_kv/src/lib.rs
StuartHarris: Good catch. Thanks. No it doesn't! I replaced them with enums.

crux_kv/src/lib.rs
StuartHarris: yep, I did that, and it's much better. Also some assertions that the response types are correct

crux_kv/src/lib.rs
```diff
@@ -2,36 +2,68 @@
...
+/// supported across the FFI boundary, when using the builtin typegen.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum KeyValueReadResult {
+    /// the value stored under the key, may be empty
+    Data { value: Vec<u8> },
+    /// whether the key exists or not
+    Exists { value: bool },
+    /// an error occurred
+    Err { error: KeyValueError },
+}
```
charypar, line 38: I suppose you could merge the Read and Write result values into a single enum, and then it would 

```rust
/// this is the capability's output
KeyValueResult {
     Ok(KeyValueResponse),
     Err(KeyValueError),
}

KeyValueResponse {
    Data { value: ... },
    Updated { previous: ... }, // assume this handles Set and Delete
    Exists { present: ... },
}
```

not sure if that's better... 🤔 Less nesting, more invalid states, but we already do asserts on the response...?

charypar (approved):
Looks good. Suggested a possible alternative for the output types, but not that sure whether it's better than what you've got.

crux_kv/src/lib.rs
StuartHarris: I like this better. Will update

StuartHarris: see https://github.com/redbadger/crux/pull/227/commits/3faa22baa18e506af63587ad834ac8edf6eead42

PR #228, opened 2024-05-02
Change all serialisable enums across Crux to be "internally tagged"

StuartHarris:
Now we introduced the ability to serialise as JSON (in #194), we want to make some ease of life improvements to JSON serialisation formats.

From our experience, handling JSON serialisation and deserialisation in TS, Kotlin and Swift is much easier when enums are internally tagged. 

For example it avoids having to look ahead in enums with different variant types in order to determine which variant to use.

Note: this will not affect `bincode` serialisation, but is a breaking change for JSON.

- [ ] Crux macros
- [ ] Every serialisable enum across all crates

### DIFF
--- a/crux_time/src/lib.rs
+++ b/crux_time/src/lib.rs
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 use crux_core::capability::{CapabilityContext, Operation};
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "camelCase", tag = "type")]
 pub enum TimeRequest {
     Now,
     NotifyAt(Instant),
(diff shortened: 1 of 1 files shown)


StuartHarris:
for completeness, this is the error from `serde_generate`
```
  Error: type tracing failed: Not supported: deserialize_any:

  An unsupported callback was called during (de)serialization tracing. In practice, this happens when an
  unsupported Serde attribute is used. Attributes specific to self-describing formats (JSON, YAML, TOML)
  are generally not supported. This includes: `#[serde(flatten)]`, `#[serde(tag = "type")]`,
  `#[serde(tag = "t", content = "c")]`, and `#[serde(untagged)]`.

  To fix this, avoid unsupported Serde attributes or use custom (de)serialize implementations with different
  behaviors depending on the Serde callback `(De)Serializer::is_human_readable()`.
```

PR #229, opened 2024-05-09
Relax callback bounds

adwhit:
I noticed bounds on several of the traits are overly restrictive - they receive callbacks of type `Fn` when type `FnOnce` (or `FnMut`) would work. In general it is preferable to use the less restrictive type in APIs. 

According to [the docs](https://doc.rust-lang.org/std/ops/trait.FnOnce.html), "Since both `Fn` and `FnMut` are subtraits of `FnOnce`, any instance of `Fn` or `FnMut` can be used where a FnOnce is expected." - therefore relaxing the trait bounds is non-breaking.


charypar (approved):
Thanks for doing this! ❤️ Looks good to me, but would like at least one more ✅. 

Can't quite see whether we might want to restrict them again later or why we would. All those `Fn`s are generally meant to be the event variant constructors which are used as a form of callback, so most of them are indeed called once. 

The ones used in subscriptions get cloned because they potentially move between threads anyway (would be nice to avoid that, but I think I once tried and couldn't work it out)

crux_time/tests/time_test.rs
```diff
@@ -106,10 +106,9 @@ mod shared {
...
     /// Helper to create an event with additional user info captured
     /// this is effectively partially applying the event constructor
-    pub fn event_with_user_info<E, F, U, T>(user_info: U, make_event: F) -> impl Fn(T) -> E
+    pub fn event_with_user_info<E, F, U, T>(user_info: U, make_event: F) -> impl FnOnce(T) -> E
     where
-        F: Fn(U, T) -> E,
-        U: Clone,
+        F: FnOnce(U, T) -> E,
     {
         move |response| make_event(user_info.clone(), response)
```
StuartHarris, line 113: this `.clone()` now needs to go

examples/counter/shared/src/capabilities/sse.rs
```diff
@@ -35,7 +35,7 @@ where
 
     pub fn get_json<F, T>(&self, url: impl AsRef<str>, make_event: F)
     where
-        F: Fn(T) -> Ev + Clone + Send + 'static,
+        F: FnOnce(T) -> Ev + Clone + Send + 'static,
```
StuartHarris, line 38: I think this needs to stay as `Fn(T)` because we need to call it repeatedly for every chunk received from the server.

StuartHarris (changes requested):
Yeah, personally, I think this is the right thing to do! Thank you, Alex! Couple of small changes to allow the examples to compile.

examples/counter/shared/src/capabilities/sse.rs
charypar: I was looking at that, but we seem to clone it for every call...?

StuartHarris: for every call, but not for every chunk in the response

charypar: Ah yes, I see it now

StuartHarris (approved):
Thanks!

PR #231, opened 2024-05-15
Simplify app-facing API of crux_kv

charypar:
This should make crux_kv a bit more idiomatic to use from inside the core. It no longer forces the user to deal with receiving a response to the wrong operation (which shouldn't happen anyway so is largely just boilerplate), and also works on standard `Results` rather than the FFI protocol result.

I've only changed the capability and its tests, haven't done examples in case people are violently opposed to this 🙂

### DIFF
--- a/crux_kv/src/lib.rs
+++ b/crux_kv/src/lib.rs
@@ -78,7 +78,7 @@ where
     /// `KeyValueOutput::Get(KeyValueResult<Option<Vec<u8>>>)` as payload
     pub fn get<F>(&self, key: String, make_event: F)
     where
-        F: FnOnce(KeyValueResult) -> Ev + Send + Sync + 'static,
+        F: FnOnce(Result<Vec<u8>, KeyValueError>) -> Ev + Send + Sync + 'static,
     {
         self.context.spawn({
             let context = self.context.clone();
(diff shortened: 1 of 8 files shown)


crux_kv/src/lib.rs
```diff
@@ -205,7 +205,34 @@ where
             };
         }
 
-        response
+        response.into()
+    }
+}
+
+impl From<KeyValueResult> for Result<bool, KeyValueError> {
```
adwhit, line 212: Personally I don't like `From` impls which can panic or do anything more than the 'obvious' thing. You don't get a lot of benefit from using the trait, and it ends up with an unhelpful method name. `my_kv.into()` isn't as good as eg `my_kv.unwrap_exists()` or `my_kv.unwrap_value`, something which indicates you are asserting some truth about the variable.

StuartHarris: Would `TryFrom` impl be better? We could map the error

StuartHarris: Or even map the error in the `From` impl, actually

StuartHarris (approved):
This is so much better. I'll update the book with doing this as a "best practice" in another PR

crux_kv/src/lib.rs
charypar: Yea, perhaps a straight up impl methods that convert - maybe something like `expect_[operation]` - would be better

charypar: `unwrap_` is a better name, just for the record.

PR #232, opened 2024-05-15
`crux_kv` capability list keys

StuartHarris:
Adds ability to list the keys in the store, providing a prefix (which can be an empty string), and a cursor for pagination (can be 0 to start from the beginning).

### DIFF
--- a/crux_kv/src/lib.rs
+++ b/crux_kv/src/lib.rs
@@ -21,6 +21,13 @@ pub enum KeyValueOperation {
     Delete { key: String },
     /// Test if a key exists
     Exists { key: String },
+    // List keys that start with a prefix, starting at the cursor
+    ListKeys {
+        /// The prefix to list keys for, or an empty string to list all keys
+        prefix: String,
+        /// The cursor to start listing from, or 0 to start from the beginning
+        cursor: u64,
+    },
 }
 
 /// The result of an operation on the store.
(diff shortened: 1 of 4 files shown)


crux_kv/src/lib.rs
```diff
@@ -21,6 +21,13 @@ pub enum KeyValueOperation {
     Delete { key: String },
     /// Test if a key exists
     Exists { key: String },
+    // List keys that start with a prefix, starting at the cursor
+    ListKeys {
+        /// The prefix to list keys for, or an empty string to list all keys
+        prefix: String,
+        /// The cursor to start listing from, or 0 to start from the beginning
+        cursor: u64,
```
charypar, line 29: Not quite sure I thought it through enough, but would it make sense to be prefix OR cursor, but not both...? E.g. what happens if I reuse a cursor given to me for one prefix with a different prefix?

Presumably the shells would already need to keep the relevant cursor state around and that can include the original prefix, and that way the app doesn't need to worry about keeping the prefix around...?

Not sure... not quite clear on my own thinking, but red flags are going off in my head about the semantics now that I see it written down.

crux_kv/src/lib.rs
```diff
@@ -175,6 +188,35 @@ where
...
+        let this = self.clone();
+
+        self.context.spawn(async move {
+            let response = this.list_keys_async(prefix, cursor).await;
+            context.update_app(make_event(response))
+        });
+    }
+
+    /// List keys that start with the provided `prefix`, starting from the provided `cursor`,
+    /// while in an async context. This is used together with [`crux_core::compose::Compose`].
```
charypar, line 209: Might be nice to reiterate the cursor and prefix mechanics in this doc comment too

charypar (approved):

crux_kv/src/lib.rs
StuartHarris: I've introduced `KeyValueError::CursorNotFound`, and updated the doc comments to describe the semantics around how cursors should work from the shell's perspective

PR #234, opened 2024-05-16
relax callback bounds in crux_kv

adwhit:
Relax a couple more callback bounds in the newly-updated `crux_kv`


StuartHarris (approved):
Thanks1

Issue #222, opened 2024-04-05
Adding Compose capability caused typegen to fail


wasnotrice:
I tried adding the Compose capability to my app. The `shared` crate compiles, but typegen fails in `shared_types`, because the `Never` enum has no cases. See the backtrace below for details.

We don't need to include `Compose` or `Never` in typegen, since they are not exposed to shells, but is there a way to mark a capability as "excluded"? Or maybe another workaround?

Here's [a minimal repo that produces the error](https://github.com/wasnotrice/crux_compose). You don't even have to use Compose, just include it in Capabilities. 

```
❯ RUST_BACKTRACE=1 RUSTFLAGS=-Awarnings cargo build -p shared_types
   Compiling shared_types v0.1.0 (/Users/eric/code/crux_compose/shared_types)
error: failed to run custom build command for `shared_types v0.1.0 (/Users/eric/code/crux_compose/shared_types)`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `/Users/eric/code/crux_compose/target/debug/build/shared_types-5b2877aeefdefaf8/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=../shared

  --- stderr
  thread 'main' panicked at /Users/eric/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-reflection-0.3.6/src/de.rs:431:18:
  variant indexes must be a non-empty range 0..variants.len()
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/std/src/panicking.rs:647:5
     1: core::panicking::panic_fmt
               at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/panicking.rs:72:14
     2: core::panicking::panic_display
               at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/panicking.rs:196:5
     3: core::panicking::panic_str
               at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/panicking.rs:171:5
     4: core::option::expect_failed
               at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/option.rs:1988:5
     5: core::option::Option<T>::expect
[...]

PR #235, opened 2024-05-16
update API for crux_kv

StuartHarris:
Introduces a `Value` enum, which is essentially a monnomorphised `Option<Vec<u8>>` to represent responses where the key does not exist (`get()`), or there was no previous value (`delete()` or `set()`). This type needs to exist as we can't (currently) get `Option<T>` over the FFI boundary. Note:  for now, `build.rs` for shared types libraries need to register this `Value` type explicitly as `serde_reflection` won't see it.

### DIFF
--- a/examples/cat_facts/shared_types/build.rs
+++ b/examples/cat_facts/shared_types/build.rs
@@ -4,7 +4,7 @@ use crux_core::typegen::TypeGen;
 
 use shared::{
     http::HttpError,
-    key_value::{error::KeyValueError, KeyValueResponse},
+    key_value::{error::KeyValueError, value::Value, KeyValueResponse},
     CatFacts,
 };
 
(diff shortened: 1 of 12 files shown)


examples/cat_facts/shared_types/build.rs
```diff
@@ -15,9 +15,13 @@ fn main() -> anyhow::Result<()> {
 
     gen.register_app::<CatFacts>()?;
 
+    // types from `crux_http` that aren't automatically discovered
     gen.register_type::<HttpError>()?;
+
+    // types from `crux_kv` that aren't automatically discovered
     gen.register_type::<KeyValueResponse>()?;
     gen.register_type::<KeyValueError>()?;
+    gen.register_type::<Value>()?;
```
charypar, line 24: It may be beating a dead horse with a stick, but we could extend the Capability trait to support automatic registration of this kind of types - that way the auto discovery can explicitly _ask_ the capabilities to register types 🤔

charypar (approved):

examples/cat_facts/shared_types/build.rs
StuartHarris: I like this idea. I guess it depends on how long it takes me to implement new typegen (could be a while), because then we'd need to take it out again. Let's chat about it. Either way, another PR.

obmarg: I started an attempt at this very idea a while back, as an attempt to fix https://github.com/redbadger/crux/issues/222.

Forget why I didn't finish it, possibly just got bored 😅

PR #236, opened 2024-05-16
crux_kv reduce cloning

StuartHarris:
Clone the context instead of the whole capability by extracting standalone functions that don't need `self`

### DIFF
--- a/crux_kv/src/lib.rs
+++ b/crux_kv/src/lib.rs
@@ -105,17 +105,14 @@ where
 
     /// Read a value under `key`, will dispatch the event with a
     /// `KeyValueResult::Get { value: Vec<u8> }` as payload
-    pub fn get<F>(&self, key: String, make_event: F)
+    pub fn get<F>(self, key: String, make_event: F)
     where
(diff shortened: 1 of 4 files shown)


crux_kv/src/lib.rs
```diff
@@ -105,17 +105,14 @@ where
...
         F: FnOnce(Result<Option<Vec<u8>>, KeyValueError>) -> Ev + Send + Sync + 'static,
     {
-        self.context.spawn({
-            let context = self.context.clone();
-            let this = self.clone();
-
+        self.context.clone().spawn({
             async move {
-                let response = this.get_async(key).await;
-                context.update_app(make_event(response));
```
obmarg, line 118: Given that `get_async` basically only needs the `context` would it not be nicer to just avoid cloning `self` entirely?

Could extract a standalone `fn get_async(context: CapabilityContext<KeyValueOperation, Ev>), call that with a cloned context and keep the `&self`?

StuartHarris: Oh yeah that'd be loads better. I will do, Thanks 🤩

StuartHarris: much nicer! ❤️

PR #238, opened 2024-05-17
Requests now use `EffectId(u32)` instead of `Uuid`

StuartHarris:
- [x] replace Uuid with new type `EffectId(u32)`
- [x] use a slab to manage storage of callbacks for resolving effects
- [x] update examples, tests, docs, book

Note, this does not affect Rust shells that use the core directly, as they bypass the bridge and the registry.

Also, we need to measure the performance of the slab (vs the HashMap we were using beforehand). 

Incidentally, we now remove the slab entry once the effect is fully resolved (although subscription style callbacks, e.g. Server Sent Events, still keep the entry in case there are more resolutions. We need to look at how to signal the end of a subscription so these can be cleaned up too). Before this change all effects were "leaking", now its only the effects that can be resolved repeatedly).


crux_core/Cargo.toml
```diff
@@ -41,3 +41,4 @@ serde = { version = "1.0.201", features = ["derive"] }
 static_assertions = "1.1"
 rand = "0.8"
 url = "2.5.0"
+uuid = { version = "1.8.0", features = ["v4", "serde"] }
```
charypar, line 44: Is it still needed for dev...? 👀

charypar (approved):

crux_core/Cargo.toml
StuartHarris: Not really. There's a test for sending UUID across the typegen IIRC. Could use another type though

obmarg (approved):
Nice, I'd been thinking this would be a good idea.

PR #240, opened 2024-05-18
http encoding feature

StuartHarris:
Nightly clippy reported that the `crux_http` crate was using the feature `encoding`, which wasn't declared in `Cargo.toml`. This fixes that. But we should agree that this is what we want to do.

Note that the `decode_body` function changes to use `encoding_rs` for non-wasm targets, and `web-sys` for wasm targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,6 +391,7 @@ dependencies = [
  "async-trait",
  "crux_core",
  "derive_builder",
+ "encoding_rs",
  "futures-test",
  "futures-util",
  "http-types-red-badger-temporary-fork",
@@ -399,6 +400,7 @@ dependencies = [
  "serde_json",
  "thiserror",
  "url",
+ "web-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


examples/counter/Cargo.toml
```diff
@@ -20,10 +20,10 @@ rust-version = "1.66"
 
 [workspace.dependencies]
 anyhow = "1.0.83"
-crux_core = "0.7"
-crux_http = "0.9"
-# crux_core = { path = "../../crux_core" }
-# crux_http = { path = "../../crux_http" }
```
charypar, line 26: Committed inadvertently?

charypar (approved):

examples/counter/Cargo.toml
StuartHarris: kinda. Was using to test. What I really want is them to point to local paths when on a PR, and to the published versions when on master. 😓 Will revert

obmarg:
Although having said that, maybe every server has been sending me utf-16 and I've just never noticed because of features like these.

StuartHarris:
> maybe forget I spoke
Haha! It looks to me like the implementation that uses `encoding_rs` _could_ be better if it uses copy-on-write semantics, not sure.

charypar:
@StuartHarris what do we wanna do about this then? 👀

StuartHarris:
Since Rust 1.80.0 (out today) CI builds will fail with a warning that the `encoding` feature doesn't exist, so I'm going to merge this and we should update all open PRs from master.

PR #241, opened 2024-05-20
Simplify construction of capabilities, so they don't require the app type

charypar:
This was preventing having Apps with generic parameters, because the Effect derive macro would want the type name for the App.

Turns out that was down to the type definition of `WithContext` which required the `App` type, but doesn't strictly need to. It's job is to map capability's output to type `Effect` (wrapping it). 

`WithEffect` can now, strictly speaking, be implemented on anything, but we do restrict it down to being the App's capabilities type further up the chain in `Core::new`.

A nice side effect (possibly independent?) is that `Core::new` no longer needs to be generic. Sadly that also makes this a breaking change 😢

### DIFF
--- a/crux_core/src/capability/mod.rs
+++ b/crux_core/src/capability/mod.rs
@@ -343,7 +343,7 @@ pub trait Capability<Ev> {
 /// #         }
 /// #     }
 /// # }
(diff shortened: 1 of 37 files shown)


StuartHarris (approved):
So much nicer!

PR #253, opened 2024-07-02
Better HttpRequest and KeyValueOperation Debug format

### LINKED ISSUE #252: crux_http and crux_kv: more ergonomic `Debug` on operations
The http requests and key value requests debug print as a giant list of numbers, which is not a great DX when used in tests. It would be nicer if they 

1. attempted to parse as UTF-8
2. otherwise printed as either numbers or hex pairs
3. truncated to the first N characters (maybe 50?)

adwhit:
fixes #252

### DIFF
--- a/crux_http/src/protocol.rs
+++ b/crux_http/src/protocol.rs
@@ -16,7 +16,7 @@ pub struct HttpHeader {
     pub value: String,
 }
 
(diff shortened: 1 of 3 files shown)


charypar (approved):
Lovely, thank you!

PR #254, opened 2024-07-04
Debug fmt handles unicode boundaries

adwhit:
I realised that #253 can introduce a panic when handling non-ascii unicode character (see e.g. https://old.reddit.com/r/rust/comments/765n76/psa_caution_with_slicing_strings/). Here is the fix.

Also bumped the `pnpm` action to `v4`, as suggested in https://github.com/pnpm/pnpm/issues/6424.

### DIFF
--- a/crux_kv/src/lib.rs
+++ b/crux_kv/src/lib.rs
@@ -49,7 +49,7 @@ impl std::fmt::Debug for KeyValueOperation {
                     if s.len() < 50 {
                         format!("\"{s}\"")
                     } else {
-                        format!("\"{}\"...", &s[..50])
+                        format!("\"{}\"...", &s.chars().take(50).collect::<String>())
                     }
                 } else {
                     format!("<binary data - {} bytes>", value.len())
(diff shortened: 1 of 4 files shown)


charypar (approved):
Oops, probably should've caught that in the last review 😓

crux_kv/src/lib.rs
```diff
@@ -45,11 +45,11 @@ impl std::fmt::Debug for KeyValueOperation {
         match self {
             KeyValueOperation::Get { key } => f.debug_struct("Get").field("key", key).finish(),
             KeyValueOperation::Set { key, value } => {
-                let body_repr = if let Ok(s) = std::str::from_utf8(&value) {
+                let body_repr = if let Ok(s) = std::str::from_utf8(value) {
                     if s.len() < 50 {
                         format!("\"{s}\"")
                     } else {
-                        format!("\"{}\"...", &s[..50])
+                        format!("\"{}\"...", s.chars().take(50).collect::<String>())
```
obmarg, line 52: Would be nice to avoid allocating a string only to write it straight into another string and drop it again.

```
let (index, _) = s.char_indices().nth(50);
format!("\"{}\"...", &s[..index])
```

PR #258, opened 2024-07-11
Delegate type registration to capabilities

### LINKED ISSUE #257: Delegate type registration to capabilities
Related to #222.

Currently the type discovery for type generation is limited, and generally doesn't see into operation types in capabilities. One way around this would be to extend the `Capability` trait to support their own type generation - the type discovery machinery would simply call the trait method to ask the capability to register any relevant types.

While we're working on our own type generation that should superseed this, it may be a quick win to make this common paper cut less sharp.

StuartHarris:
resolves #257

### DIFF
--- a/crux_core/src/capabilities/compose.rs
+++ b/crux_core/src/capabilities/compose.rs
@@ -157,4 +157,14 @@ impl<Ev> Capability<Ev> for Compose<Ev> {
     {
         Compose::new(self.context.map_event(f))
     }
+
+    #[cfg(feature = "typegen")]
+    fn register_types(_generator: &mut crate::typegen::TypeGen) -> crate::typegen::Result {
+        panic!(
+            r#"
+            The Compose Capability should not be registered for type generation.
+            Instead, use #[effect(skip)] to skip the generation of an effect variant for the Compose Capability.
+            "#
+        )
+    }
 }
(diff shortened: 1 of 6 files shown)


crux_core/src/capabilities/compose.rs
```diff
@@ -157,4 +157,14 @@ impl<Ev> Capability<Ev> for Compose<Ev> {
...
         Compose::new(self.context.map_event(f))
     }
+
+    #[cfg(feature = "typegen")]
+    fn register_types(_generator: &mut crate::typegen::TypeGen) -> crate::typegen::Result {
+        panic!(
+            r#"
+            The Compose Capability should not be registered for type generation.
+            Instead, use #[effect(skip)] to skip the generation of an effect variant for the Compose Capability.
+            "#
```
obmarg, line 167: Any consequence to just silently doing nothing here instead of panicing?

StuartHarris: The typegen blows up on the `Never` operation type, so figured panicking is an opportunity to prompt using the effect skip attirbute

charypar (approved):
Nice! Wanna follow this up with a PR extending HTTP and KV to automatically register their relevant types?

PR #260, opened 2024-07-12
crux_http and crux_kv now register their additional types

StuartHarris:


crux_kv/README.md
```diff
@@ -4,6 +4,11 @@ This crate contains the `KeyValue` capability, which can be used to ask the Shel
 
 Currently it provides an interface for getting, setting, and deleting keys, checking if keys exists in the store, and listing keys that start with a prefix.
 
+## Getting Started
+
+Add `crux_kv` as a dependency in your app's `Cargo.toml`.
+Add `crux_kv`, with feature `typegen`, as a build dependency in your `share_types` crate's `Cargo.toml`.
```
charypar, line 10:
```suggestion
Add `crux_kv`, with feature `typegen`, as a build dependency in your `shared_types` crate's `Cargo.toml`.
```

examples/cat_facts/Cargo.toml
```diff
@@ -12,16 +12,16 @@ rust-version = "1.66"
...
+# crux_core = "0.8"
+# crux_http = "0.9"
+# crux_kv = "0.4"
+# crux_platform = "0.1"
+# crux_time = { version = "0.4", features = ["chrono"] }
+crux_core = { path = "../../crux_core" }
+crux_http = { path = "../../crux_http" }
+crux_kv = { path = "../../crux_kv" }
+crux_platform = { path = "../../crux_platform" }
+crux_time = { path = "../../crux_time", features = ["chrono"] }
```
charypar, line 24: Did you mean to leave these here?

examples/counter/Cargo.toml
```diff
@@ -20,10 +20,10 @@ rust-version = "1.66"
...
 [workspace.dependencies]
 anyhow = "1.0.86"
-crux_core = "0.8"
-crux_http = "0.9"
-# crux_core = { path = "../../crux_core" }
-# crux_http = { path = "../../crux_http" }
+# crux_core = "0.8"
+# crux_http = "0.9"
+crux_core = { path = "../../crux_core" }
+crux_http = { path = "../../crux_http" }
```
charypar, line 26: As above

charypar (approved):
Love it!

crux_kv/README.md
StuartHarris: I've since changed this so that we use the re-exported Capability from the shared crate, but with the `typegen` feature enabled.

examples/cat_facts/Cargo.toml
StuartHarris: yeah for now. Tests will need to run against the local crates

examples/counter/Cargo.toml
StuartHarris: as above

crux_kv/README.md
StuartHarris: see https://github.com/redbadger/crux/pull/260/commits/316eeb300e0b3f099bb92223549ced607e3d31a1

examples/cat_facts/Cargo.toml
StuartHarris: will revert just before I merge to master

charypar (approved):

charypar:
For posterity - bit of a quesiton of how we get this out the door. 

We need to ship the new crates in order for examples to use them, but the example need the new crates to work to validate them. I think the steps probably are
1. Validate with an example using local crates as dependencies
2. Revert the example to the latest published crates
3. Publish new versions of crates 
4. Update the examples to use new crates and benefit from the new features

For the future, we might want a dedicated example which always refers to the `master` version of the crux crates which we test on. This will need to be reasonably complex, and we should probably put a big warning all over its code somehow. But then we could avoid this back and forth dance.

PR #264, opened 2024-08-13
implement type registration in crux_time

StuartHarris:


docs/src/getting_started/core.md
```diff
@@ -215,16 +215,39 @@ The `build.rs` file should now look like this:
...
 ````admonish tip
- You may also need to register any nested enum types (due to a current limitation with the reflection library, see <https://github.com/zefchain/serde-reflection/tree/main/serde-reflection#supported-features>).
-Here is an example of this from the [`build.rs`](https://github.com/redbadger/crux/blob/master/examples/notes/shared_types/build.rs) file in the `shared_types` crate of the [notes example](https://github.com/redbadger/crux/tree/master/examples/notes):
+Due to a current limitation with the reflection library,
+you may need to manually register nested enum types in your `build.rs` file.
+(see <https://github.com/zefchain/serde-reflection/tree/main/serde-reflection#supported-features>)
+
+
+*Note, you don't have to do this for the latest versions of the `crux_http`, `crux_kv` or `crux_time`
+capabilities, which now do this registration for you.*
```
charypar, line 224: Do we need instructions to import the capabilities in `shared_types` with the `typegen` feature enabled?

examples/cat_facts/Cargo.toml
```diff
@@ -12,11 +12,16 @@ rust-version = "1.66"
...
+# crux_core = "0.8"
+# crux_http = "0.9"
+# crux_kv = "0.4"
+# crux_platform = "0.1"
+# crux_time = { version = "0.4", features = ["chrono"] }
+crux_core = { path = "../../crux_core" }
+crux_http = { path = "../../crux_http" }
+crux_kv = { path = "../../crux_kv" }
+crux_platform = { path = "../../crux_platform" }
+crux_time = { path = "../../crux_time", features = ["chrono"] }
```
charypar, line 24: Tempted to say `cat_facts` should stay like this so we have an easy test case for the build and typegen process. 

We did talk about introducing an integration test for that kind of purpose didn't we 🤔 Unrelated to this PR though.

charypar (approved):
Looks good, but could do with a few more instructions for using the capability directed typegen.

docs/src/getting_started/core.md
StuartHarris: oh yeah

StuartHarris: https://github.com/redbadger/crux/pull/264/commits/ac1362a25974cff6693b6538b0dbd8e0e068de40

examples/cat_facts/Cargo.toml
StuartHarris: will update to the RC once published

PR #267, opened 2024-09-25
Mark HTTP body with serde_bytes

arendjr:
This marks the `body` field in `crux_http`'s `HttpRequest` and `HttpResponse` types with the `#[serde(with = "serde_bytes")]` annotation. This is effectively a hint for Serde to serialize it as a byte array instead of a vector of numbers.

See also: https://docs.rs/serde_bytes/latest/serde_bytes/

I could put this behind a feature flag, but given that the annotation is ignored on formats that don't support byte arrays, I think there might be no need. This also means it will remain a vector of numbers for JSON encoding, since JSON has no functionality for binary encoding. The annotation is mainly relevant for formats such as MessagePack and CBOR.

For formats that work with fixed-size vectors, such as bincode, I don't think there will be any difference either, although I have not tested this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -406,6 +406,7 @@ dependencies = [
  "http-types-red-badger-temporary-fork",
  "pin-project-lite",
  "serde",
+ "serde_bytes",
  "serde_json",
  "thiserror",
  "url",
@@ -1458,6 +1459,15 @@ dependencies = [
  "once_cell",
  "serde",
  "thiserror",
+]
+
+[[package]]
+name = "serde_bytes"
+version = "0.11.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "387cc504cb06bb40a96c8e04e951fe01854cf6bc921053c954e4a606d9675c6a"
+dependencies = [
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


StuartHarris (approved):
Looks great! Than you.

arthurgeek:
this change broke some examples, like https://github.com/redbadger/crux/blob/9e249e5ca558d85a6d7e4c3bef15bea8bc810d62/examples/counter/Android/app/src/main/java/com/example/counter/http.kt#L27

error:

`Type mismatch. Required: Bytes! Found: List<Byte>`

StuartHarris:
@arthurgeek thank you! Fixing now

StuartHarris:
About to update examples, but for anyone else that comes across this in their code:
```diff
--- a/examples/counter/Android/app/src/main/java/com/example/counter/http.kt
+++ b/examples/counter/Android/app/src/main/java/com/example/counter/http.kt
@@ -23,7 +23,7 @@
             }
         }
     }
-    val bytes: Bytes = response.body()
+    val bytes = Bytes.valueOf(response.body())
     val headers = response.headers.flattenEntries().map { HttpHeader(it.first, it.second) }
     return HttpResponse(response.status.value.toShort(), headers, bytes)
 }```

PR #272, opened 2024-10-10
AppTester Send

beltram:
Hello there,

I was trying to test an app where app instance communicate over a WebSocket. To do so I had the receiving party waiting in an async task for a message and then have the app resolve the event. To do so I required to send it to another thread, hence requiring Send.

Hence these small changes.

Thoughts.

PS: thanks for this crate, love it !

### DIFF
--- a/crux_core/src/testing.rs
+++ b/crux_core/src/testing.rs
@@ -1,7 +1,6 @@
 //! Testing support for unit testing Crux apps.
-use std::rc::Rc;
-
 use anyhow::Result;
+use std::sync::Arc;
 
 use crate::{
     capability::{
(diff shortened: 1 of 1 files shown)


charypar (approved):
Nice catch thank you!

charypar:
Unrelated to this PR, but would you be willing to share your WebSocket capability to the comunity contributed ones? https://github.com/redbadger/crux/tree/master/contrib/capabilities 

I'm working on a project with another WebSocket capability and it would be super interesting to compare notes to see whether there's one we could turn into a crux maintained crate.

charypar:
Ah, would you mind running the changes through `cargo fmt`, so CI is happy? 🙏🏻

beltram:
> Ah, would you mind running the changes through cargo fmt, so CI is happy? 🙏🏻
Done, sorry my IDE was drunk.

Regarding Websocket, sure, I'd love to ! I'm not 100% happy with the current capability so I might modify it a bit in the coming days. We also have odd requirements so I'll try to abstract them. I think a discussion in here would be the best place to discuss this and share snippets, WDYT ?

PR #275, opened 2024-10-18
Add helper methods to `Update` to remove test boilerplate

adwhit:
Here are a couple of ideas for how to make the tests less boilerplate-y. They are mostly minor helper methods but together I think they make the tests significantly more readable while also being more robust.

The most controversial parts are likely to be

* `fn expect_<effect>` to go along with `fn into_<effect>`
    - only saves a few keystrokes vs `into_<effect>().unwrap()` but I think it reads better
* Adding `#[must_use]` to the `Update` struct. This is to force the user to make it clear that they haven't forgotten that more stuff might happen in 'real life'
   - If we expect it to be empty, use `.assert_empty()`, otherwise write `let _update = ...` to silence the warning

I have ported a couple of tests to the new methods to illustrate how it might change how tests look. If we are happy with it I could expand the changes to all tests in the repo.

### DIFF
--- a/crux_core/src/testing.rs
+++ b/crux_core/src/testing.rs
@@ -76,6 +76,21 @@ where
         Ok(self.context.updates())
     }
 
+    /// Resolve an effect `request` from previous update, then run the resulting event
+    ///
+    /// This helper is useful for the comment case where  one expects the effect to resolve
+    ///  to exactly one event, which should then be run by the app.
+    pub fn resolve_to_event_then_update<Op: Operation>(
+        &self,
+        request: &mut Request<Op>,
+        value: Op::Output,
+        model: &mut App::Model,
+    ) -> Update<Ef, App::Event> {
+        request.resolve(value).expect("failed to resolve request");
+        let event = self.context.updates().expect_one_event();
+        self.update(event, model)
+    }
+
     /// Run the app's `view` function with a model state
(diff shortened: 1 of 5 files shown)


StuartHarris (approved):
Nice! (there are just a couple of doc comments that need updating)

crux_core/src/testing.rs
```diff
@@ -76,6 +76,21 @@ where
         Ok(self.context.updates())
     }
 
+    /// Resolve an effect `request` from previous update, then run the resulting event
+    ///
+    /// This helper is useful for the comment case where  one expects the effect to resolve
+    ///  to exactly one event, which should then be run by the app.
+    pub fn resolve_to_event_then_update<Op: Operation>(
```
charypar, line 83: perhaps this should be `expect_one_event_then_update`? There's a fluent API hiding in there as well.

charypar (approved):
Like it! One naming suggestion.

crux_core/src/testing.rs
adwhit: I chose the word `resolve` to mirror the `Update::resolve` method. I think there is a distinction between running an event and resolving an effect.

The fluent API of something like `resolve_to_event().then_update(model)` might be there somewhere but as written the types won't line up...

PR #203, opened 2024-02-20
[WIP] Update counter example to use Tauri v2

nuke-web3:
***WIP Draft***

Working (on my machine :upside_down_face: ) example updated using https://beta.tauri.app/blog/tauri-2-0-0-beta/ .

Feel free to close if not wanted, as I know it's too early to update here to an unstable dependency.I am hopeful once v2 is out, this PR will still be viable. :crossed_fingers: 

### Todo

- [x] Built tauri app emits "Unhandled Promise Rejection: event.listen not allowed. Permissions associated with this command: event:allow-listen, event:default"... https://beta.tauri.app/references/v2/js/core/namespaceevent/#listen seems to be related... maybe tauri has new security measures?
  - [x] https://beta.tauri.app/references/v2/acl/ indeed, you must add a new config file :ok:

PR #276, opened 2024-10-18
Update dependencies

StuartHarris:
Updates all the rust and ppm dependencies for the core, crates and examples. 
Also updates the Dioxus examples (using `0.6.0-alpha.3`)
Also updates the Tauri example to use `2.0.3`, which closes https://github.com/redbadger/crux/pull/203


charypar (approved):
Trusting the CI 😅

PR #279, opened 2024-10-21
feat(crux_time): cancellable timer

PatrykBuniX:
Hi! I was playing around with the `time` capability and realised there's no way to clear the pending timer.  This PR adds this possibility. 
 
This PR adds `TimeId` field for relevant requests/responses of `crux_time` capability. Since we know the id of the timer we are able to clear it using the new time request: `TimeRequest::Clear {id: TimeId}`. Once core notifies the shell (`notify_shell` since we don't expect any response to this particular request) with this effect, the shell should fire the initially set timer sooner with a `TimeResponse::Cleared` instead of `TimeResponse::InstantArrived`/`TimeResponse::DurationElapsed`.

I'm open to any suggestions, I'm also aware that this is a breaking change, maybe you guys see a non-breaking solution to this? Otherwise I will just create a custom capability 😉

### DIFF
--- a/crux_time/src/lib.rs
+++ b/crux_time/src/lib.rs
@@ -15,21 +15,32 @@ pub use instant::Instant;
 use serde::{Deserialize, Serialize};
 
 use crux_core::capability::{CapabilityContext, Operation};
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum TimeRequest {
     Now,
-    NotifyAt(Instant),
-    NotifyAfter(Duration),
+    NotifyAt { id: TimerId, instant: Instant },
+    NotifyAfter { id: TimerId, duration: Duration },
+    Clear { id: TimerId },
+}
+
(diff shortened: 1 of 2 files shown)


charypar (approved):
Looks great, we've had something like that in mind as well, just didn't get around to it. Thank you!

Can I ask what the use for this was that you found?

PatrykBuniX:
> Can I ask what the use for this was that you found?

I was persisting the timer fire date and resuming it upon app reload and of course I didn't want the timer to live forever, hence the need to clear it.

charypar:
Makes a lot of sense!

We've got an issue in the back log to think about how we do effect cancellation universally for all effects, but this is good for timer 👍🏻

PatrykBuniX:
@charypar I don't think the ci failure is related with the changes I made, can you take a look or re-run? 🙏

charypar:
Yea, that looks like a `must_use` has been added `Update`, which the examples have not been updated for. /cc @StuartHarris

StuartHarris:
@PatrykBuniX my apologies. I really should have published `crux_core` as 0.10.0 instead of 0.9.1, although it should be only tests that are affected. I have merged a temporary fix into master, so if you could rebase on top of that you should be good.

PR #280, opened 2024-10-21
`Clone` bound on `Operation` (breaking) and test helper methods

StuartHarris:
- Adds a `Clone` bound on  the `Operation` trait so that we can examine the operation and still resolve its owning request later on — this is a breaking change.
- Adds a `take_effects` method on `Update` to allow you to take effects off the Update that match the predicate
- Adds a `take_effects_partitioned_by` method on `Update` to allow you to take effects off the Update that match the predicate and also the remaining effects that don't match
- Updates all the crates' tests and the examples' tests to make use of these new testing helpers (and the ones already released in `crux_core` 0.9.0 and 0.9.1)
- Note: the examples now reference local crates, but I'll change that after the helpers are published

### DIFF
--- a/crux_http/tests/with_tester.rs
+++ b/crux_http/tests/with_tester.rs
@@ -143,7 +143,7 @@ mod tests {
         let app = AppTester::<App, _>::default();
         let mut model = Model::default();
 
-        let mut request = app
+        let request = &mut app
             .update(Event::Get, &mut model)
             .expect_one_effect()
             .expect_http();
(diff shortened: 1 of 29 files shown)


crux_http/tests/with_tester.rs
```diff
@@ -188,9 +186,10 @@ mod tests {
         let app = AppTester::<App, _>::default();
         let mut model = Model::default();
 
-        let mut update = app.update(Event::Post, &mut model);
-
-        let Effect::Http(request) = update.effects_mut().next().unwrap();
+        let request = &mut app
+            .update(Event::Post, &mut model)
+            .expect_one_effect()
+            .expect_http();
```
charypar, line 192: ❤️

charypar (approved):
Ahhh 💆🏻‍♂️

PR #283, opened 2024-10-30
feat(crux-time): Allow for custom timer ids

Orphe0s10:
# Objective

Allow for the *core* to specify custom timer ids, this enables persisted timers which can be resumed and canceled accross app runs

# Solution

`TimerId` inner type becomes a `String` for flexibility, which can be constructed by any `ToString` type.
`Timer::Id` still can be constructed automatically via incremental `AtomicUsize` via `TimerId::new_atomic`

### DIFF
--- a/crux_time/src/lib.rs
+++ b/crux_time/src/lib.rs
@@ -12,10 +12,43 @@ pub use duration::Duration;
 pub use error::TimeError;
 pub use instant::Instant;
 
-use serde::{Deserialize, Serialize};
-
 use crux_core::capability::{CapabilityContext, Operation};
-use std::sync::atomic::{AtomicUsize, Ordering};
+use serde::{Deserialize, Serialize};
+use std::{
+    fmt::Display,
+    ops::Deref,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
+#[serde(transparent)]
+pub struct TimerId(pub String);
+
+impl Deref for TimerId {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Display for TimerId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl TimerId {
+    #[inline]
+    pub fn new(v: impl ToString) -> Self {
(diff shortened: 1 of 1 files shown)


charypar:
Hello, thanks for the PR! 

I unfortunatly must say that I'm not convinced this is a good idea, and I believe the motivating use-case can be achieved without changing the time capability API. (In fact the async API of the capability has already gone too far in accepting IDs which I've missed, and we'll probably need to revert that change.)

The problem with allowing custom IDs to be passed in is preventing clashes - we'd need to keep some sort of used set and make sure we don't reuse accidentally, returning an error if a duplicate is used, making the API a lot more complex. I think the web browser time APIs are reasonably good evidence that this shouldn't be necessary, and having the initial timer API calls return IDs used for cancellation is a better idea.

As for resuming timers across app restarts, I don't see why this would be necessary. In my mind, a timer is only relevent while the app (and the shell holding the system clock resource) are running. Once the app closes, timer is cancelled and will not fire, unless there's a specific operating system mechanism to relaunch the app. Therefore, there are two kinds of timers in that scenario - the _logical_ timer which "runs" across app launches with a stable ID, and the ephemeral timer providing interface to the operating system clock, which does not survive termination of the app. The latter is what `crux_time` is supporting.

The logical timer can be built on top of the clock interface (and persistent storage), by:
1. Creating a stable ID for the logical timer and storing the intended deadline under that ID in persistents storage
2. Kicking off a `crux_time` timer and remembering the mapping between the returned timer ID and the logical timer ID
3. When the timer expires, looking up the logical ID and clearing the persistent storage.
4. If the app stops before the timer fires, on the next startup, there will be timers in storage, which can be restarted as new `crux_time` timers if the deadline hasn't passed yet, and their ID mapping updated.
5. if the OS supports relaunching the app on timer, then the timer ID can be provided to the app on relaunch, so that the timer can be looked up.

This is definitely more complicated for the core, but as far as I can see, this logic for resuming timers would have to be implemented in the shell anyway, so this just moves it into the core...?

Am I missing something?

charypar:
I appreciate use cases for such API exist, but making the implementation part of the capability doesn't automatically support them, it moves the work of implementing the details of the necessary bookkeeping from the app into the shell(s), which does not look like a good trade off to me (one core impl for N shell impls).

Given that:
* this can be achieved with the current API and a small amount of user code
* adding the new API would force all shell implementations to support it

I don't think we want to make it part of the default time capability, sorry to disappoint. 

You can always fork the capability and implement this API in your fork if the trade-off works out in your specific case.

PR #284, opened 2024-10-30
crux_time API improvement

adwhit:
This is a suggestion to remove the `id: TimerId` argument from various `crux_time` APIs. In my opinion it doesn't make sense to provide this value, it should be considered an opaque type useful only inside `crux_time` (so perhaps consider removing the ability to independently create `TimerId`s too). YMMV!

Obviously this is a *breaking change*.

### DIFF
--- a/crux_time/src/lib.rs
+++ b/crux_time/src/lib.rs
@@ -44,7 +44,7 @@ fn get_timer_id() -> TimerId {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum TimeResponse {
-    Now(Instant),
+    Now { instant: Instant },
     InstantArrived { id: TimerId },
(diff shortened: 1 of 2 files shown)


charypar (approved):
I think I agree with this - allowing the ID to be set was an oversight 🙈 

@StuartHarris any opinions?

StuartHarris:
Yes I agree. Thanks @adwhit !

adwhit:
While we're breaking things, refactoring `TimeResponse::Now(Instant)` to `TimeResponse::Now { instant: Instant }` would make the bindings nicer

charypar:
> While we're breaking things, refactoring TimeResponse::Now(Instant) to TimeResponse::Now { instant: Instant } would make the bindings nicer

wanna do that before we merge it?

adwhit:
Yep done

charypar (approved):
Happy with this, thank you!

/cc @StuartHarris this will be a breaking change to the time capability once more.

PR #288, opened 2024-11-26
Clarified experimental nature of crux in its own section, which state…

mhedgpeth:
…s the path to 1.0.

As I read through the docs, I was curious as to what made the software experimental. When I asked on Zulip, I learned that the app is used in production, and that the experimental nature of it is really related to its API and use cases.

So I clarified that in the topic, to hopefully make it easier for people to onboard.

In addition, when looking at this, one early problem I had with crux is the lack of Windows Desktop support, and a few hours of research later, I realized that it would be super early and barely an inconvenience with C#/WinUI3 or WPF/PInvoke - it just needs to be done. So I included that callout to hopefully help people who were thinking about crux for the first time.

### DIFF
--- a/docs/src/overview.md
+++ b/docs/src/overview.md
@@ -93,5 +93,5 @@ Below is a list of some of the things we know we want to do before 1.0:
 * Revised capabilities and effects to allow
   for [app composition](./guide/composing.md) in larger apps
 
-Until then, we hope you will work with us on the rough edges, and factor in
-frequent API updates as we evolve.
+Until then, we hope you will work with us on the rough edges, and adapt to the necessary
+API updates as we evolve. We strive to minimise the impact of changes as much as we can, but before 1.0, some breaking changes will be unavoidable. 
--- a/docs/src/overview.md
+++ b/docs/src/overview.md
@@ -90,8 +90,9 @@ Below is a list of some of the things we know we want to do before 1.0:
   to deal with
 * Better FFI code generation to enable support for more languages (e.g. C#, Dart, even C++...) 
   and in trurn more Shells (e.g. .NET, Flutter) which will also enable Desktop apps for Windows
-* Revised capabilities and effects to allow
-  for [app composition](./guide/composing.md) in larger apps
+* Revised capabilities and effects to allow for better, more natural
+  [app composition](./guide/composing.md) in larger apps, for composing capabilities, 
+  and generally for a more ergonomic effect API overall
 
 Until then, we hope you will work with us on the rough edges, and adapt to the necessary
 API updates as we evolve. We strive to minimise the impact of changes as much as we can, but before 1.0, some breaking changes will be unavoidable. 
--- a/docs/src/overview.md
+++ b/docs/src/overview.md
@@ -81,9 +81,9 @@ overall goals of Crux are to:
 
 ## Path to 1.0
 
-Crux is used in production apps today, at least one quite large one. In that sense
-we consider it production ready. However, to truly be at 1.0 with a more stable
-API, we will take the following steps:
+Crux is used in production apps today, and we consider it production ready. However, we still have a number of things to work on to call it 1.0, with a stable API, and other things one would expect from a mature framework. 
+
+Below is a list of some of the things we know we want to do before 1.0:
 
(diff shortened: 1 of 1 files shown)


fraschm1998:
I was reading the same thread and you mentioned updating the docs for newcomers, I've been using chatgpt to help wrap my head around how crux works and I found it easier to understand than the current docs. Here's a partial chat history, perhaps this part of this could be used: https://chatgpt.com/share/674b2e9d-323c-800f-996a-3b80d1fe8a91

docs/src/overview.md
```diff
@@ -1,6 +1,6 @@
 # Overview
 
-Crux is an **experimental** approach to building cross-platform applications
+Crux is an approach to building cross-platform applications
```
charypar, line 3:
```suggestion
Crux is a framework for building cross-platform applications
```

docs/src/overview.md
```diff
@@ -78,3 +78,20 @@ overall goals of Crux are to:
 - Separate the behavior from the look and feel and interaction design
 - Use the native UI tool kits to create user experience that is the best fit for
   a given platform
+
+## Path to 1.0
+
+Crux is used in production apps today, at least one quite large one. In that sense
+we consider it production ready. However, to truly be at 1.0 with a more stable
+API, we will take the following steps:
```
charypar, line 86:
```suggestion
Crux is used in production apps today, and we consider it production ready. However, we still have a number of things to work on to call it 1.0, with a stable API, and other things one would expect from a mature framework. 

Below is a list of some of the things we know we want to do before 1.0:
```

docs/src/overview.md
```diff
@@ -78,3 +78,20 @@ overall goals of Crux are to:
...
+
+* Improved documentation, code examples, and example apps for newcomers
+* Improved onboarding experience, with less boilerplate code that end users have
+  to deal with
+* Better FFI code generation to enable support for more Shells (e.g. .NET,
+  Flutter) which will also enable Desktop apps in Windows
```
charypar, line 92:
```suggestion
* Better FFI code generation to enable support for more languages (e.g. C#, Dart, even C++...) 
  and in trurn more Shells (e.g. .NET, Flutter) which will also enable Desktop apps for Windows
```

docs/src/overview.md
```diff
@@ -78,3 +78,20 @@ overall goals of Crux are to:
...
+* Revised capabilities and effects to allow
+  for [app composition](./guide/composing.md) in larger apps
```
charypar, line 94:
```suggestion
* Revised capabilities and effects to allow for better, more natural
  [app composition](./guide/composing.md) in larger apps, for composing capabilities, 
  and generally for a more ergonomic effect API overall
```

docs/src/overview.md
```diff
@@ -78,3 +78,20 @@ overall goals of Crux are to:
...
+
+Until then, we hope you will work with us on the rough edges, and factor in
+frequent API updates as we evolve.
```
charypar, line 97:
```suggestion
Until then, we hope you will work with us on the rough edges, and adapt to the necessary
API updates as we evolve. We strive to minimise the impact of changes as much as we can, but before 1.0, some breaking changes will be unavoidable. 
```

charypar (approved):
Thanks for making a start on this! Sorry it took me so long to get to it. I've suggested some minor wording changes, please let me know if you're happy with them.

charypar:
@fraschm1998 thanks for that! I enjoed reading it, the robot analogy is great 😄 Definitely some analogies and explanations worth stealing in that conversation.

charypar:
Some of ChatGPT's ... chat :) about when to send an event in a coroutine and when to call it sunchronously are not quite right I don't think, but they point to a missing piece of documentation to write - how the shells should deal with concurrency and how the core fits in.

charypar:
@mhedgpeth There's about three more suggestions to get in unless you don't agree with those. Let me know. 

Other than that I'm happy to merge this.

mhedgpeth:
Fantastic, everything is accepted. Thanks for working with me on this

charypar:
Thanks for pushing us to do this @mhedgpeth!

PR #289, opened 2024-12-03
feat(examples): add sparkline and stop/start for bridge_echo on iOS

Binlogo:
This pull request updates the `examples/bridge_echo` project, specifically improving the iOS application and the shared Rust library by adding a sparkline view and implementing stop/start logic, as mentioned in https://github.com/redbadger/crux/pull/136#issuecomment-1700793055 by @charypar 

### An Issue to Mention

As depicted in the graph below, the number of calls will decrease over time:

image

### DIFF
--- a/examples/bridge_echo/iOS/BridgePerf.xcodeproj/project.pbxproj
+++ b/examples/bridge_echo/iOS/BridgePerf.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


charypar:
Nice, thank you! The gradual drop is really interesting - it shows the impact of the view model size (and the serialisation work) on the bridge throughput. As the log for the sparkline gets larger, it takes longer and longer to ser/de it over the boundary on every round trip. What's _really_ interesting is that the impact is non-linear - the throughput seems to be converging to some overall minimum.

It's probably good to have that, and show what can be expected with larger view models. It will also be useful for comparison with with fine-grained view model updates, once we get that support in place.

charypar (approved):

Binlogo:
@charypar, please approve the workflow if this PR meets your standards. Thank you! 🫶

charypar:
Thanks again! Great work 🫶

Issue #247, opened 2024-05-31
Effect cancellation


charypar:
On a number of occassions I've ended up implementing domain specific signaling in capabilities for canceling effects. This is especially useful for subscriptions.

There are two directions:

* Shell cancells a subscription, communicating that the stream has eneded
* Core is no longer interesgted in the subscription (the relevant task has ended)

In the latter case, attempting to resolve the effect again will, I'm pretty sure, result in a panic at the moment, which is not ideal.

It would be good if Crux could automatically signal to the shell that the subscription has been terminated and any related resources on the shell side can be cleaned up.

obmarg:
Just dropping my idea for implementing this from my chat w/ @charypar 

1. Each App/Executor whatever has a (synchronous) unbounded MPSC for dead subscriptions
2. Each stream gets a sender for that when they start (probably at the point they are first awaited rather than created, though 🤷 whatevers easiest)
3. We add a Drop impl to our stream struct that sends a message on the channel
4. At the end of each event loop run we look in this channel and create "i'm a dead stream" messages to pass up to the shell

There's probably some details that'd need worked out, but I think the approach would work

fraschm1998:
Does #279 close this issue and if so will it also update on the roadmap? https://github.com/orgs/redbadger/projects/11

charypar:
It does not - it only adds cancellation to timers and makes it an explicit part of the time capability protocol. This issue is about supporting this for _all_ effects without the capability having to explicitly implement the cancellation semantics.

charypar:
Since the [command abstraction](https://github.com/redbadger/crux/pull/291) has been released, the basic cancellation from Core is supported by aborting commands. Resolving aborted effect request is a no-op.

Rust shells can cancel effects by dropping the associated `crux_core::Request`, which immediately aborts their associated task (since they can't be resolved and continued any more).

The only missing piece is support in the `Bridge` for cancelling effects by ID.

It's worth saying that for most capabilities, this basic support may not be quite enough, as it doesn't easily allow for resource cleanup on the shell side when effects are cancelled. 

An example of how that might work is the [new implementation in `crux_time`](https://github.com/redbadger/crux/pull/312). I'm not sure whether there's a generic solution which would allow for this, since the information that needs to be passed alongside the cancellation request is capabilty specific.

PR #292, opened 2025-01-06
cancelling a timer also aborts the cancelled task

adwhit:


crux_time/src/lib.rs
```diff
@@ -162,23 +176,93 @@ where
...
+            // short-circuit return
+            return Poll::Ready(TimeResponse::Cleared { id: self.timer_id });
+        };
+        // see if the timer has been cleared
+        let timer_is_cleared = {
+            let mut lock = CLEARED_TIMER_IDS.lock().unwrap();
+            lock.remove(&self.timer_id)
+        };
+        let this = self.project();
+        *this.is_cleared = timer_is_cleared;
```
charypar, line 237: I'd have to try to have specific advice, but I think you might be able to avoid the `pin_project` by `self.deref_mut()` and an `Unpin` trait bound for whatever is complaining about not being guaranteed unpin. 

Was that problematic?

crux_time/tests/time_test.rs
```diff
@@ -249,27 +253,33 @@ mod tests {
...
+        // this is a little strange-looking. We have cleared the timer,
+        // so the in-flight debounce should have resolved. But to force that
+        // to happen, we have to run the app, and the easiest way to do that
+        // is to resolve the original debounce effect with a fake outcome -
+        // which will be ignored in favour of TimeResponse::Cleared
+        let ev = app
+            .resolve(
+                &mut debounce,
+                TimeResponse::DurationElapsed { id: timer_id },
+            )
```
charypar, line 277: A good test that this doesn't cause issues if the shell still fires the timer.

In a way, we could simply not tell the shell about the cancellation and just stop the task like you are doing which will ignore the timer when it fires.

charypar (commented):
That's quite a clever way to get cancellation, wonder if it could be generalised for #247 somehow.

Bar the question about the need for pin_project, I think I'm happy to ship it.

crux_time/src/lib.rs
adwhit: Have removed `pin_project`, you're right, it wasn't necessary in this instance.

charypar (approved):
Great, thank you!

PR #296, opened 2025-01-21
http capability: add body_form to http request builder

andrew-goldie:
For the http capability... Following the same pattern as constructing other body types (body as bytes plus specific content-type header), this little change adds the ability to create a body for posting forms. Any name-value pair structure that can be serialised and url-encoded can be used as the data type. The doc comment example uses a HashMap for instance.


StuartHarris (approved):
Thanks @andrew-goldie! ❤️

PR #297, opened 2025-01-21
Command support in Core

charypar:
This makes the necessary changes to make Commands usable with `Core` and `AppTester`.

It makes a fair few breaking changes to type signatures, but all quite mechanical and simple. I'll follow this up with a migration guide which people can follow.


StuartHarris (approved):
What a thing of beauty!

PR #301, opened 2025-01-22
Command support in crux_kv and tweaks and fixes to the Command API to enable the capability migration

charypar:
This adds a new module called`command` to `crux_kv` which provides KV command builder constructors.

Because KV presents a different interface to the core than the protocol it has with the shell, it needs to do post-processing_ on the results. I took the opportunity to add `.map` to `RequestBuilder` and `StreamBuilder`. I also made some missed out parts of this API public in `crux_core`. 

This should put is in a good place to tackle more complicated capabilities.

**note**: this will require a patch release of `crux_core` and therefore a patch release of `crux_kv` (and potentially all capabilities - haven't thought about that too hard)

### DIFF
--- a/crux_core/src/command/builder.rs
+++ b/crux_core/src/command/builder.rs
@@ -33,6 +33,13 @@ where
         RequestBuilder { make_task }
     }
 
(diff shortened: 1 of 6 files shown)


StuartHarris (approved):
❤️

PR #303, opened 2025-01-23
`Command` support in `crux_http`

StuartHarris:


crux_http/tests/with_tester.rs
```diff
@@ -48,76 +48,66 @@ mod shared {
...
                 Event::PostForm => {
                     let form = HashMap::from([("key", "value")]);
-                    caps.http
-                        .post("http://example.com")
+                    command::Http::post("http://example.com")
                         .body_form(&form)
                         .expect("could not serialize form data")
                         .expect_string()
-                        .send(Event::Set);
+                        .build()
```
charypar, line 70: Maybe `send` or `submit` or something more HTTP-like? Not sure, maybe not 😅

charypar (commented):
Rather than changing the tests in place, should we add the changed one? We probably still want regression coverage for the original API for now

PR #304, opened 2025-01-24
update cat_facts and counter examples to use crux_http::command

StuartHarris:


examples/cat_facts/shared/src/app.rs
```diff
@@ -126,89 +127,97 @@ impl App for CatFacts {
...
+                    Command::done()
+                }
+            },
             Event::SetState(Ok(None)) => {
                 // no state to restore
+                Command::done()
             }
             Event::SetState(Err(_)) => {
                 // handle error
+                Command::done()
```
charypar, line 217: Nit: lots of these branches return `Command::done()` might be easier to do early returns, but I don't mind either way.

charypar (approved):

examples/cat_facts/shared/src/app.rs
StuartHarris: I kinda wanted them to be markers for "something still to do here"

PR #312, opened 2025-01-29
Command interface for crux_time

charypar:
As part of this, I've revised how the cancellation works so that idea of IDs doesn't leak into the app.

### DIFF
--- a/crux_time/src/command.rs
+++ b/crux_time/src/command.rs
@@ -33,20 +33,15 @@ where
         let (sender, mut receiver) = oneshot::channel();
 
         let builder = RequestBuilder::new(move |ctx| async move {
-            // We loop so that we can ignore the Err case of the oneshot receiver and still wait for the original request
-            loop {
-                select! {
-                    response = ctx.request_from_shell(TimeRequest::NotifyAt { id, instant }).fuse() => return response,
-                    cleared = receiver => {
-                        match cleared {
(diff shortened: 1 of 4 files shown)


crux_time/src/command.rs
```diff
@@ -0,0 +1,206 @@
...
+        instant: Instant,
+    ) -> (
+        RequestBuilder<Effect, Event, impl Future<Output = TimeResponse>>,
+        TimerHandle,
+    ) {
+        let id = get_timer_id();
+        let (sender, mut receiver) = oneshot::channel();
+
+        let builder = RequestBuilder::new(move |ctx| async move {
+            // We loop so that we can ignore the Err case of the oneshot receiver and still wait for the original request
```
adwhit, line 36: What are the semantics here? If the sender is dropped, will the `select!` still block on the receiver (never resolving) or will the receiver immediately resolve to `Err(_)`, causing a busy loop?

charypar: From some testing I've done, dropping the `Sender` does not wake the future, so I think the `Err` branch of the cancel channel select is actually unreachable 🤔 I think it's because the `Receiver` is a `FusedFuture` and slect checks it's termination status.

Might change the code to remove the loop and put `unreachable` in there

StuartHarris (approved):

PR #317, opened 2025-02-09
Experiment: Allow Core to be wrapped in a middleware

charypar:
This is a bit work in progress, but it's a start of a potential approach to allowing "hybrid" shells - part written in Rust, part written in Kotlin/Swift/TypeScript... 

The main change is an introduction of a `crux_core::Middleware` trait, which is implemented by the `Core` and expected by the `Bridge` to be implemented by the provided core type. This allows middlewares to be inserted between the Bridge and the Core. Because the middleware may stack up, I've also switched the Core to return an iterator over effects, to avoid unnecessary allocations in case the middleware chain ends up long.

As part of the experiment I've implemented a Key-Value capability using a HashMap for the storage, which catches the KV effects and resolves them against the store. It turns out it's pretty involved, because resolving he effects may produce follow up effects which need to also be potentially processed and so on, creating a mutually recursive loop. 

Doing that with existing iterators turned out quite difficult, so I've resorted to an odd specialised queueing FlatMap sort of thing, which seems like it might be a beginning of some more helpful abstraction.

I'll keep this open for now so that folks can experiment and see how it works out for them and we can iterate on it before we land on something solid.

Note that this would be a breaking change, so I'd like us to get it right, rather than releasing five breaking iterations 😓

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,6 +367,7 @@ dependencies = [
  "bincode",
  "crossbeam-channel",
  "crux_http",
+ "crux_kv",
  "crux_macros",
  "crux_time",
  "doctest_support",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,6 +370,7 @@ dependencies = [
  "crux_macros",
  "crux_time",
  "doctest_support",
+ "either",
  "erased-serde",
  "futures",
  "rand 0.8.5",
@@ -553,6 +554,12 @@ dependencies = [
  "serde",
  "serde_json",
 ]
+
+[[package]]
+name = "either"
+version = "1.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "60b1af1c220855b6ceac025d3f6ecdd2b7c4894bfe9cd9bda4fbb4bc7c0d4cf0"
 
 [[package]]
 name = "encode_unicode"
--- a/crux_core/Cargo.toml
+++ b/crux_core/Cargo.toml
@@ -33,6 +33,7 @@ thiserror = "2.0.11"
 # So I've left them 0.26.0 and 0.4.0 respectively.
 serde-generate = { version = "0.26.0", optional = true }
 serde-reflection = { version = "0.4.0", optional = true }
+either = "1.13.0"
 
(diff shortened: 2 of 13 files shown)


charypar:
After a few experiments people have done and a bit more thinking about this, I suspect this is not the way forward, or at least not the way to have _some_ capabilities implemented in Rust, whilst others are in the actual shell.

The problem si control flow. What having middleware handle effects amounts to is effectively having more than one shell. And since the shell is driving the execution, that doesn't work. The problem is best illustrated with a scenario:

1. Shell sends an event to the core
2. Core processes this and responds with effect request A and B
3. Middleware A takes effect A and starts processing it on another thread, passes B along to the shell. The original call from step 1 completes and returns control to the shell.
4. Shell receives B and starts processing it
5. Effect A completes on the background thread, the middleware resolves it (so inside the core its associated future is now ready to proceed). The now advances the effect runtime by calling `process_effects`, and, for the sake of the scenario, receives another effect C of the same type as B. What does it do now? 

The middleware has no way to handle effect C now. It doesn't know how to process it, only the shell does. It can't pass it along, because it is not in the context of being called by the shell this time. 

The various ways to handle this are all imperfect. Something needs to get queued up somewhere - either the resolution of effect A (or the subsequent `process_effect` call), or the passing along of the follow-up effect C. No matter which way one goes, the loop will only continue next time the shell si involved.

This is fundamental to the execution model of Crux - the core is assumed to be _synchronous_. And this effectively changes the call into an asynchronous one (even though it returns early).

Another way of looking at this is that introducing middleware that can execute effects has the same difficulties the shell would face if it wanted a straight up asynchronous FFI into Rust: The return from the call at step 1 is the same as a yield (or `.await`), and the core would need to get polled again somehow, once effect B is ready to proceed, so that any followups could be returned to the shell. And to make this efficient, there'd need to be a waking mechanism telling the shell the core is ready to be polled again.

StuartHarris:
Great catch! Crux is passive and must be driven by the shell. Shall we close this and create examples/update docs to show how to achieve the same thing with a tiny bit of shell-side coordination and async calls into a Rust capability?

PR #324, opened 2025-02-26
Use `std::time::Duration` and `std::time::SystemTime` in the `crux_time` API

StuartHarris:
**Note: this is a breaking change**

Changes the `crux_time` API to use `std::time::Duration` and `std::time::SystemTime` instead of `crux_time::Duration` and `crux_time::Instant` (although those types remain and are used for serialization across the FFI boundary). Conversions between the `crux-time` types and the `std::time` types have been added (most useful in tests where we are checking requests and responses).
Also updates `cat_facts` that had a bug where the time wasn't being correctly fetched in the `web-nextjs` shell.
Also fixes a few clippy warnings etc.


adwhit:
Maybe worth doing a brief survey of `chrono::Duration` and `time::Duration` to see what the differences are, if any?

StuartHarris:
> Maybe worth doing a brief survey of `chrono::Duration` and `time::Duration` to see what the differences are, if any?

I kinda don't want to enforce that dependency. We already support chrono conversions under the `chrono` feature, which allows opt-in if users are already using chrono..

crux_core/src/bridge/request_serde.rs
```diff
@@ -87,7 +87,8 @@ impl<Out> Resolve<Out> {
             Resolve::Never => ResolveSerialized::Never,
             Resolve::Once(resolve) => ResolveSerialized::Once(Box::new(move |deser| {
                 let out = func(deser)?;
-                Ok(resolve(out))
+                resolve(out);
+                Ok(())
```
charypar, line 91: Do we need to align the error handling across the `Once` and `Many`? It's quite suspicious that one returns its `resolve` return and the other doesn't... 🤔 

I vaguely remember needing to mess with this for the error handling changes recently, but now I can't remember the details. But I am a little concerned about changing this - can't immediately see what the consequences are.

crux_time/src/command.rs
```diff
@@ -126,7 +153,7 @@ mod tests {
 
     #[test]
     fn timer_can_be_cleared() {
-        let (cmd, handle) = Time::notify_after(Duration::from_secs(2).unwrap());
+        let (cmd, handle) = Time::notify_after(Duration::from_secs(2));
```
charypar, line 156: Yay! 🥳

crux_time/src/lib.rs
```diff
@@ -286,7 +268,7 @@ mod test {
 
         let now = TimeRequest::NotifyAfter {
             id: TimerId(2),
-            duration: Duration::from_secs(1).expect("valid duration"),
+            duration: crate::Duration::from_secs(1).unwrap(),
```
charypar, line 271: Ah annoying - our own Duration still needs unwrapping 😕 so in tests it's still a bit of a pain... I wonder if we should just panic on overflow instead - nobody is actually going to put numbers large enough in there anyway... probably? 😅

charypar (approved):

crux_core/src/bridge/request_serde.rs
StuartHarris: That resolve returns the unit type, it's only `ResolveMany` that returns a result (to carry an error from trying to resolve when already finished). So nothing changes. Just fixes a clippy warning.

crux_time/src/lib.rs
StuartHarris: yeah. Good idea. will do that

StuartHarris: I did that and tidied up a bunch of other stuff (including fixing a big in yew shell for cat_facts). If you could review again, that'd be great — thanks https://github.com/redbadger/crux/pull/324/commits/f0c760289b9269fd72a6514cd107adf71a6a41be

crux_core/src/bridge/request_serde.rs
charypar: Alright. I guess we need to revisit whether it _should_ return a result, but not in this PR.

crux_time/src/protocol/instant.rs
```diff
@@ -12,27 +10,27 @@ const NANOS_PER_SEC: u32 = 1_000_000_000;
 /// - seconds: number of seconds since the Unix epoch (1970-01-01T00:00:00Z)
 /// - nanos: number of nanoseconds since the last second
 ///
-/// Note: We cannot use `std::time::Instant` because it is not deserializable.
+/// Note: We cannot use `std::time::Instant` because it is opaque and not deserializable.
```
charypar, line 13: And also semantically not the right thing, if we want to get pedantic 😅

charypar (approved):

crux_time/src/protocol/instant.rs
StuartHarris: Yeah. Good point. I removed the comment altogether

PR #331, opened 2025-03-20
Allow ShellRequest future to be polled many times after the request is dropped.

charypar:
This one is quite hard to explain.

`ShellRequest` is constructed on top of a `ShellStream`, and we use [`StreamFuture` from futures](https://docs.rs/futures/latest/futures/prelude/stream/struct.StreamFuture.html) to represent this. 

The peculiar thing about `ShellRequest` is that it panics if it's polled again after it returned Ready. Logically, this shouldn't happen in our case - the stream is actually a future, it will resolve at most once. The problem is the "at most" part. If it never resolves and the shell side of it (the `Request`) is cleaned up, the channel underneath closes and the stream terminates without ever producing a value. 

In this case `ShellRequest` returns `Pending` when polled, and is immdiately discovered by the check for wakers - there are no more wakers and the task is evicted. Unfortunately in the case of timer requests, there's also the cancelation handle holding another waker for the task, and so it survives.

If we then poll it, we attempt to poll the `StreamFuture` again and panic. The solution is to create a fresh `StreamFuture` every time we get polled in such scenario.

### DIFF
--- a/crux_core/src/command/context.rs
+++ b/crux_core/src/command/context.rs
@@ -163,8 +163,9 @@ impl<T: Unpin + Send> ShellStream<T> {
         // Since neither part is Clone, we'll need to do an Indiana Jones
 
         // 1. take items out of self
+        let dummy = ShellStream::Sent(mpsc::unbounded().1);
         let ShellStream::ReadyToSend(send_request, output_receiver) =
-            std::mem::replace(self, ShellStream::Sent(mpsc::unbounded().1))
+            std::mem::replace(self, dummy)
         else {
             unreachable!();
         };
@@ -215,7 +216,14 @@ impl<T: Unpin + Send> Future for ShellRequest<T> {
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.inner.poll_unpin(cx) {
             Poll::Ready((Some(output), _rest)) => Poll::Ready(output),
-            Poll::Ready((None, _rest)) => Poll::Pending,
+            Poll::Ready((None, rest)) => {
+                // We need to make sure that the inner future can be polled again,
+                // because we could be used in a select! and polled repeatedly, even though
+                // the request on the shell side is a goner.
+                self.inner = rest.into_future();
+
+                Poll::Pending
+            }
             Poll::Pending => Poll::Pending,
         }
     }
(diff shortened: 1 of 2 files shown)


crux_time/src/command.rs
```diff
@@ -313,4 +316,18 @@ mod tests {
...
+        let (cmd, handle) = Time::notify_after(Duration::from_secs(2));
+        let mut cmd = cmd.then_send(Event::Elapsed);
+
+        let effect: Effect = cmd.effects().next().expect("Expected an effect!");
+
+        drop(effect);
+        assert!(!cmd.is_done());
+
+        drop(handle);
+        assert!(cmd.is_done());
```
charypar, line 331: This line caused a panic before the fix.

crux_core/src/command/context.rs
```diff
@@ -215,7 +216,14 @@ impl<T: Unpin + Send> Future for ShellRequest<T> {
...
             Poll::Ready((Some(output), _rest)) => Poll::Ready(output),
-            Poll::Ready((None, _rest)) => Poll::Pending,
+            Poll::Ready((None, rest)) => {
+                // We need to make sure that the inner future can be polled again,
+                // because we could be used in a select! and polled repeatedly, even though
+                // the request on the shell side is a goner.
+                self.inner = rest.into_future();
+
+                Poll::Pending
+            }
```
DeltaEvo, line 226: nit: I think this would be clearer by either Fusing the inner future

```
pub struct ShellRequest<T: Unpin + Send> {
    inner: Fuse<StreamFuture<ShellStream<T>>>,
}
```

Or by using the `is_terminated` on `self.inner` to properly show that if we get a `Poll::Ready((None, _rest))` the future will never resolved and stay pending


```
fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
    if self.inner.is_terminated() {
        Poll::Pending
    } else {
        ...
    }
}
```

DeltaEvo (approved):

crux_core/src/command/context.rs
charypar: Fair point! Can't quite decide which one I like better 🤔

charypar: Decided to go with `Fuse<T>`

PR #334, opened 2025-03-26
Effect registration when using `effect!`

StuartHarris:
This PR moves the `Export` trait implementation from `Capabilities` to `Effect` so that we can call `Effect::register_types` from `register_app`.


charypar (approved):
Nice! This is all starting to make a lot of sense :)

PR #326, opened 2025-03-13
Fix path of Xcode-generated “uniffi-bindgen” executable artifact

regexident:
The actual path of uniffi-bindgen's build artifact is `"${BUILD_DIR}/Debug/uniffi-bindgen"`, not `"${BUILD_DIR}/debug/uniffi-bindgen"`.

The latter only happens to work on case-insensitive volumes and with "Build configuration: Debug", afaict.

To reproduce (on `master`):

1. Open "TapToPay" iOS project
2. Clean the project's build folder via "Main Menu" > "Product" > "Clean Build Folder…" (or via <kbd>command</kbd> <kbd>shift</kbd> <kbd>K</kbd>)
3. Build the project via "Main Menu" > "Product" > "Build" (or via <kbd>command</kbd> <kbd>B</kbd>)

The build should succeed. As expected.

Now repeat the steps, but before building the project (step 3) change the project's build configuration from "Debug" to "Release" via "Main Menu" > "Product" > "Scheme" > "Edit Scheme…" (or via <kbd>command</kbd> <kbd>\<</kbd>).

The build now fails with this:

```plain
line 16: …/Build/Products/debug/uniffi-bindgen: No such file or directory
Command RuleScriptExecution failed with a nonzero exit code
```

Now perform the same steps on this branch and both should succeed. 🥳

PR #335, opened 2025-03-27
Fix capitalization of Xcode’s `${CONFIGURATION}` env var

regexident:
This is a post-fix for https://github.com/redbadger/crux/pull/326.

I somehow embarrassingly got the env var's capitalization wrong. 🫣

For off-the-shelf Xcode configs it seems to work just fine —which is why I didn't catch it.
But depending on whether or not your system's shell is case-sensitive it might not.
And a case of the latter was just reported to me. Using `${CONFIGURATION}` fixed it.


StuartHarris (approved):
Amazing, thanks :-) I have been doing the same on the branch I'm working on, but it hasn't been merged yet, so glad to get this in first.

PR #336, opened 2025-03-31
improve effect macro and docs

StuartHarris:
Some very minor tweaks to the `effect!` macro (that includes a better error message), and an update to the `getting started` guide in the book.

### DIFF
--- a/docs/src/getting_started/core.md
+++ b/docs/src/getting_started/core.md
@@ -5,7 +5,7 @@ itself, and the shared types crate which does type generation for the foreign
 languages.
 
 ```admonish warning title="Sharp edge"
-Most of these steps are going to be automated in future tooling, and published as crates. For now the set up is effectively a copy & paste from one of the [example projects](https://github.com/redbadger/crux/tree/master/examples).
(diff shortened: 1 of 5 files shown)


docs/src/getting_started/core.md
```diff
@@ -157,12 +162,28 @@ well as our public types from the shared library.
...
+The macro generates code for:
+- an _actual_ effect enum (where the inner `Operation` type is wrapped with [`crux_core::Request`](https://docs.rs/crux_core/latest/crux_core/struct.Request.html))
+- an enum, similar to the one we specify, for use across the FFI boundary,
+- an implementation of the `Effect` trait
+- some useful test helpers
+- code to support the current foreign type generation (that uses `serde-reflection`)
+
+You can also use the `effect!` macro with the [new typegen](https://github.com/redbadger/crux/pull/217).
+```
+
```
charypar, line 186: Since this is a Getting Started, should we just speak to the way people should be doing it now, if they're coming to it fresh? So that means commenting on the reason there is a `Capabilities` associated type (tail end of migration), rather than pointing out the `effect!` macro, I think?

StuartHarris: Good point. Added in several places.

charypar (approved):

PR #337, opened 2025-03-31
Android build fixups / documentation

seam345:
This is kinda 3 issues in 1, happy to break them up if wanted

1. Slight tweak in when type and bind gen happen
    - Fairly confident this has no negative side effects
2. Added workaround for the gradle plugin
    - I'm a bit unsure if we want to add workarounds to the examples or not 
3. Started documenting shrinking the APK for android builds
    - Probably want to reword most of it to fit, i also haven't got round to building the docs myself so i may have linked it wrong in summery


StuartHarris (approved):
Thanks @seam345 ❤️ I fixed a couple of typos. Also I will update our CSS so that mdbook can render the diffs properly in a separate PR.

PR #338, opened 2025-04-07
NotificationBuilder

StuartHarris:
**Breaking Change**: Command API changed — `notify_shell` now returns `NotificationBuilder`, instead of `Command`

Adds `NotificationBuilder` which has symmetry with `RequestBuilder` and `StreamBuilder`, but without continuations. This is useful where you need to notify the shell from an async context. There is also now a `From` implementation to convert the builder into a command. 

Updated the `Render` capability accordingly. 

Note: it is recommended to implement capabilities in terms of builders rather than commands, so that they can be used in both sync and async contexts. However, `render::render()` currently returns a command, so we kept it that way to minimise breakages, and instead introduced `render::render_builder()` to return a builder.

### DIFF
--- a/crux_core/src/capabilities/render.rs
+++ b/crux_core/src/capabilities/render.rs
@@ -1,9 +1,12 @@
 //! Built-in capability used to notify the Shell that a UI update is necessary.
 
+use std::future::Future;
+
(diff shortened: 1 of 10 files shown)


charypar (approved):
:shipit:

PR #339, opened 2025-04-07
Fix an executor bug that evicts tasks which wake themselves immediately and drop the waker

charypar:
This fixes a subtle bug which can drop longer-running tasks to aggressively, when they return `Pending` from poll, but have already marked themselves ready and dropped the waker.

To handle the exception, I had to slightly rework how we process the ready queue, adding an explicit buffer which can be scanned for the id of a task which is suspended with no wakers, to see if it's also ready.

### DIFF
--- a/crux_core/src/command/executor.rs
+++ b/crux_core/src/command/executor.rs
@@ -15,7 +15,7 @@ use futures::task::AtomicWaker;
 
 use std::sync::Arc;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) struct TaskId(pub(crate) usize);
 
 pub(crate) struct Task {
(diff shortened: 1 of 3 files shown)


crux_core/src/command/executor.rs
```diff
@@ -158,25 +159,48 @@ impl<Effect, Event> Command<Effect, Event> {
...
             }
 
-            while let Ok(task_id) = self.ready_queue.try_recv() {
-                match self.run_task(task_id) {
+            let mut ready_tasks: VecDeque<_> = self.ready_queue.try_iter().collect();
+
+            while !ready_tasks.is_empty() {
+                let task_id = ready_tasks
+                    .pop_front()
+                    .expect("ready_task should not be empty!");
```
DeltaEvo, line 167: nit:

```suggestion
            while let Some(task_id) = ready_tasks.pop_front() {
```

DeltaEvo (approved):

StuartHarris (approved):

crux_core/src/command/executor.rs
```diff
@@ -177,7 +179,7 @@ impl<Effect, Event> Command<Effect, Event> {
 
                         drop(task);
                     }
-                }
+                };
```
DeltaEvo, line 182: nit: Remove semicolon ?

DeltaEvo (approved):

PR #343, opened 2025-04-09
`effect!` macro becomes `#[effect]` or `#[effect(typegen)]`

### LINKED ISSUE #342: `effect!` macro expects a `typegen` feature flag in the core crate.
If an app doesn't use Crux's type generation, it is forced to add this feature flag even though it doesn't need it.

StuartHarris:
Fixes #342 

This PR introduces a **BREAKING CHANGE**.

The `effect!` function macro has been replaced with an `#[effect]` attribute proc macro.

It also expects you to opt into foreign type generation using `#[effect(typegen)]` attribute on your Effect enum.  

This is to allow the macro to be used in applications that either don't use the builtin typegen _or_  use the new typegen (from #217) by using `#[effect]` (without any arguments).

e.g. with typegen:

```rust
#[effect(typegen)]
pub enum Effect {
    Render(RenderOperation),
    Http(HttpRequest),
    ServerSentEvents(SseRequest)
}
```

or, without typegen:

```rust
#[effect]
pub enum Effect {
    Render(RenderOperation),
    Http(HttpRequest),
    ServerSentEvents(SseRequest)
}
```


charypar (approved):
'utiful!

PR #345, opened 2025-04-10
update guide for hello world

StuartHarris:
This updates the ["Hello World" page](https://redbadger.github.io/crux/guide/hello_world.html) in the development guide section of the book to be compatible with the `#[effect]` macro.

It also updates the `hello-world` example, from which code snippets are drawn.


charypar (approved):
Noice.

PR #347, opened 2025-04-11
Add a new `counter-next` example for new typegen (no `shared_types` crate)

StuartHarris:
Adds a `counter-next` example, which has no `shared_types` crate, in order to see how DX is improved with the new typegen and bindings generation, and an improved CLI.

This PR should be safe to merge and shouldn't affect any existing Crux apps.

- [x] add `bindgen` command to `crux_cli` to generate bindings for Swift and Kotlin
- [x] re-export `crux_cli` in `crux_core` under feature `cli`, so that users build a crate-local binary of the right version
- [x] fix bug in typegen where TypeScript was being installed as a dependency, rather than a devDependency.
- [x] unify flag names between `codegen` and `bindgen` commands
- [x] introduce `counter-next` example without `shared_types` crate
- [x] add `crux_cli` binary to `shared`
- [x] add script (for now) to do both codegen and bindgen
- [x] switch to uniffi proc macros and remove udl file
- [x] remove shells other than Android, iOS, web/TS and web/Rust
- [x] update Android gradle files to use generated code
- [x] update XCode build to use generated code
- [x] ensure all shells work properly
- [x] lock uniffi versions so that the version used by `shared` is the same as that used by `crux_cli` bin


StuartHarris:
@srfsh It probably will be merged soon, but not released just yet. There is some way to go before we would recommend it as part of the build process for a Crux app. There are some outstanding issues, a migration plan to think about and a fair amount of confidence building to do. — It would be wonderful if you wanted to try it out and feedback any issues you might find 😀

crux_cli/src/bindgen/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::args::BindgenArgs;
+
+pub fn bindgen(args: &BindgenArgs) -> Result<()> {
+    let status = Command::new("cargo")
+        .args(["build", "--package", &args.crate_name])
+        .status()?;
+    assert!(status.success());
+
+    let library_path = Utf8Path::new("target/debug/libshared.dylib");
```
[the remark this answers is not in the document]
StuartHarris: Great catch! I missed that. Thank you! I need to improve this, but won't have time to look properly until early next week, but will catch up then. Feel free to stack a PR on top of this if you want, otherwise I'll get to it as soon as I can. Thanks

StuartHarris:
> I have a few patches applied locally that changes the suggestions I mentioned. I would like to upstream these if that's okay. There are a few additional things I would like to propose as well. I will list them here.
> 
> * I think we can directly embed which crate to use for generation inside our crux_cli.rs. This way, you'd not need to specify `--crate-name`. Since it is already in the shared crate, I think it only makes sense and allows for better DX. I didn't test this (I will), but I am pretty sure it is possible using the `CARGO_*` variables.

This could work. We need to ensure the cli works if run independently from the shared lib (e.g. if installed with cargo install).

> * Since we got rid of the udl file for a better DX using macros, I think we can get rid of `uniffi.toml` that is used not much as well. I figured out a way using [the `BindgenCrateConfigSupplier` trait](https://docs.rs/uniffi_bindgen/latest/uniffi_bindgen/trait.BindgenCrateConfigSupplier.html) . I would really love this to go in, but I can understand if this would be rejected as it hides this from the developer. I think it is better if this information is provided in Cargo.toml as a metadata.

What would be nice is to let the user choose how to specify the config, with levels of preference for where the values come from.

> 
> Tell me what you think. And thanks! I will be testing these with my changes.
> 
> P.S.: One might wonder why I had a change of heart all of a sudden. It is really simple: I figured I can leverage uniffi to fill in the gaps crux being pure. :)

❤️

examples/counter-next/shared/.gitignore
```diff
@@ -0,0 +1,4 @@
+!/generated
```
charypar, line 1: Are we intentionally comitting the generated code? 🤔

crux_macros/src/bridge.rs
```diff
@@ -0,0 +1,107 @@
...
+        #input
+
+        static CORE: ::std::sync::LazyLock<::crux_core::bridge::Bridge<#ident>> =
+            ::std::sync::LazyLock::new(|| {
+                ::crux_core::bridge::Bridge::new(::crux_core::Core::new())
+            });
+
+        #[cfg(not(target_family = "wasm"))]
+        const _: () = assert!(
+            uniffi::check_compatible_version("0.29.2"),
```
charypar, line 18: Elsewhere we do `::uniffi` - does this make the macro potentially leaky if someone redefines `uniffi`?

crux_macros/src/bridge.rs
```diff
@@ -0,0 +1,107 @@
...
+        ::uniffi::setup_scaffolding!();
+
+        #[cfg_attr(not(target_family = "wasm"), ::uniffi::export)]
+        #[cfg_attr(target_family = "wasm", ::wasm_bindgen::prelude::wasm_bindgen)]
+        pub fn process_event(data: &[u8]) -> Vec<u8> {
+            match CORE.process_event(data) {
+                Ok(effects) => effects,
+                Err(e) => panic!("{e}"),
+            }
+        }
```
charypar, line 32: More of a uniffi noob question: Are the exports indpendent of module path? Does it matter where the macro is used?

And am I assuming correctly that this can only be used once per crate?

crux_macros/src/bridge.rs
```diff
@@ -0,0 +1,107 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::ItemStruct;
+
+pub fn bridge_impl(input: ItemStruct) -> TokenStream {
+    let ident = &input.ident;
+    quote! {
+        #[derive(Default)]
```
charypar, line 8: I don't think the derive is guaranteed to work for all `Apps` - some may have fields with non-derivable `Default` and will need to hand implement `Default`, we probably should leave writing this line up to the user

charypar (commented):
A few questions, but no major problems I can see, other than the `derive(Default)`. 

The main thing that worries me about this is that it takes on the job of making the core instance, which is where the middleware stacking will go in the near future. I'm not super sure what the syntax for doing that with an attribute macro would look like 🤔 It could be as simple as allowing the user to opt out of that part - adding an override for the name of the static variable holding the core (e.g. `#[bridge(core_instance = CORE)]`), which would stop generating that bit of code and rely on it being defined already.

It also slightly worries me that it does a fair bit o magic by convention. I'd be more comfortable to commit to a `#[bridge]` macro or `#[export]` macro on individiaul functions (which would setup the `uniffi` and `wasm` annotations). Because it's
1. More explicit - the names of the functions and arguments are more visible and less magic
2. Useful to export any additional FFI calls in the same way, which is sometimes a handy thing to do

crux_macros/src/bridge.rs
StuartHarris: thanks. I missed this. Done now.

examples/counter-next/shared/.gitignore
StuartHarris: as per our convo, they're no longer comitted

StuartHarris:
> * I think we can directly embed which crate to use for generation inside our crux_cli.rs. This way, you'd not need to specify `--crate-name`. Since it is already in the shared crate, I think it only makes sense and allows for better DX. I didn't test this (I will), but I am pretty sure it is possible using the `CARGO_*` variables.

Thanks for this suggestion. The `--crate-name` arg can now be read from `CRATE_NAME` env var, so the calling code can look like this:
```rust
    env::set_var("CRATE_NAME", env!("CARGO_PKG_NAME"));
    crux_core::run_cli()
```

crux_cli/src/args.rs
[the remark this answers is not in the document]
StuartHarris: Thanks @srfsh — the `*-package` args are now in a group that requires at least one to be specified. Code will only be generated for the ones you specify.

StuartHarris:
This PR is quite big now. I'll investigate the `uniffi.toml` config in another PR. This should be good to review now.

examples/counter-next/shared/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+edition = "2021"
+rust-version = "1.80"
+
+[lib]
+crate-type = ["lib", "staticlib", "cdylib"]
+name = "shared"
+
+[[bin]]
+name = "crux_cli"
+required-features = ["cli"]
```
charypar, line 13: Does this make the building of the bin target conditional?

examples/counter-next/shared/src/bin/crux_cli.rs
```diff
@@ -0,0 +1,8 @@
+use std::env;
+
+use uniffi::deps::anyhow;
```
charypar, line 3: Does the core still need to depend on `uniffi` anyway or can it depend on `uniffi` through `crux_core` with the right flags? 🤔 Not really a blocker to merge this, just wondering how far we can take the taking ownership / hiding of our underlying tools of choice.

examples/counter-next/shared/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+/// # Panics
+/// If the view cannot be serialized
+#[cfg_attr(not(target_family = "wasm"), uniffi::export)]
+#[cfg_attr(target_family = "wasm", wasm_bindgen::prelude::wasm_bindgen)]
+pub fn view() -> Vec<u8> {
+    match CORE.view() {
+        Ok(view) => view,
+        Err(e) => panic!("{e}"),
+    }
+}
```
charypar, line 57: Ah, answers my other questions - this is where it still depends on uniffi directly.

examples/counter-next/shared/uniffi.toml
```diff
@@ -0,0 +1,7 @@
+[bindings.kotlin]
+package_name = "com.crux.example.counter.shared"
+cdylib_name = "shared"
+
+[bindings.swift]
+cdylib_name = "shared_ffi"
+omit_argument_labels = true
```
charypar, line 7: More out of curiosity - does this still get read by the uniffi implementation which the crux_cli is wrapping?

charypar (approved):
Bunch of questions out of curiosity, but none blocking merging this.

Send it 🚀

examples/counter-next/shared/Cargo.toml
StuartHarris: It means we have to build it with the `cli` feature, which, in turn, brings in the "cli" feature from `crux_core`, which in turn re-exports from `crux_cli` 😓

so:

```bash
cargo run --package shared --bin crux_cli --features cli -- --help
```

examples/counter-next/shared/uniffi.toml
StuartHarris: I think so. But there are some helpful comments above (by @srfsh) on how we can avoid this. I'm going to visit this configuration and how we best provide it in a new PR

PR #348, opened 2025-04-11
remove the unused and deprecated `multilingual` field from `book.toml`

szabgab:
See https://github.com/szabgab/public-mdbooks/issues/10


StuartHarris (approved):
Thanks @szabgab ❤️

Issue #164, opened 2023-12-03
Thoughts on Windows 11 C# Windows App SDK WinUI 3 XAML


kaidelorenzo:
I'm wondering about 2 things

1. Are there plans to build out C#/WinUI 3 support?
2. Is that something that folks would get behind integrating into Crux?

charypar:
Hi @kaidelorenzo! Thanks for the questions.

1. Not at present. We don't have a use for it, and it would likely depend on #145 
2. We would definitely welcome your thoughts on what if anything should be supported (beyond C# typegen) and contributions towards it.

There are a few things that depend on the custom type gen in #145, so it's relatively likely we'll start investigating options there pretty soon.

kaidelorenzo:
I haven't given much thought to what limitations would be acceptable. I've read online that the interop between c# and rust isn't great right now so that might make things more challenge.

My goal would probably be minimal typegen support so that it is possible to run the framework even if the data types sent are simple (maybe restrict to integer or string structs only).

I don't have much time right now to work on this but glad that you'd be open to contributions. To me Windows seems like the last native platform that doesn't already work with crux.

charypar:
One problem for which I can't see an immediate solution is representing Rust enum's, which are quite central to how Crux communicates with the shells (for events and effects). We'd need to do something like this probably https://bugsquash.blogspot.com/2012/01/encoding-algebraic-data-types-in-c.html. I'm not very familiar with the windows ecosystem, but maybe it would make more sense to target F# and rely on its interop with C# if that's more mature? 

@StuartHarris Do you know much about the state of interop in C# land?

StuartHarris:
I don't. But I'm putting this link here mainly so I can find it again to do a bit more reading on it. https://github.com/FractalFir/rustc_codegen_clr

huwaireb:
👋

with windows-rs and WinRT components _you_ can integrate Rust into a C#/WinUI 3 or any WinRT based application.
I've done it the other day, works fine, just have to write an interface using [MIDL 3](https://learn.microsoft.com/en-us/uwp/midl-3/intro) and then windows-rs has a bindgen that generates bindings, of which you implement for, [samples](https://github.com/microsoft/windows-rs/tree/master/crates/samples/components) (c++ no official c# as of yet).

it's pretty simple outside of having to conform to the quirks of WinRT on implementation, works well.

[...]

PR #351, opened 2025-04-23
C# Support and WinUI 3 Simple Counter

kaidelorenzo:
This uses [uniffi-bindgen-cs](https://github.com/NordSecurity/uniffi-bindgen-cs) to add C# support for Crux type generation and interop.

There is a Simple Counter integrated into the examples folder.

I've only tested the example setup with Visual Studio though I imagine it could work with other IDEs.

The example is based off of the Visual Studio template "Blank App, Packaged (WinUI 3 Desktop)".

I haven't tested publishing to the Microsoft Store or installing on other devices, but I imagine it would work. 

Fixes https://github.com/redbadger/crux/issues/164

I plan to work on some updates to the Crux Book in time.


StuartHarris:
Hey @kaidelorenzo thanks for this! It'll be great to have a C# example (although not having a windows machine might make it hard for me to test). Anyway, CI is failing, if you get a chance to look. Otherwise I can later. 👍

examples/simple_counter/Windows/SimpleCounter/MainWindow.xaml.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            this.InitializeComponent();
+            _core = new Core();
+            UpdateUI();
+        }
+
+        private void UpdateUI()
+        {
+            // Update the UI based on the ViewModel
+            CounterText.Text = _core.View.count;
```
mhedgpeth, line 40: I'm curious if you found a way to bind to the ViewModel from XAML? I think that would mean the generated type would implement `INotifyPropertyChanged`? In your mind would an app at scale just manually copy from the view to the data elements?

kaidelorenzo: Yeah that's a great question. I think something along the lines of creating an observable object should be possible. I'm honestly not very experienced with the framework but I'll plan to update the example the example once I figure something better out.

kaidelorenzo:
The build is failing because the environment doesn't have dotnet. I've currently setup the build.rs in the example to compile the generated project into a dll to reference from C#. Potentially that should move to the C# project build step but for whatever reason I couldn't get it to work nicely. I'll take another look to see if that would make sense

examples/simple_counter/Windows/SimpleCounter/MainWindow.xaml.cs
kaidelorenzo: I added what seems to me to be an idiomatic change listener.

kaidelorenzo:
Though now that I think about it. Crux doesn't have a direct dependency on uniffi? This might not actually be as much of an issue as I thought...

mknet:
I am just trying to get the [rebased version](https://github.com/mknet/crux-windows-example/commit/4196bdc88dad778e577d490e2d99a2999582bd0a) running. It compiles but won't start. If you got 15 mins in the next couple of days we could try it together. I also pinged you on crux's Zulip.

examples/simple_counter/Windows/LICENSE.txt
charypar, line 1: Is this included intentionally?

examples/simple_counter/shared/Cargo.toml
```diff
@@ -15,11 +15,12 @@ typegen = ["crux_core/typegen"]
 crux_core.workspace = true
 serde = { workspace = true, features = ["derive"] }
 lazy_static = "1.5.0"
-uniffi = "0.29.1"
+uniffi = "0.28"
+uniffi-bindgen-cs = { git = "https://github.com/NordSecurity/uniffi-bindgen-cs", tag = "v0.9.1+v0.28.3", version = "0.9.0" }
```
charypar, line 19: I saw this mentioned in the previous comments - does `uniffi-bindgen-cs` not work with `uniffi` 0.29?

If that's the case - do we think that's fixable and we can contribute the compatibility upstream to `uniffi-bindgen-cs`? And if not, are there any other alternatives?

I'm a bit concerned about pinning ourselves to 0.28, uniffi does add very useful features still

examples/simple_counter/Windows/LICENSE.txt
kaidelorenzo: That should be removed. I think it was part of some template thing

examples/simple_counter/shared/Cargo.toml
kaidelorenzo: Yeah I think it makes sense to wait for the bring up to the latest uniffi. Definitely something I would like to do but isn't a priority for me right now

mknet: @charypar @kaidelorenzo and myself gave it a try the day before yesterday. Unfortunately, I does not work and migrating `uniffi-bindgen-cs` is waymore than just fixing some lines.

PR #354, opened 2025-04-29
crux_macros: Implement TryFrom for effect

DeltaEvo:
This allow to convert the effect enum into the underlying request in a generic way. This is mainly usefull for testing and allow to write functions that convert into an operation type without knowing the concrete Effect type.

```rust
fn expect_http_effect(effect: impl TryInto<crux_core::Request<HttpRequest>>) {
  ...
}
```


charypar:
Oh that's super nice, should've thought of that! 😅

charypar (approved):
Great! Thank you 🙏🏻

PR #355, opened 2025-04-29
crux_core/testing: Add track_caller to resolve_to_event_then_update

DeltaEvo:
This allow to show the caller location in in panic messages. In practice it means that panic messages will point the the user test code instead of crux testing code.


charypar (approved):
Oh, good spot! Thanks a lot!

PR #358, opened 2025-05-14
Counter example clean up

Bilaltariq98:
- Simplified the `update` function signature in `app.rs` for better readability.
- Added a check in tests to ensure the app state reflects pending server data.
- Removed unnecessary fields from the `ServerSentEvents` struct in `sse.rs` to streamline the implementation.


examples/counter/shared/src/capabilities/sse.rs
```diff
@@ -32,24 +28,15 @@ impl Operation for SseRequest {
...
-}
-
-impl<Event> ServerSentEvents<Event>
-where
-    Event: Send + 'static,
-{
-    pub fn new(context: CapabilityContext<SseRequest, Event>) -> Self {
-        Self { context }
-    }
+pub struct ServerSentEvents {}
```
charypar, line 31:
```suggestion
pub struct ServerSentEvents;
```

charypar (approved):

PR #363, opened 2025-05-20
add .query() method to protocol::HttpRequestBuilder

Bilaltariq98:
Introduces query on protocol::HttpRequestBuilder

The `crux_http` library has a query method to create query strings on `command::RequestBuilder`, but when writing tests for a crux core application

we would do something like 

```rust
let mut request = cmd.effects().next().unwrap().expect_http();

let expected_url = format!(
    "{}?lat={}&lon={}&appid={}",
    WEATHER_URL, lat_lng.0, lat_lng.1, API_KEY
);


// Before the changes in this PR
assert_eq!(&request.operation, &HttpRequest::get(expected_url).build());

// After the changes in the PR expose a query method
assert_eq!(
    &request.operation, // If we had query strings, the URL would already have it appended
    &HttpRequest::get(API_URL)
        // .query doesn't exist here as this is Protocol::HttpRequestBuilder. So we would have to format a string as an alternative for the meantime
        .query(&CurrentQueryString {
            lat: lat_lng.0.to_string(),
            lon: lat_lng.1.to_string(),
            appid: API_KEY,
        })
        .expect("could not serialize query string")
        .build()
);
```

Using `serde_qs` instead of `serde_urlencoded` as:
 - http_types uses it under the hood for .query on command::RequestBuilder
 - support abitrarily nested structs encoded in arbitrary orders

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -538,6 +538,7 @@ dependencies = [
  "serde",
  "serde_bytes",
  "serde_json",
+ "serde_qs 0.10.1",
  "thiserror 2.0.12",
  "url",
  "web-sys",
@@ -1162,7 +1163,7 @@ dependencies = [
  "rand 0.7.3",
  "serde",
  "serde_json",
- "serde_qs",
+ "serde_qs 0.8.5",
  "serde_urlencoded",
  "url",
 ]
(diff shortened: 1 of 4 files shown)


StuartHarris (changes requested):
Looks great @Bilaltariq98 Thank you. It feels like there should be one more test though. To verify that an existing URL that already has a query string gets the new stuff appended on the end correctly.

StuartHarris (approved):
Awesome!

PR #364, opened 2025-05-21
Make Typegen errors more informative

anvlkv:
Hi, if you don't mind a change like this I would be happy to extend in onto other Typegen errors.


StuartHarris:
Thanks @anvlkv looks great! By all means we can extend to other errors. Bear in mind though that this typegen will be deprecated at some point so might not be worth too much effort! :-)

PR #368, opened 2025-05-26
WIP: Introduce core-side middleware processing effects

charypar:
This is still very much work in progress, and is inspired by ideas demonstrated by @o0Ignition0o and @DeltaEvo. 

The goal is to allow the Core to be wrapped with a "middleware" that can modify events and view model, and process some of the effects. There are two key problems to solve:

1. Passing follow-up effects received when the midleware resolves an effect it can handle and encounters effects it cannot handle. This second resolution may well happen on a background thread handling the effect resolution, and is definitely no longer in the context of being called by the Shell. 
2. "Narrowing" the effect type received by the Shell - removing the variants which are processed by the middleware.

The basic idea is for the middlewares to be provided a callback to use when unknown effects are encountered. This can bubble up as high as the Shell, using [UniFFIs callback interfaces](https://mozilla.github.io/uniffi-rs/latest/types/callback_interfaces.html). For the narrowing, it should be possible to simply define an additional Effect type and implement a `TryFrom`. This conversion step can potentially be a middleware in its own right.

## Implementation steps

- [x] Demonstrate the basic mechanics with a single middleware
- [ ] Tease apart the common middleware machinery and the demo implementation - introduce a `Middlware` trait
- [ ] Allow middlewares to be stacked (by implementing `Middleware` on Core).
- [ ] Add Effect type narrowing using `TryFrom`
- [ ] Show the entire system working over FFI to Swift and WASM

### DIFF
--- a/crux_core/src/lib.rs
+++ b/crux_core/src/lib.rs
@@ -180,7 +180,7 @@ pub trait App: Default {
     /// `Event`, typically an `enum`, defines the actions that can be taken to update the application state.
     type Event: Unpin + Send + 'static;
     /// `Model`, typically a `struct` defines the internal state of the application
-    type Model: Default;
+    type Model: Default + Send + Sync;
     /// `ViewModel`, typically a `struct` describes the user interface that should be
     /// displayed to the user
     type ViewModel: Serialize;
@@ -190,7 +190,7 @@ pub trait App: Default {
     ///
     /// Note: this `Capabilities` associated type will be deprecated soon as part of the completion of
     /// the migration to the new [`Command`](command) API.
-    type Capabilities;
+    type Capabilities: Sync;
     /// `Effect`, the enum carrying effect requests created by capabilities.
     /// Normally this type is derived from `Capabilities` using the `crux_macros::Effect` derive macro
     type Effect: Effect + Unpin;
--- a/crux_core/tests/middlware.rs
+++ b/crux_core/tests/middlware.rs
@@ -1,0 +1,348 @@
+mod app {
+    use crux_core::{capability::Operation, macros::effect, render::render, App, Command};
+    use crux_http::command::Http;
+    use serde::{Deserialize, Serialize};
+
+    // Inline minimal random number capability
+
+    #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+    pub struct RandomNumberRequest(pub usize); // request a random number from 1 to N, inclusive
+
+    #[derive(Debug, PartialEq, Eq, Deserialize)]
+    pub struct RandomNumber(pub usize);
+
+    impl Operation for RandomNumberRequest {
+        type Output = RandomNumber;
+    }
+
+    // Random will be handled in a middleware
+    // Other effects will be passed to the shell
+
+    #[effect]
+    pub enum Effect {
+        Random(RandomNumberRequest),
+        Http(crux_http::protocol::HttpRequest),
+        Render(crux_core::render::RenderOperation),
+    }
+
+    #[derive(Serialize, PartialEq, Eq, Debug)]
+    pub enum Event {
+        Roll(Vec<usize>), // roll N dice with F_i faces
+
+        #[serde(skip)]
+        Random(RandomNumber),
+        #[serde(skip)]
+        Http(crux_http::Result<crux_http::Response<()>>),
+    }
+
+    #[derive(Default, Debug)]
+    pub struct Model {
+        roll: Roll,
+    }
+
+    #[derive(Default, Debug)]
+    pub enum Roll {
+        #[default]
+        NotStarted,
+        InProgress(Vec<(usize, Option<usize>)>), // Vec<(dice, value?)>
+        Complete(Vec<usize>),                    // Vec<value>
+    }
+
+    #[derive(Default)]
+    pub struct Dice;
+
+    impl App for Dice {
+        type Event = Event;
+        type Model = Model;
+        type ViewModel = ();
+        type Capabilities = ();
+        type Effect = Effect;
+
+        fn update(
+            &self,
+            event: Self::Event,
+            model: &mut Self::Model,
+            _caps: &(),
+        ) -> crux_core::Command<Self::Effect, Self::Event> {
+            match event {
+                Event::Roll(dice) => {
+                    if dice.is_empty() {
+                        return Command::done();
+                    }
+
+                    let first_dice = dice[0];
+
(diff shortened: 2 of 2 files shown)


o0Ignition0o:
Looking great so far, i have high hopes that + the callback udl shenanigans i did in #333 will work!

o0Ignition0o:
Right, I just checked things out, and it works! 🚀

I'm still not 100% sure about the dx of the "narrower" but it's certainly something that can be iterated on over time, looks great!

crux_core/src/core/resolve.rs
```diff
@@ -50,4 +52,6 @@ pub enum ResolveError {
     Never,
     #[error("Attempted to resolve a request that has concluded.")]
     FinishedMany,
+    #[error("Request with {0:?} not found.")]
+    NotFound(EffectId), // FIXME: since it uses EffectId, should ResolveError be defined further up the module tree?
```
StuartHarris, line 56: should this be `NotFound(u32)` instead then?

crux_core/src/middleware/bridge.rs
```diff
@@ -0,0 +1,179 @@
...
+        };
+
+        result.map(|()| requests_bytes)
+    }
+
+    /// Resolve a requested effect, providing the output to the core
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`BridgeError`] when the effect fails to resolve, or any of the (de)serialization fails.
```
StuartHarris:
```suggestion
    /// Returns a [`BridgeError`] when the effect fails to resolve, or any of the (de)serialization fails.
```

crux_core/src/bridge/mod.rs
```diff
@@ -7,7 +7,8 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use crate::{core::ResolveError, App, Core};
-use registry::{EffectId, ResolveRegistry};
+pub use registry::EffectId;
```
charypar, line 10: Type was already semi-public as part of `EffectFfi` I think, this just makes it explicit.

crux_core/src/middleware/effect_conversion.rs
```diff
@@ -0,0 +1,89 @@
+use std::marker::PhantomData;
+
+use crate::{capability::Operation, Request, ResolveError};
+
+use super::Layer;
+
+/// Middleware for converting the effect type to another type.
+///
+/// Typically, this is used to eliminate some of the effect variants which are processed
+/// by the layers below, so that code using this stack is not forced to have exteaneous
```
StuartHarris:
```suggestion
/// by the layers below, so that code using this stack is not forced to have extraneous
```

crux_core/src/middleware/effect_handling.rs
```diff
@@ -0,0 +1,269 @@
...
+pub struct HandleEffectLayer<Next, EM>
+where
+    Next: Layer + Sync + Send + 'static,
+    Next::Effect: TryInto<Request<EM::Op>, Error = Next::Effect>,
+    EM: EffectMiddleware<Next::Effect> + Send + Sync,
+{
+    inner: Arc<EffectMiddlewareLayerInner<Next, EM>>,
+}
+
+impl<Next, M> Layer for HandleEffectLayer<Next, M>
```
StuartHarris: `EM` as above

crux_core/src/middleware/effect_handling.rs
```diff
@@ -0,0 +1,269 @@
...
+pub trait EffectMiddleware<Effect>
+where
+    Effect: TryInto<Request<Self::Op>, Error = Effect>,
+{
+    /// The operation type this middleware can process
+    type Op: Operation;
+
+    /// Try to process `effect` if is of the right type (can convert in to a `Request<Self::Op>`).
+    ///
+    /// The implementation should return `Ok(())` if the conversion succeds, and call the `resolve_callback`
```
StuartHarris:
```suggestion
    /// The implementation should return `Ok(())` if the conversion succeeds, and call the `resolve_callback`
```

examples/counter-next/web-leptos/src/core.rs
```diff
@@ -64,5 +62,19 @@ pub fn process_effect(core: &Core, effect: Effect, render: WriteSignal<ViewModel
...
+        Effect::Random(mut request) => {
+            let RandomNumberRequest(min, max) = request.operation;
+            #[allow(clippy::cast_precision_loss)]
+            let number = js_sys::Math::random() * (max as f64 - min as f64) + min as f64;
+            #[allow(clippy::cast_possible_truncation)]
+            let number = number.floor() as isize;
+
+            for effect in core
+                .resolve(&mut request, RandomNumber(number))
+                .expect("shoudl resolve")
```
StuartHarris:
```suggestion
                .expect("should resolve")
```

examples/counter-next/iOS/CounterApp/core.swift
```diff
@@ -2,16 +2,40 @@ import Foundation
...
+    private var handler: EffectHandler
+    private var core: CoreFfi
+
     init() {
-        self.view = try! .bincodeDeserialize(input: [UInt8](Shared.view()))
+        self.handler = EffectHandler()
+        self.core = CoreFfi(handler)
+        self.view = try! .bincodeDeserialize(input: [UInt8](core.view()))
+
+        // the handler assignment needs to be defered, otherwise we create a circular
```
StuartHarris:
```suggestion
        // the handler assignment needs to be deferred, otherwise we create a circular
```

crux_core/src/middleware/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+pub use bridge::{Bridge, FfiFormat};
+pub use effect_conversion::MapEffectLayer;
+pub use effect_handling::{EffectMiddleware, HandleEffectLayer};
+pub use formats::BincodeFfiFormat;
+use serde::Deserialize;
+
+/// A layer in the middleware stack.
+///
+/// This is implemented by the Core and the different types of middlewares,
+/// so that they can be composed as requied.
```
StuartHarris:
```suggestion
/// so that they can be composed as required.
```

crux_core/src/core/resolve.rs
charypar: No, I don't think so, `EffectId` is now fully public anyway - it's more that this new error case means this error type is no longer limited to the `Core` resolving effects, it covers resolution errors across `Core` and `Bridge` now, so I was thinking it might need to be defined a level above.

It's a minor tidying question 🤷🏻‍♂️

StuartHarris (approved):

PR #369, opened 2025-05-27
crux_core: Add a way to take out the operation from a request

DeltaEvo:
This is useful for Operation types that don't implement Default, as for the one that do
```rust
let operation = std::mem::take(request.operation);
```
already fulfill the use.

For example today with the crux_kv Operation type
```rust
let mut operation = KeyValueOperation::Get {
    key: "".to_string(),
};
std::mem::swap(&mut request.operation, &mut operation);
```
is needed to take out the operation, with this commit it can be written with
```rust
let (operation, request) = request.take_operation();
```

### DIFF
--- a/crux_core/src/core/request.rs
+++ b/crux_core/src/core/request.rs
@@ -1,9 +1,22 @@
-use std::fmt::{self, Debug};
+use std::{
+    fmt::{self, Debug},
+    marker::PhantomData,
+};
 
 use crate::{
     capability::Operation,
     core::resolve::{Resolve, ResolveError},
 };
+
+/// A type that act like an operation `T`.
+///
+/// See [`Request::take_operation`] for details.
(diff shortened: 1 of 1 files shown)


DeltaEvo:
> Any good alternative name suggestions for Resolve - "the thing that identifies an effect request to resolve (and also is the mechanism to pass the output to the task waiting on it)"?

Wdyt, about just naming it `RequestHandle` ? To make it more future proof is something is added to the request

charypar:
Added https://github.com/redbadger/crux/issues/374 as a task to implement those changes.

charypar:
The alternative path implementation is now completed in #376.

StuartHarris:
Thanks @DeltaEvo for this. I implemented your suggestion (which we prefer) in #376 so closing this now. Feel free to reopen it if you have further queries/comments.

PR #372, opened 2025-05-27
fix(crux_cli): Handle built-in crates and add f32/f64 primitive support

fraschm1998:
- Skip built-in Rust crates (std, core, alloc, proc_macro, test) that don't have manifests
- Add support for f32 and f64 primitive types in codegen formatter
- Fixes 'unknown crate alloc/std' and 'need to implement primitive f32/f64' errors

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,13 +326,53 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-platform"
+version = "0.2.0"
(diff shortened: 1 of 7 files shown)


StuartHarris (approved):
Thanks @fraschm1998. I pushed a couple of small tweaks to get it past Clippy in CI. But happy to merge this. Will look at #373  now.

Issue #374, opened 2025-05-28
Add API to split `Request` into operation and `RequestHandle`


charypar:
Following conversation on #369, change the `resolve` API on `Core` to accept a `RequestHandle`, which is a new name for what's currently called `Resolve`. 

For backwards compatibility, we should update the API to be generic over an impl of `Resolvable` trait implemented on both `Request<Op>` and `RequestHandle<Out>`.

To do this, we need to

* Rename `Resolve` to `RequestHandle` and make the type public
* Introduce a `Resolvable` trait and implement it on `Request` and `RequestHandle`
* Change `Core::resolve` to accept an instance of the trait
* Add a `Request::split(self) -> (Op, RequestHandle<Op::Out>)` API to `Request`

PR #376, opened 2025-05-28
Request handle

StuartHarris:
fixes: #374

Following conversation on #369, change the `resolve` API on `Core` to accept a `RequestHandle`, which is a new name for what's currently called `Resolve`. 

For backwards compatibility, we update the API to be generic over an impl of `Resolvable` trait implemented on both `Request<Op>` and `RequestHandle<Out>`.

- [x] Rename `Resolve` to `RequestHandle` and make the type public
- [x] Introduce a `Resolvable` trait and implement it on `Request` and `RequestHandle`
- [x] Change `Core::resolve` to accept an implementation of the trait
- [x] Add a `Request::split(self) -> (Op, RequestHandle<Op::Out>)` API to `Request`

### DIFF
--- a/crux_core/src/core/request.rs
+++ b/crux_core/src/core/request.rs
@@ -2,10 +2,8 @@ use std::fmt::{self, Debug};
 
 use crate::{
     capability::Operation,
-    core::resolve::{RequestHandle, ResolveError},
+    core::resolve::{RequestHandle, Resolvable, ResolveError},
 };
-
-use super::resolve::Resolvable;
 
 /// Request represents an effect request from the core to the shell.
 ///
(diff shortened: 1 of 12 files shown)


crux_core/src/core/request.rs
```diff
@@ -17,17 +17,28 @@ where
...
+    pub handle: RequestHandle<Op::Output>,
 }
 
 impl<Op> Request<Op>
 where
     Op: Operation,
 {
+    /// Resolve the request with the given output.
+    /// # Errors
+    /// Returns an error if the request cannot be resolved.
```
charypar, line 29:
```suggestion
    /// Returns an error if the request does not expect to be resolved.
```

charypar (approved):

PR #378, opened 2025-05-29
fix(crux_cli): Add support for Mutex<T> and Instant types in formatter

fraschm1998:
- Mutex<T> is handled like other smart pointers (Box, Arc, Rc) - extracts inner type
- Instant is serialized as u64 (milliseconds since epoch)

This fixes TypeScript compilation errors when these types are used in the API.

### DIFF
--- a/crux_cli/src/codegen/formatter.rs
+++ b/crux_cli/src/codegen/formatter.rs
@@ -407,6 +407,20 @@ impl From<&Type> for Format {
                                     )
                                 }
                             }
+                            "Mutex" => {
+                                // Mutex<T> is just a thread-safe wrapper, extract the inner type
+                                match args.first() {
+                                    Some(GenericArg::Type(ref type_)) => type_.into(),
+                                    Some(other) => {
+                                        panic!("Mutex<T> expects a type parameter, got: {other:?}")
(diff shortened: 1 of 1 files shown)


charypar:
Hmm, I'd like to understand the expected behaviour of this across languages and the actual use scenario, because I'm not sure what should even be emitted in the foreign language for `Arc`, `Rc`, `Mutex` or the likes of `RefCell`.  

In my mind, it makes little sense to attempt to replicate the type strucutre in languages where these are simply not a thing (like TypeScript). It just about makes sense to skip `Box` (make it transparent), but containers with special behaviours are a step too far, IMO. What should probably happen instead is converting the structures containing Mutexes et al. into something straightforwardly `Serializable` first, then sending this over the FFI and deserializing into primitive types. The typegen is only intended to support plain data after all.

I'm curious as to what the scenario in which this comes up actually is in a Crux app - do you have `Mutex`es and `Arc`s in your Event or ViewModel?

fraschm1998:
> do you have `Mutex`es and `Arc`s in your Event or ViewModel?

I did from before, however unsure whether or not I'll need it.. Was used for livekit with the old Middleware integration

PR #379, opened 2025-05-30
Merge codegen fixes

fraschm1998:
Fixes `Error: unknown crate std` and `Error: unknown crate alloc` errors
Adds type alias support from https://github.com/redbadger/crux/pull/375
Adds external crate support

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -56,6 +56,8 @@ jobs:
 
       - name: Build Documentation (stable)
         working-directory: stable
+        env:
+          RUSTFLAGS: -A warnings
         run: cargo doc --all --no-deps
 
       - name: Build book (latest)
--- a/xtask/src/check.rs
+++ b/xtask/src/check.rs
@@ -21,7 +21,7 @@ impl Check {
                 .args(package_args)
                 .run()?;
             if self.clippy {
-                cmd!(ctx.sh, "{CARGO} clippy")
+                cmd!(ctx.sh, "{CARGO} clippy --all-targets")
                     .args(package_args)
                     .args(vec!["--", "--no-deps", "-Dclippy::pedantic", "-Dwarnings"])
                     .run()?;


fraschm1998:
Also adds type alias support from https://github.com/redbadger/crux/pull/375

fraschm1998:
@StuartHarris should be ready.. Do you want me to update all the examples to use new typegen?

crux_cli/src/codegen/formatter.rs
```diff
@@ -407,6 +447,30 @@ impl From<&Type> for Format {
...
+                                Format::F64
+                            }
+                            "Mutex" | "RefCell" => {
+                                panic!(
+                                    "{name} should not be used in types that cross FFI boundaries. \
+                                    These are runtime constructs that cannot be serialized. \
+                                    Consider refactoring your data model to use plain data types instead."
+                                )
+                            }
+                            "Instant" => {
```
StuartHarris: `std::time::Instant` is not generic, so will we ever match here?

crux_cli/src/codegen/mod.rs
```diff
@@ -8,25 +8,116 @@ mod serde;
...
+    if previous.contains_key(crate_name) {
+        return true;
+    }
+
+    // Built-in Rust crates
+    if STD_CRATES.contains(&crate_name) {
+        return true;
+    }
+
+    // Workspace members are handled separately
```
StuartHarris: I don't think I understand why workspace crates are handled separately. Could you explain to me a bit why they are not treated like any other "foreign" crate?

crux_cli/src/codegen/filter.rs
```diff
@@ -2,11 +2,25 @@
...
-use rustdoc_types::Crate;
+use rustdoc_types::{Crate, Generics, Id, Item, ItemEnum, Struct, StructKind, Visibility};
+use std::collections::HashMap;
 
-use super::item::{is_enum, is_relevant, is_struct, is_struct_unit};
+use super::item::{is_enum, is_relevant, is_struct, is_struct_unit, is_type_alias};
 use super::node::{CrateNode, ItemNode, SummaryNode};
 
+/// Standard library crate names that should be skipped during type generation
+pub const STD_CRATES: &[&str] = &["std", "core", "alloc", "proc_macro", "test"];
```
StuartHarris: I know we already merged something like this. But I'm not quite sure why we need to exclude standard library crates. They weren't causing me a problem before. There is a test somewhere which points to the location of the json for the core libs, but it's something like `share/doc/rust/json/` directory of the rustup toolchain.
I can imagine that there are serializable types in the core libs that we might want across the bridge.

crux_cli/src/codegen/formatter.rs
```diff
@@ -285,6 +318,17 @@ fn make_enum(formats: &[(&Indexed<Named<VariantFormat>>,)]) -> ContainerFormat {
     ContainerFormat::Enum(map)
 }
 
+fn make_type_alias(alias_node: &ItemNode) -> Option<ContainerFormat> {
+    // For type aliases like `type TimerId = String`, we generate a wrapper struct
```
StuartHarris: I didn't have any type aliases in the examples I was testing against, but when I introduced one in `simple_counter` as `type Count = String` it just seemed to work. It looked like it was transparent (as though I'd used a String directly). I'm guessing you must have seen different behaviour?

crux_cli/src/codegen/formatter.rs
```diff
@@ -8,16 +8,32 @@ use rustdoc_types::{GenericArg, GenericArgs, Item, ItemEnum, Type};
...
-        is_tuple_variant,
+        is_tuple_variant, is_type_alias,
     },
     node::ItemNode,
     serde::case::RenameRule,
     serde_generate::format::{ContainerFormat, Format, Named, VariantFormat},
 };
 
+/// Check if a type name is a framework internal type that shouldn't be generated
+fn is_framework_internal(name: &str) -> bool {
```
StuartHarris: I wasn't sure why this was needed. It's gonna be problematic if this gets out of date. Originally, we only explored the types we were interested in, and I _think_ that we should emit everything we encounter. I think I want to keep this "rules based" as in we find specific roots, and follow those through to the end, emitting all the types. Do you see a problem with this approach?

StuartHarris (commented):
Hey @fraschm1998 thank you ❤️  for your work on this so far. I'd love to incorporate some (if not all) of these changes, but had a few discussion points that we should align on first.

crux_cli/src/codegen/filter.rs
fraschm1998: They're being generated because we first generate all external types referenced in the type graph, then filter out what shouldn't become custom types. Without the STD_CRATES filter, std::String gets included in the initial collection and then treated as a custom type that conflicts with kotlin types.

1. Type Discovery Phase: external_type_needed relation (filter.rs:78-82) identifies ALL external types referenced in dependencies
2. Synthetic Type Creation: add_workspace_external_types() (filter.rs:243+) creates synthetic container types for these external references
3. Code Generation: These synthetic types get passed to serde-generate which creates custom classes

Without filtering, std::String flows through this pipeline and becomes com.example.project.shared.String instead of mapping to kotlin.String. *Without this I was getting hundreds of errors.*

I can remove the constant and instead have a maps_to_primitive_format function like: 

```rust
fn maps_to_primitive_format(type_name: &str) -> bool {
    matches!(
        type_name,
        "String" | "Vec" | "HashMap" | "BTreeMap" | "Option" | "Box" | "Arc" | "Rc"
    )
}
```

But then it would miss out on filtering:
image

However, you're right that there could be legitimate serializable types in core libs. The trade-off is:
  - Broad filtering (STD_CRATES): Safer, prevents primitive conflicts, but might miss legitimate std types
  - Targeted filtering (primitive list): More flexible, but requires maintaining the list of problematic types

crux_cli/src/codegen/mod.rs
fraschm1998: Workspace crates get special treatment because they represent your application's API surface vs external  dependencies so they have different filtering parameters.

  Workspace crates
  // Line 190 in mod.rs - ALL public types become roots
```rust
  filter.add_all_public_types_as_roots(&workspace_crate);
```
  - Generate ALL public types - comprehensive API exposure
  - Processed first in the pipeline
  - You control these types and designed them for FFI

  External crates (dependencies):
  // Only types actually referenced get included
```rust
  relation external_type_needed(SummaryNode);
  external_type_needed(t) <--
      remote_type_of(_, t),  // Only if referenced
      summary(t);
```
  - Generate only referenced types - minimal inclusion
  - Processed after workspace crates
  - These types weren't necessarily designed for FFI

  Practical Example:
  If your workspace has:
  - shared crate with 50 public types
  - serde dependency with 200 types

  Result:
  - All 50 shared types are available in generated Java/Swift/TypeScript
  - Only the 3 serde types you actually use get generated
  - Prevents generating 197 unused serde types

crux_cli/src/codegen/formatter.rs
fraschm1998: You're right, this was left over from before. Just removed it.

crux_cli/src/codegen/filter.rs
fraschm1998: I haven't yet encountered any issues with the constant filtering.. lmk if you want me to switch it to the maps_to_primitive_format

crux_cli/src/codegen/formatter.rs
fraschm1998: Yes removing it deleted TimerId and KeyValueEffect:

rg "KeyValueEffect"
```
src/effects/storage_effects.rs
4:pub type KeyValueEffect = KeyValueOperation;

src/effects/mod.rs
5:use storage_effects::KeyValueEffect;
19:    Storage(KeyValueEffect),
```

rg "TimerId"
```
src/capabilities/timer.rs
10:pub type TimerId = String;
18:        id: TimerId,
24:    ClearTimeout(TimerId),
29:        id: TimerId,
35:    ClearInterval(TimerId),
```

Which resulted in:

```
  location: class SetInterval
/Users/massimo/Workspace/projects/saas/shared/generated/java/me/TimerOperation.java:148: error: cannot find symbol
        public SetInterval(TimerId id, @com.novi.serde.Unsigned Integer interval_ms) {
                           ^
  symbol:   class TimerId
  location: class SetInterval
/Users/massimo/Workspace/projects/saas/shared/generated/java/me/TimerOperation.java:190: error: cannot find symbol
            public TimerId id;
                   ^
  symbol:   variable TimerId
  location: class SetInterval
/Users/massimo/Workspace/projects/saas/shared/generated/java/me/TimerOperation.java:220: error: cannot find symbol
            builder.value = TimerId.deserialize(deserializer);
                            ^
  symbol:   variable TimerId
  location: class ClearInterval
```

fraschm1998: And similar errors for KeyValueEffect

crux_cli/src/codegen/mod.rs
StuartHarris: Thanks @fraschm1998 this makes sense to me now. I was originally going with the concept of only following types referenced from types that implement the `App` trait, and its associated `Event`s and `Effect`s (and `ViewModel`). This has the upside that we don''t generate foreign types that we don't need (but are maybe accidentally public), but also has the downside that we don't generate foreign types that we _may_ need (say for additional, non-crux, FFI support). @charypar not sure what you think here? — should we generate foreign types for our whole public API?

PR #380, opened 2025-05-31
feat(crux_core): add `then_notify` and `build` methods to `*Builder`

StuartHarris:
Adds `then_notify()` to `RequestBuilder` so that you can follow up with a `notify_shell` after processing the response.
Also adds `build()` methods to all 3 builders that turns the builder into a `Command` (ignoring any responses that may come back). Typically you'd use `then_send` to get any responses back to the app via `Event`s, but sometimes you might not care, say, if it's a 3rd party capability or the last in a chain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1175,13 +1175,14 @@ dependencies = [
 
 [[package]]
 name = "guppy"
-version = "0.17.18"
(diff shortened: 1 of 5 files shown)


charypar (approved):
Nice.

PR #381, opened 2025-06-02
fix(simple_counter): update to use Rust 2024 edition, and other deps

StuartHarris:
Updates the [`simple_counter`](https://github.com/redbadger/crux/tree/master/examples/simple_counter) example to allow using Rust 2024 edition.

This includes using `resolver = 3`, which allows us to use `leptos` >= 0.8 (which needs resolver >= 2). However, the rules regarding supplying optional features to a colocated binary changed after resolver = 1. So the `uniffi-bindgen` binary in the `shared` package, now requires the feature `cli`, which in turn brings in the `uniffi/cli` feature. This resulted in changes being required for both iOS and Android project definitions.

Also updates other dependencies for all the `web-*` examples.

# Update
Still using `resolver = 1` because the `build.rs` in `shared_types` fails with resolver >= 2, failing to converge on a single `crux_core` (conflict with typegen enabled vs typegen not enabled). This is fine for now. The other changes still hold. It just means that if anyone wants to use leptos >= 0.7.8 then they will have to use a more recent resolver and hence not use `shared_types/build.rs`


examples/simple_counter/iOS/SimpleCounter.xcodeproj/project.pbxproj
charypar, line 1: Not sure if it's super relevant to the changes in this PR, but should we stop commiting the xcode project files, if we're driving them with `project.yml` anyway?

charypar (approved):
I know it's still WIP, but looks simple enough!

Small question about what files we commit and don't

examples/simple_counter/iOS/SimpleCounter.xcodeproj/project.pbxproj
StuartHarris: we probably should. But I think it needs more thought, as in a build step of some description to generate the project files (both for `shared` and `iOS`

PR #384, opened 2025-06-03
Weather example 

Bilaltariq98:
Adding a Weather Example that uses HTTP, KV and a custom Location Effect alongside an opinionated approach on structuring Pages/Views within CRUX.


examples/weather/iOS/Weather.xcodeproj/xcshareddata/xcschemes/Weather.xcscheme
```diff
@@ -0,0 +1,85 @@
...
+            BlueprintIdentifier = "45CD8315A1AD1CFAAAD5F135"
+            BuildableName = "Weather.app"
+            BlueprintName = "Weather"
+            ReferencedContainer = "container:Weather.xcodeproj">
+         </BuildableReference>
+      </BuildableProductRunnable>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "OPENWEATHER_API_KEY"
+            value = "FOOBAR"
```
Bilaltariq98, line 56: TODO: Add a note in README.md

examples/weather/iOS/Weather/AddFavoriteView.swift
```diff
@@ -0,0 +1,99 @@
...
+                    }
+                    
+                    Spacer()
+                }
+            }
+            .navigationTitle("Add Favorite")
+            .toolbar {
+                ToolbarItem(placement: .navigationBarLeading) {
+                    Button("Cancel") {
+//                        core.update(.addFavorite(.cancel))
```
Bilaltariq98, line 88: Up for debate but could rely on 

core.update(.navigate(Workflow.favorites(FavoritesState.idle)))

or 

core.update(.addFavorite(.cancel))

but using core.update(.addFavorite(.cancel)) seems more idiomatic (from the pov of testing a flow within the core and binding the right namings)

Bilaltariq98: On the flip side all we do is navigate so may not need a cancel event ..

examples/weather/ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+```
++-------------------+      +-------------------+
+|   iOS App (UI)    | <--> |   Rust Core (Crux) |
++-------------------+      +-------------------+
+         |                          |
+         |        FFI (UniFFI)      |
+         +--------------------------+
+```
+
```
charypar: What's the middle arrow about? I Claude making things up again?

examples/weather/ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+- **iOS App (SwiftUI)**: Handles all user interface, user input, and platform integration (e.g., Core Data for persistence, CoreLocation for location services).
+- **Rust Core (shared/)**: Contains all business logic, state, and effect management using the Crux pattern.
+- **FFI Bridge (shared_types/)**: Generates type-safe bindings for Swift (and other platforms) using UniFFI and Crux typegen.
```
charypar: Not technically quite true - the UniFFI bindgen is generated from `shared`, it's only the foreign types built using `serde-generate` that are sourced from `shared_types` and driven by its `built.rs`.

examples/weather/ARCHITECTURE.md
charypar, line 1: Worth reading through this and making sure it's still all true - some poor soul might read it 😅

examples/weather/shared/src/config.rs
```diff
@@ -0,0 +1,15 @@
...
+        "test_api_key".to_string()
+    }
+    #[cfg(not(test))]
+    {
+        use std::env;
+
+        env::var("OPENWEATHER_API_KEY")
+            .expect("OPENWEATHER_API_KEY must be set in .env or environment")
+    }
+});
```
charypar, line 15: Alternative, if you wanted to avoid the LazyLock and static, would be to have a function that either returns a constant or pulls the env var. You could likely mark it `#[inline]` as well.

No need to change, just thought I'd say.

examples/weather/shared/src/app.rs
```diff
@@ -0,0 +1,187 @@
...
+    location::{
+        capability::{LocationOperation, LocationResponse},
+        model::geocoding_response::GeocodingResponse,
+    },
+    weather::{self, events::WeatherEvent, model::current_response::CurrentResponse},
+};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum Event {
+    Navigate(Box<Workflow>),
```
charypar, line 25: Interesting approach - it does technically enable someone to navigate to the Favourites in the `ConfirmDelete` state, which is slightly odd, but probably fine.

examples/weather/shared/src/app.rs
```diff
@@ -0,0 +1,187 @@
...
+    Favorites(FavoritesState),
+    AddFavorite,
+}
+
+#[derive(Default, Debug)]
+pub struct Model {
+    pub weather_data: CurrentResponse,
+    pub page: Workflow,
+    pub favorites: Vec<Favorite>,
+    pub show_add_modal: bool,
```
charypar: Looks unused now

examples/weather/shared/src/app.rs
```diff
@@ -0,0 +1,187 @@
...
+                    favorites,
+                }
+            }
+            Workflow::AddFavorite => WorkflowViewModel::AddFavorite {
+                search_results: model.search_results.clone(),
+            },
+        };
+
+        ViewModel { workflow }
+    }
```
charypar: Some of this could be refactored into `From` implementations for the view model types.

Another approach I've seen was a `View` trait which added a `view()` method to the models that implemented it.

examples/weather/shared/src/favorites/events.rs
```diff
@@ -0,0 +1,595 @@
...
+use crate::location::model::geocoding_response::{
+    GeocodingQueryString, GeocodingResponse, GEOCODING_URL,
+};
+
+use crate::{Effect, Workflow};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum FavoritesEvent {
+    // Workflow - Favorites view
+    DeletePressed(Box<Favorite>),
```
charypar: Is there a simpler way of identifying favourite than the entire value?

examples/weather/shared/src/favorites/events.rs
```diff
@@ -0,0 +1,595 @@
...
+    DeleteConfirmed,
+    DeleteCancelled,
```
charypar, line 21: I wonder whether the confirmation dialog should be considered a "UI only concern" ... very debatable of course.

examples/weather/shared/src/location/model/geocoding_response.rs
```diff
@@ -0,0 +1,167 @@
...
+    LazyLock::new(|| serde_json::to_string(&*SAMPLE_GEOCODING_RESPONSE).unwrap());
+
+#[derive(Debug, Serialize, Deserialize, PartialOrd, PartialEq, Default, Clone)]
+pub struct ZipCodeResponse {
+    pub zip: String,
+    pub name: String,
+    pub lat: f64,
+    pub lon: f64,
+    pub country: String,
+}
```
charypar, line 143: Might be worth putting those into JSON tests fixtures.  They look like they are only used in tests anyway.

examples/weather/shared/src/weather/events.rs
```diff
@@ -0,0 +1,465 @@
...
+    pub lat: String,
+    pub lon: String,
+    pub units: &'static str,
+    pub appid: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum WeatherEvent {
+    Show,
+    LocationEnabled(bool),
```
charypar, line 25: This one feels like it should also be tagged `#[serde(skip)]`

examples/weather/shared/src/weather/events.rs
```diff
@@ -0,0 +1,465 @@
...
+        WeatherEvent::Fetch(lat, long) => Http::get(WEATHER_URL)
+            .expect_json()
+            .query(&CurrentQueryString {
+                lat: lat.to_string(),
+                lon: long.to_string(),
+                units: "metric",
+                appid: API_KEY.clone(),
+            })
+            .expect("could not serialize query string")
+            .build()
```
charypar, line 83: A good future enhancement might be a weather API client encapsulating this request building

examples/weather/shared/src/weather/events.rs
```diff
@@ -0,0 +1,465 @@
...
+        }
+        WeatherEvent::FetchFavorites => {
+            // Create a sequence of commands to fetch weather for each favorite
+            let mut cmd = Command::done();
+
+            for favorite in &model.favorites {
+                cmd = cmd.and(Command::event(Event::Home(Box::new(
+                    WeatherEvent::FetchFavorite(favorite.geo.lat, favorite.geo.lon),
+                ))));
+            }
```
charypar: Nit - this could be written as a `.map` - `Command` supports collecting from iterator of commands as well.

So something like

```
model.favourites.map(|f| Command::event(...)).collect()
```

charypar: Another good reason to have a weather API client too - to avoid the indirection through the `FetchFavourite` event

examples/weather/shared/src/weather/model/current_response.rs
```diff
@@ -0,0 +1,144 @@
...
+        .timezone(1)
+        .id(1_usize)
+        .name("Phoenix".to_string())
+        .cod(200_usize)
+        .build()
+        .expect("Failed to build sample response")
+});
+
+pub static SAMPLE_CURRENT_RESPONSE_JSON: LazyLock<String> =
+    LazyLock::new(|| serde_json::to_string(&*SAMPLE_CURRENT_RESPONSE).unwrap());
```
charypar, line 144: Same comment as for the other fixtures.

examples/weather/iOS/Weather/AddFavoriteView.swift
charypar: Cancel feels like it captures the intent better.

charypar (approved):
Nice work! 

There are a few opportunities fro refactoring here and there, more so on the Swift side, but I think this can get merged as is and evolved from there.

I like your take on the domain oriented module layout!

examples/weather/shared/src/favorites/events.rs
Bilaltariq98: Yeah I guess it depends, if the entire modal is done shell side without crux having awareness then it would be a purely UI only concern, but if we're modelling all the behaviours and that's consistent across all platforms then it fits. 

A real hard one to say either or - I think its ok as is though

examples/weather/shared/src/favorites/events.rs
Bilaltariq98: just passing lat, lng at the simplest end

PR #385, opened 2025-06-04
feat(counter_next): remove legacy boilerplate from sse capability

StuartHarris:
Modifies the `sse` capability in `counter-next` example to remove support for deprecated capabilities.


charypar (approved):
:shipit:

PR #388, opened 2025-06-13
fix(crux_cli): remove need for env::set_var as unsafe in 2024 edition

StuartHarris:
When building `crux_cli` for typegen in `counter-next` we were passing the crate name as an environment variable to the cli. This is now unsafe in the 2024 edition of Rust, so instead we pass it as an argument.  Also removes the lib name from the examples' manifests as this was determining the name of the file containing the rustdoc json.

### DIFF
--- a/examples/counter-next/shared/Cargo.toml
+++ b/examples/counter-next/shared/Cargo.toml
@@ -1,12 +1,11 @@
 [package]
 name = "shared"
 version = "0.1.0"
-edition = "2021"
-rust-version = "1.80"
+edition = "2024"
+rust-version = "1.87"
 
 [lib]
 crate-type = ["lib", "staticlib", "cdylib"]
-name = "shared"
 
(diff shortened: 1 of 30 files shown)


examples/counter-next/shared/Cargo.toml
```diff
@@ -1,12 +1,11 @@
 [package]
 name = "shared"
 version = "0.1.0"
-edition = "2021"
-rust-version = "1.80"
+edition = "2024"
+rust-version = "1.87"
```
charypar, line 5: How widely does this apply? 😬

StuartHarris: just in a couple of examples. MSRV for `crux_cli` is 1.86 in order to get v39 of rustdoc json. Other than that it's 1.80 for `crux_core` etc. But that will likely change if we merge facet-typegen

PR #390, opened 2025-06-17
Facet typegen

StuartHarris:
This adds a new frontend to the typegen that is based on [`facet`](https://github.com/facet-rs/facet) instead of [`serde-reflection`](https://github.com/zefchain/serde-reflection).

The `counter_next` example has been updated in this PR to use this facet-based typegen (for which it again needs the `shared_types` member in the workspace). I tested iOS, Android, Web(rust) and Web(TS).

### Notes
1. relies on [`facet`](https://github.com/facet-rs/facet), which is evolving very rapidly.
2. relies on our own [`facet-generate`](https://github.com/redbadger/facet-generate) crate. 
3. also depends on v3.0.0 of [our temporary fork of http-types](https://crates.io/crates/http-types-red-badger-temporary-fork), which adds `#[derive(Facet)]` to all the types we get from that crate that we send over the FFI (@charypar is in the middle of removing this dependency though in favour of the [`http`](https://crates.io/crates/http) crate).
4. requires us to up our MSRV to 1.87


Cargo.toml
```diff
@@ -10,27 +11,27 @@ members = [
   "doctest_support",
   "xtask",
 ]
-resolver = "1"
 
 [workspace.package]
 authors = ["Red Badger Consulting Limited"]
-edition = "2021"
+edition = "2024"
+rust-version = "1.87"
```
charypar, line 18: Are we forcing everyone to go onto the very latest Rust version with this? 😓

examples/counter-next/shared/src/app.rs
```diff
@@ -4,35 +4,55 @@ use crux_core::{
...
+    for HttpResultxx<crux_http::Response<T>, HttpError>
+{
+    fn from(value: crux_http::Result<crux_http::Response<T>>) -> Self {
+        match value {
+            Ok(response) => HttpResultxx::Ok(response),
+            Err(error) => HttpResultxx::Err(error),
+        }
+    }
+}
+
```
charypar: 🙈

examples/counter-next/shared_types/build.rs
```diff
@@ -0,0 +1,21 @@
...
+    let output_root = PathBuf::from("./generated");
+
+    typegen.swift("SharedTypes", output_root.join("swift"))?;
+
+    typegen.java("com.crux.example.counter.shared", output_root.join("java"))?;
+
+    typegen.typescript("shared_types", output_root.join("typescript"))?;
+
+    Ok(())
+}
```
charypar, line 21: Back with a vengeance 😅

examples/counter-next/typegen.sh
```diff
@@ -9,11 +9,11 @@ rm -rf headers sources *.swift *.h *.modulemap
...
-cargo run --package shared --bin crux_cli --features cli -- \
-    codegen \
-        --java com.crux.example.counter.shared \
-        --swift SharedTypes \
-        --typescript shared_types
+# cargo run --package shared --bin crux_cli --features cli -- \
+#     codegen \
+#         --java com.crux.example.counter.shared \
+#         --swift SharedTypes \
+#         --typescript shared_types
```
charypar, line 16: This can come back even with Facet at some point, right?

examples/counter-next/web-remix/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+export RUSTFLAGS='--cfg getrandom_backend="wasm_js"'
+
+pushd ../shared && RUST_LOG=info wasm-pack build --target web
```
charypar, line 5: Instead of this, In my PR I've avoided `rand` in wasm and used `js_sys` for random numbers - get them directly from the browser. 

Might be a better way, and avoids needing various shell scripts and magic env vars...?

crux_core/src/type_generation/facet.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// # Errors
+    ///
+    /// Errors that can occur during type registration.
+    pub fn register_type<'a, 'de, T>(&mut self) -> Result
+    where
+        T: serde::Deserialize<'de> + Facet<'a>,
+    {
+        match &mut self.state {
+            State::Registering(registry) => {
+                let incoming = facet_generate::reflect::<T>();
```
charypar, line 247: Does `facet_generate::reflect` return a serde-generate compatible IR? 

That might be a good, low-risk way of keeping compatibility - have a `reflect_serde` next to `reflect`, where the former returns a serde-generate compatible IR, and `reflect` returns our one with additional info.

charypar (approved):
Good stuff. Got a bunch of questions, but generally happy with it. Nice work!!

crux_core/src/type_generation/facet.rs
charypar: Ignore this - we need to be able to do serde_generate compatible output _per type_ anyway, via our new IR and backend to smooth out the migration for people. No point _also_ maintaining a sideways step to serde IR.

crux_core/tests/middleware.rs
```diff
@@ -1,14 +1,15 @@
...
     use serde::{Deserialize, Serialize};
 
     // Inline minimal random number capability
 
-    #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+    #[derive(Facet, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
     pub struct RandomNumberRequest(pub usize); // request a random number from 1 to N, inclusive
 
-    #[derive(Debug, PartialEq, Eq, Deserialize)]
+    #[derive(Facet, Debug, PartialEq, Eq, Deserialize)]
```
charypar: This is _interesting_. In this integration test, no type generation should be happening, it's curious that Facet is required.

charypar (approved):

Cargo.toml
StuartHarris: Well, kinda. 1.88 is out now, so not the very latest 😅

examples/counter-next/typegen.sh
StuartHarris: Yes, indeed. In a separate PR we should update `crux_cli` to support facet typegen as well

crux_core/tests/middleware.rs
StuartHarris: Good spot, thanks. Removed now.

PR #391, opened 2025-06-17
turn uniffi optional

beltram:
Hi,
by default crux_core was pulling uniffi which is only required at build step, not at compile step... and is very slow to compile on cold builds (sometimes hot ones as well).
This allows the consumer of the lib to feature gate uniffi and turn it off when just compiling the core

### DIFF
--- a/crux_core/Cargo.toml
+++ b/crux_core/Cargo.toml
@@ -11,8 +11,10 @@ keywords.workspace = true
 rust-version.workspace = true
 
 [features]
+default = ["bindgen"]
 cli = ["dep:crux_cli"]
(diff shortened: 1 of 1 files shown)


crux_core/Cargo.toml
```diff
@@ -40,7 +42,7 @@ serde-reflection = { version = "=0.4.0", optional = true }
 wasm-bindgen = "0.2.100"
```
charypar, line 42: Should `wasm-bindgen` be optional too, I wonder 🤔

beltram: No I think it's fine

PR #395, opened 2025-06-23
Weather example improvements

Bilaltariq98:
- [x] Weather client API & clean up  
- [x] Geocoding (location) client API


examples/weather/shared/src/favorites/events.rs
```diff
@@ -70,15 +67,7 @@ pub fn update(event: FavoritesEvent, model: &mut crate::Model) -> Command<Effect
...
-        FavoritesEvent::Search(query) => Http::get(GEOCODING_URL)
-            .expect_json()
-            .query(&GeocodingQueryString {
-                q: query,
-                limit: "5",
-                appid: API_KEY.clone(),
-            })
-            .expect("could not serialize query string")
-            .build()
+        FavoritesEvent::Search(query) => LocationApi::fetch::<FavoritesEvent>(&query)
```
charypar, line 70: I would not go as far as the client making the decision about the event to send - I think returning a `RequestBuilder` is fine, but what you can also do in the `LocationApi` is ther result processing, so that the callback just gets the search results.

examples/weather/shared/src/location/client.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Fetch current weather for a specific location
+    pub fn fetch<Event>(
+        query: &str,
+    ) -> RequestBuilder<
+        Effect,
+        Event,
+        impl std::future::Future<
+            Output = Result<crux_http::Response<Vec<GeocodingResponse>>, crux_http::HttpError>,
+        >,
```
charypar: IMO this should be
```
impl Future<Output = Result<Vec<GeocodingResponse>, crux_http::HttpError>>
```

or, for extra brownie points, with a custom error enum (e.g. `LocationError` or `WeatherApiError`).

charypar (approved):

PR #401, opened 2025-06-28
Experiment: compile the `counter_next` example to a WebAssembly Component

StuartHarris:
This is an experiment to run a Crux application as a WebAssembly Component (outside the browser). This should help with dynamically loading a core into an MCP server so that it can be called as a tool by an LLM, e.g. Claude 4, during an agentic coding session, as though the LLM was a shell.

```bash
cd examples/counter-next
cargo build --package shared --target wasm32-wasip2 --features facet_typegen
```

Alternatively we could load a dylib with something like [`libloading`](https://crates.io/crates/libloading) but this seems like more fun

### DIFF
--- a/examples/counter-next/shared/src/ffi.rs
+++ b/examples/counter-next/shared/src/ffi.rs
@@ -167,87 +167,71 @@ pub mod wasm_ffi {
 
 #[cfg(all(target_os = "wasi", target_env = "p2"))]
 pub mod wasip2 {
-    use crux_core::{Core, bridge::BridgeWithSerializer, type_generation::facet::TypeRegistry};
-    use std::sync::LazyLock;
+    use crux_core::{
+        Core,
+        bridge::EffectId,
+        middleware::{Bridge, FfiFormat, Layer as _},
+        type_generation::facet::TypeRegistry,
+    };
+    use serde_json::{Deserializer, Serializer, de::SliceRead};
 
     use crate::App;
-
-    /// The main interface used by the shell
-    pub struct CoreFFI {
-        core: BridgeWithSerializer<App>,
-    }
-
-    impl CoreFFI {
-        pub fn new() -> Self {
-            let core = BridgeWithSerializer::new(Core::new());
-
-            Self { core }
-        }
-
-        #[must_use]
-        pub fn update(&self, data: &[u8]) -> Result<Vec<u8>, String> {
-            let mut deser = serde_json::Deserializer::from_slice(data);
-
-            let mut return_buffer = vec![];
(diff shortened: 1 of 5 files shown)


examples/counter-next/shared/wit/world.wit
```diff
@@ -0,0 +1,7 @@
+package counter-next:shared;
+
+world shared {
+    export update: func(data: list<u8>) -> list<u8>;
+    export resolve: func(effect-id: u32, data: list<u8>) -> list<u8>;
+    export view: func() -> list<u8>;
+}
```
charypar: Not saying to do this now, but to align with the other FFIs, we could treat the core instance as a resource owned by the component, which the host creates using a function like `new`. 

That would then allow use of middleware (I think 🤔 ) and also multiple instances.

StuartHarris:
It compiles so it must work lol ;). Was just about to load it into my MCP server but ran out of time. Trying to keep it for weekend fun times, but real fun took over :)

examples/counter-next/shared/wit/world.wit
StuartHarris: I started there, and then wound it back to something simpler for now. But deffo worth exploring once we prove this bit out.

examples/counter-next/shared/src/ffi.rs
```diff
@@ -164,3 +164,90 @@ pub mod wasm_ffi {
...
+    use crate::App;
+
+    /// The main interface used by the shell
+    pub struct CoreFFI {
+        core: BridgeWithSerializer<App>,
+    }
+
+    impl CoreFFI {
+        pub fn new() -> Self {
+            let core = BridgeWithSerializer::new(Core::new());
```
charypar: You _could_ use the new bridge with plug-in FFI format here. 

Not that you'll get much for that, it's just a bit less hassle to deal with the serializers.

examples/counter-next/shared/src/ffi.rs
```diff
@@ -164,3 +164,90 @@ pub mod wasm_ffi {
...
+            let mut return_buffer = vec![];
+            let mut ser = serde_json::Serializer::new(&mut return_buffer);
+
+            self.core.view(&mut ser).map_err(|e| e.to_string())?;
+
+            Ok(return_buffer)
+        }
+    }
+
+    static CORE: LazyLock<CoreFFI> = LazyLock::new(|| CoreFFI::new());
```
charypar: Why not follow the pattern above^ where the host creates the core instance? 👀

charypar (approved):

charypar (approved):

PR #402, opened 2025-06-29
Deprecate capabilities api

charypar:
This marks the key types in the capabilities API deprecated to help users migrate and warn them about future removal of this API.

- [x] add deprecations and migrate remaining tests et. al
- [ ] review docs

### DIFF
--- a/crux_core/src/capabilities/compose.rs
+++ b/crux_core/src/capabilities/compose.rs
@@ -37,7 +37,10 @@ use futures::Future;
 /// Note that testing composed effects is more difficult, because it is not possible to enter the effect
 /// transaction "in the middle" - only from the beginning - or to ignore some of the effects with out
 /// stalling the entire downstream dependency chain.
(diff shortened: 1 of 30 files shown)


StuartHarris (approved):
Amazing! Feels good! 
I wonder if we should also mark the `Capabilities` associated type on the `App` trait as deprecated. This would then warn if you create an `update` method with `_caps: &Self::Capabilities` instead of `_caps: &()`. WDYT?

StuartHarris:
Also, thanks for updating old tests 😓

charypar:
I looked at that, but it would mean everyone gets a warning, even though technically they are doing nothing wrong for now, and in one of the next versions the type and the argument will disappear anyhow.

The deprecation warnings already in make it such that it's impossible to define a type which can be used as `App::Capabilities` without a warning, so that should catch all the people who still need to migrate, without bothering the ones who are up to the latest.

PR #405, opened 2025-07-02
feat(crux_core): facet typegen with module support

StuartHarris:
Use `facet_generate` v0.4, which has namespace support for creation of specified modules in Swift, Kotlin and TypeScript. Updated `counter-next` example to put the types from the `ServerSentEvents` capability into their own namespace and tested for iOS, Android and NextJS.

### DIFF
--- a/examples/weather/Cargo.toml
+++ b/examples/weather/Cargo.toml
@@ -15,6 +15,9 @@ anyhow = "1.0.98"
 crux_core = { path = "../../crux_core" }
 crux_http = { path = "../../crux_http" }
 crux_kv = { path = "../../crux_kv" }
+# crux_core = "0.15.0"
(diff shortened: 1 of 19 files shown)


examples/weather/Cargo.toml
```diff
@@ -15,6 +15,9 @@ anyhow = "1.0.98"
 crux_core = { path = "../../crux_core" }
 crux_http = { path = "../../crux_http" }
 crux_kv = { path = "../../crux_kv" }
+# crux_core = "0.15.0"
+# crux_http = "0.14.0"
+# crux_kv = "0.9.0"
```
charypar, line 20: Did you mean to switch these to master?

charypar (approved):

PR #407, opened 2025-07-07
feat(crux_core): support registering multiple facet types

StuartHarris:
For facet-based typegen, support registering multiple types that are marked with the `#[derive(Facet)]` attribute.


crux_core/src/type_generation/facet.rs
```diff
@@ -201,14 +204,13 @@ impl TypeGen {
...
     {
-        match &mut self.state {
-            State::Registering(registry) => {
-                let incoming = reflect::<T>();
-                registry.extend(Registry::from(incoming));
-                Ok(())
-            }
-            State::Generating(_) => Err(TypeGenError::LateRegistration),
-        }
+        let state = mem::replace(&mut self.state, State::None);
```
charypar: `state.take()` would be nice?

crux_core/src/type_generation/facet.rs
```diff
@@ -247,7 +249,7 @@ impl TypeGen {
         };
 
         let root_module = package_name;
-        for (module, registry) in namespace::split(root_module, registry.clone()) {
+        for (module, registry) in module::split(root_module, registry.clone()) {
```
charypar: More of a thought - should `split` borrow registry and clone the individual types...?

charypar (approved):
nice!

crux_core/src/type_generation/facet.rs
StuartHarris: good shout. will do

crux_core/src/type_generation/facet.rs
StuartHarris: yeah. much better. here comes `facet-generate` v0.6.0 lol

PR #409, opened 2025-07-10
feat(crux_core): facet_typegen support for external dependencies

StuartHarris:
Minor changes to work with `facet_generate` v0.7.0, which supports emitting namespaces as external SPM packages.
The `counter_next` example emits 4 packages now:
1. `Shared` — the binary for the shared lib
2. `SharedTypes` — most of the generated foreign types
3. `ViewModel` — generated foreign types for the `ViewModel` struct
4. `Serde` — the generated serialization code as a separate package that `SharedTypes` and `ViewModel` both depend on.

Note: it's a bit hack at the moment, but will evolve as we find the rough edges

### DIFF
--- a/examples/counter-next/shared_types/build.rs
+++ b/examples/counter-next/shared_types/build.rs
@@ -1,21 +1,55 @@
-use crux_core::type_generation::facet::TypeGen;
-use shared::App;
+use crux_core::type_generation::facet::{ExternalPackage, PackageLocation, TypeGen};
+use shared::{App, ViewModel};
 use std::path::PathBuf;
 
 fn main() -> anyhow::Result<()> {
     println!("cargo:rerun-if-changed=../shared");
 
-    let mut typegen = TypeGen::new();
+    let mut typegen_app = TypeGen::new();
 
-    typegen.register_app::<App>()?;
+    typegen_app.register_app::<App>()?;
 
     let output_root = PathBuf::from("./generated");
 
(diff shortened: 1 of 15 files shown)


examples/counter-next/shared_types/build.rs
```diff
@@ -1,21 +1,55 @@
...
+            for_namespace: "Serde".to_string(),
+            location: PackageLocation::Path("../Serde".to_string()),
+            version: None,
+        }],
+        false,
+        false,
+    )?;
+
+    let mut typegen_serde = TypeGen::new();
+    typegen_serde.swift("Serde", output_root.join("swift"), vec![], true, false)?;
```
charypar, line 52: These look _fine_ honestly... In future versions, we might want some kind of configuration structure with a builder to make the config less "positional", but it looks alright.

xtask/src/publish.rs
```diff
@@ -11,7 +11,7 @@ const CARGO: &str = crate::CARGO;
 
 // in order for publishing
 const PACKAGES: &[&str] = &[
-    "crux_cli",
+    // "crux_cli",
```
charypar, line 14: Intentional?

charypar (approved):

xtask/src/publish.rs
StuartHarris: kinda. It's better out for now, while we think about how to effectively treat the cli and the "rest" as 2 separate things

PR #410, opened 2025-07-15
Make payloads in bridge echo variable to test impact of size

charypar:


examples/bridge_echo/cli/src/main.rs
```diff
@@ -21,6 +21,19 @@ async fn main() -> Result<()> {
     let (tx, rx) = unbounded::<Effect>();
     let arc_tx = Arc::new(tx);
 
+    let payload: Vec<_> = (1..10)
+        .map(|id| DataPoint {
+            value: 3.0 * f64::from(id),
+            id: 17,
```
StuartHarris, line 27: Totally inconsequential, but wondered if you meant this to be id rather than 17 each time? ;-)

examples/bridge_echo/web-leptos/src/main.rs
```diff
@@ -3,23 +3,43 @@ mod core;
...
 use leptos::prelude::*;
-use shared::Event;
+use shared::{DataPoint, Event};
 
 #[component]
 fn RootComponent() -> impl IntoView {
+    let payload: Vec<_> = (1..10)
+        .map(|id| DataPoint {
+            value: 3.0 * f64::from(id),
+            id: 17,
```
StuartHarris, line 13: and here

StuartHarris (approved):
nice

PR #411, opened 2025-07-16
feat(crux_core): Config, with builder, for type generation

StuartHarris:
Adds a `Config` and associated builder for foreign type generation (currently Swift only)

### DIFF
--- a/crux_core/src/type_generation/facet.rs
+++ b/crux_core/src/type_generation/facet.rs
@@ -70,9 +70,8 @@
 //!    typegen.typescript("shared_types", output_root.join("typescript"))?;
 //!}
 //! ```
-use derive_builder::Builder;
 use facet::Facet;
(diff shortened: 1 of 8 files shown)


crux_core/src/type_generation/facet.rs
```diff
@@ -138,6 +139,63 @@ impl Export for () {
...
+
+    /// # Panics
+    /// If any required fields are not initialized.
+    #[must_use]
+    pub fn build(&self) -> Config {
+        self.fallible_build()
+            .expect("All required fields were initialized")
+    }
+}
+
```
charypar: Should these be owned by `facet_typegen`? 🤔

StuartHarris: hmm, probably. The `add_extensions()` is in the way though. I'll separate that out and then move it.

PR #412, opened 2025-07-18
feat(crux_core): Add into_future to Command

DeltaEvo:
This aligns the Command api with {Request,Stream}Builder and allow to chain commands from inside the task function allowing to use capabilities that don't return values (for example an analytics capability) and use Commands as return values.


StuartHarris (approved):
Nice! Thanks David ❤️

PR #413, opened 2025-07-21
Relax Serialize bound when not using Bridge.

charypar:
The `Serialize`/`Deserialize` bounds are only really required when working with FFI and type generation, but are currently on the `App` trait and the `Effect` type for all uses.

This moves them up to the two bridge implementations and splits the effect derive macros to allow use without serialization, if no typegen is requested.


StuartHarris (approved):
Amazing. Will also try it out on `crux_mcp`

PR #415, opened 2025-07-22
TypeRegistry and CodeGenerator for typesafe transition between stages

StuartHarris:
Note: applies to facet-based typegen only.

Replace `TypeGen` (whose `ensure_registry()` would panic if misused) with `TypeRegistry` and `CodeGenerator` to ensure a type safe transition between the 2 stages. Also provides a nicer api:

```rust
TypeRegistry::new()
    .register_app::<App>()
    .register_type::<OtherType>()
    .build()
    .swift(config)?;
```

### DIFF
--- a/crux_core/src/type_generation/facet.rs
+++ b/crux_core/src/type_generation/facet.rs
@@ -107,21 +107,6 @@ pub enum TypeGenError {
     PnpmNotFound(#[source] std::io::Error),
 }
 
-#[derive(Debug, Default)]
-pub enum State {
-    #[default]
-    None,
-    Registering(RegistryBuilder),
-    Generating(Registry),
-}
-
-impl State {
-    #[must_use]
-    pub fn take(&mut self) -> Self {
-        std::mem::take(self)
-    }
-}
-
 pub trait Export {
     /// Register types with the type generator.
     ///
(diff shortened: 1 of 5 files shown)


crux_core/src/type_generation/facet.rs
```diff
@@ -107,56 +107,39 @@ pub enum TypeGenError {
...
+/// The `TypeGen` struct stores the registered types so that they can be generated for foreign languages
+/// use `TypeGen::new()` to create an instance
+impl TypeRegistry {
+    /// Creates an instance of the `TypeGen` struct for registration only
     #[must_use]
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            builder: RegistryBuilder::new(),
+        }
```
charypar: I reckon it might be better to have a `TypeRegistry::builder()` which gives you `RegistryBuilder`, on which you can then call `register_app` and `register_type` and once you call `build()`, you get `TypeRegistry` which is read-only.
  
That's than the input to the CodeGenerator for the various languages.

charypar: Ah never mind, I see what's going on. `TypeRegistry` is really just an extension on `RegistryBuilder` to get the extra functions 🤔

StuartHarris: Cool, thanks. Yeah, I made it a newtype to be clearer.

PR #420, opened 2025-07-27
feat(counter-next): update facet typegen for typescript

StuartHarris:
Updates the TypeScript type generation to use the new features in [`facet-generate`](https://github.com/redbadger/facet-generate) v0.9.0

### DIFF
--- a/crux_core/src/type_generation/facet.rs
+++ b/crux_core/src/type_generation/facet.rs
@@ -88,14 +88,15 @@ use facet_generate::{
     generation::{
         Encoding, SourceInstaller, java,
         module::{self, Module},
-        swift, typescript,
+        swift,
+        typescript::{self, InstallTarget},
     },
     reflection::RegistryBuilder,
 };
 use std::{
     fs::{self, File},
     io::Write,
-    path::{Path, PathBuf},
+    path::PathBuf,
 };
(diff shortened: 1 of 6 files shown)


crux_core/src/type_generation/facet.rs
```diff
@@ -377,25 +378,21 @@ impl CodeGenerator {
...
-
             let mut output = File::create(types_dir.join(format!("{module_name}.ts")))?;
-            write!(output, "{out}")?;
+            write!(output, "{}", String::from_utf8_lossy(&source))?;
         }
 
         if config.add_extensions {
             let extensions_dir = Self::extensions_path("typescript");
-            copy(extensions_dir, &config.out_dir)?;
+            for file in ["package.json", "tsconfig.json"] {
```
charypar: Are those really extensions? 🤔

charypar (commented):
Getting in early with the 🔍

crux_core/src/type_generation/facet.rs
StuartHarris: nope. `facet-generate` can now create `package.json` so it now does that. And we build `tsconfig.json` here, for now. 😓

charypar (approved):

PR #423, opened 2025-08-18
crux_core/command: Keep Command pending if a context is alive

DeltaEvo:
If a future using a CommandContext is stored (e.g. for caching)
but the associated Command terminates before, crux will panic when
using the context due to the inner channel being disconnected since
the Command (holding the Receiver) was dropped.

A nicer implementation would be to look at the try_recv error and
mark the stream Ready when both reports the Disconnected error,
but this cannot be done today as the Command holds a
CommandContext to be able to provide the `spawn` api.

### DIFF
--- a/crux_core/src/command/context.rs
+++ b/crux_core/src/command/context.rs
@@ -21,6 +21,7 @@ pub struct CommandContext<Effect, Event> {
     pub(crate) effects: Sender<Effect>,
     pub(crate) events: Sender<Event>,
     pub(crate) tasks: Sender<Task>,
+    pub(crate) rc: Arc<()>,
 }
 
 // derive(Clone) wants Effect and Event to be clone which is not actually necessary
@@ -30,6 +31,7 @@ impl<Effect, Event> Clone for CommandContext<Effect, Event> {
             effects: self.effects.clone(),
             events: self.events.clone(),
             tasks: self.tasks.clone(),
+            rc: self.rc.clone(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crux_core/src/command/stream.rs
```diff
@@ -44,7 +45,8 @@ where
             return Poll::Ready(Some(CommandOutput::Effect(effect)));
         }
 
-        if self.is_done() {
+        // Check if there is no context alive that could send events in the future
+        if self.is_done() && Arc::strong_count(&self.context.rc) == 1 {
```
charypar: Ohh, snap. That makes sense. Can we move that check into `Command::is_done()`? 

I think you're right that this is another condition for whether anything may still happen with the Command - it's also not done while anyone else is holding the context.

I'm a little bit worried that people might accidentally keep commands dangling with this, but it doesn't have any major impacts, and storing context feels pretty unusal, so... probably okay.

crux_core/src/command/tests/stream.rs
```diff
@@ -0,0 +1,30 @@
...
+    );
+
+    drop(context);
+
+    assert!(
+        command
+            .as_mut()
+            .poll_next(&mut Context::from_waker(Waker::noop()))
+            .is_ready()
+    );
```
charypar: Having the check as part of `is_done` would also make this test a smidge easier to write and read.

crux_core/src/command/stream.rs
DeltaEvo: Ah yeah, good catch ! I started the implementation with looking at the try_recv error and so I wasn't able to put it in is_done

DeltaEvo: Done !

DeltaEvo:
https://github.com/redbadger/crux/pull/425 is required to make the failing test pass

PR #425, opened 2025-08-20
crux_core/command: Clear spawn_queue when a command is aborted

DeltaEvo:
This allows to drop the pending tasks futures.

This was discovered with the failing doctest of https://github.com/redbadger/crux/pull/423


charypar (approved):
Ohh, good catch!

PR #427, opened 2025-08-27
Support POST/PUT/PATH with body from shell

cl11e:
without this code it send body size 0 to the server.


StuartHarris:
Hello @cl11e — thank you for this contribution. You are correct in that we're currently not setting the body in the examples. I've since updated all the examples, so am going to close this PR.
Incidentally, I decided that it's probably more correct to set the body for all HTTP methods — even though really methods like GET don't need a body, all HTTP messages are allowed to have one.

PR #430, opened 2025-08-29
Support for Kotlin typegen

StuartHarris:
Add support for the new Kotlin type generation in [`facet_generate`](https://github.com/redbadger/facet-generate), and update the `counter-next` example to use Kotlin typegen instead of Java in the Android shell.

Note: also back-revs some deps to be equal to that of crux_core 0.16, so that we don't accidentally break users when they update to 0.16.1

### DIFF
--- a/examples/counter-next/shared/src/bin/codegen.rs
+++ b/examples/counter-next/shared/src/bin/codegen.rs
@@ -18,8 +18,10 @@ fn main() -> Result<()> {
 
     info!("Generating code for Serde");
     serde(&out_dir.join("serde"))?;
+
     info!("Generating code for Server Sent Events");
     sse(&out_dir.join("sse"))?;
+
     info!("Generating code for App");
     app(&out_dir.join("app"))?;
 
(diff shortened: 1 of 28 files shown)


examples/counter-next/shared/src/bin/codegen.rs
```diff
@@ -145,8 +147,8 @@ fn serde(out_dir: &Path) -> Result<()> {
             .build(),
     )?;
 
-    typegen_serde.java(
-        &Config::builder("com.crux.example.counter.serde", out_dir.join("java"))
+    typegen_serde.kotlin(
+        &Config::builder("com.crux.example.counter.serde", out_dir.join("kotlin"))
```
charypar, line 151: Wonder if in the long run, we should move this package splitting into a separate example dedicated to these kinds of shared module shenanigans, so we don't make people think this is what the new typegen requires them to do 🤔

StuartHarris: Good point! We should deffo do that before this example replaces the old one. Probably do a cleanup at the same time as we update all the docs

PR #432, opened 2025-09-02
crux_cli: changed bindgen args (--kotlin, --swift) to accept out dirs

StuartHarris:
Breaking change in `crux_cli`.

This will probably need a breaking bump to `crux-core` as it reexports this from `crux_cli`, so I took the opportunity to also update deps.

The `--swift` and `--kotlin` arguments to `bindgen` now accept an optional output directory:

```
Usage: crux bindgen [OPTIONS]

Options:
  -c, --crate-name <STRING>  Package name of the crate containing your Crux App [default: shared]
  -k, --kotlin <DIR>         Generate bindings for Kotlin at the specified path
  -s, --swift <DIR>          Generate bindings for Swift at the specified path
  -h, --help                 Print help
  -V, --version              Print version
```

### DIFF
--- a/crux_cli/CHANGELOG.md
+++ b/crux_cli/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.0](https://github.com/redbadger/crux/releases/tag/crux_cli-v0.2.0) - 2025-09-02
+
+Breaking changes:
+
+The `--swift` and `--kotlin` arguments to `bindgen` now accept an optional output directory:
+
+```
+Usage: crux bindgen [OPTIONS]
+
+Options:
+  -c, --crate-name <STRING>  Package name of the crate containing your Crux App [default: shared]
+  -k, --kotlin <DIR>         Generate bindings for Kotlin at the specified path
+  -s, --swift <DIR>          Generate bindings for Swift at the specified path
+  -h, --help                 Print help
+  -V, --version              Print version
+```
+
 ## [0.1.1](https://github.com/redbadger/crux/releases/tag/crux_cli-v0.1.1) - 2025-08-19
 
 Added support for calling bindgen through the library API.
--- a/crux_cli/Cargo.toml
+++ b/crux_cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crux_cli"
-version = "0.1.1"
+version = "0.2.0"
 description = "Command line tool for crux_core"
 authors.workspace = true
 edition.workspace = true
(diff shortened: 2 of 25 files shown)


crux_cli/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde = { workspace = true, features = ["derive"] }
 serde-generate = "0.32.0"
 serde-reflection = "0.5.0"
 serde_json = "1.0.143"
-thiserror.workspace = true
+thiserror = "2.0.16"
```
charypar: Intentionally decoupling the versions from the workspace one? Seems reasonable, just curious what your thinking is

crux_core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "crux_core"
-version = "0.16.1"
+version = "0.16.2"
```
charypar: Are we doing a release as part of this? 🤔

crux_cli/CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.0](https://github.com/redbadger/crux/releases/tag/crux_cli-v0.2.0) - 2025-09-02
+
+Breaking changes:
+
+The `--swift` and `--kotlin` arguments to `bindgen` now accept an optional output directory:
```
charypar, line 14: If I'm reading the code correctly, the presence or absence of the argument also decides whether or not the bindings for the language get generated? 

We should probably be more explicit about that - it's not that they accept optional output directory, it's that generating the code is optional, and requires an ouput path when selected.

charypar (approved):

crux_cli/Cargo.toml
StuartHarris: my thinking was not that exciting. The cli is published independently and I didn't wan't to bump the workspace versions of anything as that then becomes the min version for any minor bumps on crux core, which I've seen cause problems downstream when Cargo can't find a suitable version for a dependency that comes from multiple places.

crux_core/Cargo.toml
StuartHarris: I was, but decided against it, bumping `facet_generate` instead. Will revert.

crux_cli/CHANGELOG.md
StuartHarris: Yeah good point. The help helps, but still confusing. It _is_ more ergonomic though. And if we split each into two args (a switch and an out dir) then we then have the problem of communicating to the user that they have to use both, or none, of each. Can you think of a better way?

charypar: Oh I think the behaviour is fine, we just need to be explicit about it in the various docs and help texts. If there isn't one (can't remember now), might be worth adding a warning if none of the flags are selected

crux_cli/Cargo.toml
StuartHarris: Sorry, this is a lie. `crux_cli` is now re-exported from `crux_core` so they are all tyied together 😓 . Will revert

crux_cli/CHANGELOG.md
StuartHarris: modified to put the languages into a required group, with multiple allowed, so at least one language is now required

PR #436, opened 2025-09-14
[Breaking] Fix effect middleware to support stream requests and multipl resolutions

### LINKED ISSUE #428: EffectMiddleware makes it impossible to resolve stream effects more than once
The trait defines the callback as taking the request handle by move, but the middleware needs to also keep it in order to resolve it again later on.

charypar:
Fixes #428, by adjusting the effect callback in middleware to take request handle by mutable borrow, instead of value.

### DIFF
--- a/crux_core/src/middleware/effect_handling.rs
+++ b/crux_core/src/middleware/effect_handling.rs
@@ -39,7 +39,7 @@ where
     fn try_process_effect_with(
         &self,
         effect: Effect,
-        resolve_callback: impl Fn(
+        resolve_callback: impl FnMut(
             &mut RequestHandle<<Self::Op as Operation>::Output>,
             <Self::Op as Operation>::Output,
         ) + Send
--- a/crux_core/src/middleware/effect_handling.rs
+++ b/crux_core/src/middleware/effect_handling.rs
@@ -39,8 +39,8 @@ where
     fn try_process_effect_with(
         &self,
         effect: Effect,
-        resolve_callback: impl FnOnce(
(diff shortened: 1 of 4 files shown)


crux_core/src/middleware/effect_handling.rs
```diff
@@ -39,8 +39,8 @@ where
     fn try_process_effect_with(
         &self,
         effect: Effect,
-        resolve_callback: impl FnOnce(
-            RequestHandle<<Self::Op as Operation>::Output>,
+        resolve_callback: impl Fn(
```
StuartHarris: Can this be `impl FnMut`, which is less restrictive than `Fn`?

StuartHarris (changes requested):
Looks great 😍. Just one question.

crux_core/src/middleware/effect_handling.rs
charypar: Done! I think I will never remember which way the restrictiveness goes 😅

StuartHarris: IKR 😓

StuartHarris (approved):
❤️

crux_core/src/middleware/effect_handling.rs
StuartHarris: This is how I think about it. `FnOnce` is the widest set, because it also includes closures that can only be called once because they move data out into their environment. `FnMut` is the next widest because it includes closures that mutate their environment. `Fn` only includes closures that don't move out or mutate. And finally `fn()` is the most restrictive because it's a single type.

PR #439, opened 2025-10-20
bridge: remove `erased_serde` dependency and usage

uael:
`erased_serde` is adding a good amount of bloat to the end binary, and prevent us to improve the registry, like the ability to resolve with typed responses (see https://github.com/redbadger/crux/pull/440).

In order to remove it, re-use and unify the concept of `FfiFormat` from the middlware module and make `ResolveSerialized` generic over it to allow for generic deserialization without erasing the types.

This is both a breaking change and regresssion as it remove the ability to pass custom `Deserializer` (e.g. serde_json::Value) in favor of just bytes. The later use case feeling niche enough to be removed.

### DIFF
--- a/crux_cli/src/codegen/formatter.rs
+++ b/crux_cli/src/codegen/formatter.rs
@@ -471,7 +471,7 @@ impl From<&Type> for Format {
                 args: _,
                 self_type: _,
                 trait_: _,
-            } => Format::TypeName(name.to_string()),
+            } => Format::TypeName(name.clone()),
         }
     }
 }


charypar:
Thanks for submitting these, I will review as soon as I can, but realistically, tomorrow night is my earliest chance, sorry to keep you waiting!

crux_core/src/bridge/mod.rs
```diff
@@ -31,30 +55,32 @@ where
 
 /// Bridge is a core wrapper presenting the same interface as the [`Core`] but in a
 /// serialized form, using bincode as the serialization format.
-pub struct Bridge<A>
+pub struct Bridge<A, T = BincodeFfiFormat>
```
charypar, line 58: Not major, but can we call the type parameter F or Format, plase?

crux_core/src/bridge/mod.rs
```diff
@@ -184,19 +193,16 @@ where
     /// # Errors
     ///
     /// Returns an error if the event could not be deserialized.
-    pub fn process_event<'de, D, S>(&self, event: D, requests_out: S) -> Result<(), BridgeError>
+    pub fn process_event<'a>(
+        &self,
+        event: &'a [u8],
+        requests_out: &mut Vec<u8>,
+    ) -> Result<(), BridgeError<T>>
```
charypar, line 200: Is the return via mutable argument essentially the only difference between the two bridge implementations? 

If so, can we also rename `BridgeWithSerializer` to something more descriptive? 

More broadly, I'm quite tempted to say let's just make this interface the default - it's a small change in the FFI and allows more flexibility in the memory ownerhship.

@StuartHarris what do you think, can you think of any downsides?

crux_core/src/bridge/registry.rs
```diff
@@ -13,15 +13,15 @@ use crate::{EffectFFI, ResolveError};
 #[facet(transparent)]
 pub struct EffectId(pub u32);
 
-pub struct ResolveRegistry(Mutex<Slab<ResolveSerialized>>);
+pub struct ResolveRegistry<T: FfiFormat>(Mutex<Slab<ResolveSerialized<T>>>);
```
charypar, line 16: Same minor comment on generic naming

uael:
Thank you both for your reviews. Renaming might be nice! Feel free to take the ownership to apply @charypar suggestions or any other tweaks as I may not have the bandwidth this week

PR #440, opened 2025-10-20
bridge: allow from typed events and responses

uael:
⚠️ To be rebased on top of https://github.com/redbadger/crux/pull/439

This allows to resolve ongoing requests with typed responses, enabling
un-serializable response types like target specific objects:
```rust
struct Canvas {
  #[serde(skip)]
  inner: web_sys::HtmlCanvasElement
}

bridge.handle_response_typed(id, Canvas { inner }, &mut effects_bytes)?;
```

Add the `Response` enum which is ever a bag of bytes or a `Box<dyn Any>`
and change the `ResolveSerialized::resolve` input to this new type. When
handling the response we try to downcast the passed `Any` to the
expected output type, trowing out a new `BrideError::OutputTypeMismatch`
on failure.

### DIFF
--- a/crux_core/src/bridge/mod.rs
+++ b/crux_core/src/bridge/mod.rs
@@ -12,7 +12,7 @@ use crate::{App, Core, core::ResolveError};
 pub use formats::{BincodeFfiFormat, JsonFfiFormat};
 pub use registry::EffectId;
 pub(crate) use registry::ResolveRegistry;
-// ResolveByte and Response are public to be accessible from crux_macros
+// ResolveSerialized and Response are public to be accessible from crux_macros
 #[doc(hidden)]
 pub use request_serde::{ResolveSerialized, Response};
 


charypar:
@uael sorry, I've been all over the place in the past few weeks! 

Thanks for the extra detail, I think I get what you're after now. In all honesty I'm not sure that use is common enough to deserve support in the main interface to the core. 😕 I'll keep thinking about it. 

One thing we've successfully used in the past was adding a special (even typed) FFI "side channel" for those kinds of values (I think it was a Core Text font pointer in that case, but similar idea), which would place them into a dedicated "store" and return a serializable handle to them, which could be used in the main Crux FFI. 

The transaction would be something like:
1. Shell placess a `web-sys` value into the store using the dedicated FFI, gets a handle in return
2. Shell resolves the effect request with the handle as output
3. Capability exchanges the handle for the `web-sys` value with the store and returns it to the app code

It is a bit more complicated and adds an FFI call, so it depends on your specific situation, but it should be possible to hide most of the complications from the majority of the application code and make it look like the `web-sys` object is being passed through directly.

I'd love to come up with a general way to do support this kind of idea for handing over ... basically pointers directly, but haven't quite worked out what a general one might look like with at least _some_ safety.

mhedgpeth:
It's looking like this has gone stale, so I'm going to close it. Feel free to discuss on Zulip, let's align on a ticket in GitHub, then let's reopen.

Issue #443, opened 2025-10-22
Remove deprecated capability runtime


charypar:
Remove the deprecated types, and the remaining infrastructure around QueuingExecutor. The Compose capability can also be removed alongside all of the capability module. 

The executor and command_spawner in Core should be replaced by a root `Command` which can double as an executor.

PR #448, opened 2025-10-27
Remove deprecated capability support

charypar:
This removes the capability support ahead of the 0.17 release (#443).

# Steps
- [x] Remove deprecated components
- [x] Remove all old capability implementations
- [x] Move the capability implementations from their `command` modules to their roots and deprecate `command::*` exports
- [x] Remove the old executor and internals
- [x] Update any examples using path dependencies

### DIFF
--- a/crux_core/src/testing.rs
+++ b/crux_core/src/testing.rs
@@ -298,10 +298,27 @@ where
     pub fn expect_one_event(&mut self) -> Event {
         let mut events = self.events();
         match (events.next(), events.next()) {
-            (None, _) => panic!("expected one effect but got none"),
+            (None, _) => panic!("expected one event but got none"),
             (Some(event), None) => event,
             _ => panic!("expected one effect but got more than one"),
         }
+    }
+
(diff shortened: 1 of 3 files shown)


crux_core/src/capability/channel.rs
StuartHarris, line 1: can this file be removed now?

crux_core/src/capability/executor.rs
StuartHarris, line 1: can this file be removed now?

crux_core/src/capability/mod.rs
StuartHarris, line 1: I think the only thing we need from this file now is the `Operation` trait

crux_core/src/testing.rs
StuartHarris, line 1: Can the `AppTester` go now?

charypar: Deprecated it and migrated the remaining tests

StuartHarris (approved):
really nice. Thanks ❤️

PR #449, opened 2025-10-28
Update dependencies (prep for v0.17.0)

StuartHarris:
Updates Rust dependencies, including namespace inheritance support from `facet_generate` v0.12.0

### DIFF
--- a/crux_core/tests/middleware.rs
+++ b/crux_core/tests/middleware.rs
@@ -493,6 +493,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "TODO: occasionally fails probably dues to effect ordering, fix this"]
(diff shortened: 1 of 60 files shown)


crux_core/tests/middleware.rs
```diff
@@ -493,6 +493,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "TODO: occasionally fails probably dues to effect ordering, fix this"]
```
charypar, line 496: Oh, I was suspecting this of discovering a problem with my capability runtime migration, good to know it's a general problem

StuartHarris: Yes, noticed it on another PR related to adding derive Facet in crux_time. I look at it in another PR :-)

PR #450, opened 2025-10-29
Update the `bridge_echo` example ready for 0.17

StuartHarris:
- [x] removes shared types lib
- [x] adds `CoreFFI`
    - [x] uses `crux_core::bridge::Bridge` rather than  `crux_core::middleware::Bridge`
    - [x]  introduces new features for `uniffi` and `wasm_bindgen`, which means that `web-leptos` and `cli` shells no longer include a bridge
- [x] add codegen binary, for facet based typegen of Swift, Kotlin, TypeScript, and bindgen for Android
- [x] add `./build.sh`
- [x] update `iOS` app, uses `uniffi` feature
- [x] update `Android` app (including deps, moving from gradle to kts), uses `uniffi` feature
- [x] update `web-remix` (including wasm package in generated folder), uses `wasm_bindgen` feature
- [x] update `web-leptos`, doesn't use `uniffi` or `wasm_bindgen` features, so no bridge
   - [x] adds display of average events per second
   - [x] updates workspace resolver to "2", and now uses latest version of `leptos`
- [x] update cli, doesn't use `uniffi` or `wasm_bindgen` features, so no bridge
- [x] adds a `Justfile` for each shell, so that you can just `cd` into the relevant directory and run `just dev` to get going.


examples/bridge_echo/Android/shared/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+                    File(
+                        File(layout.buildDirectory.asFile.get(), "rustJniLibs"),
+                        buildTask.toolchain?.folder!!
+                    )
+                )
+                dependsOn(buildTask)
+            }
+        }
+    }
+}
```
charypar, line 108: Somewhere here there might be an opportunity to run the bindgen and typegen commands. Not sure of the specifics, but seems like a good idea to rely on the build tooling of the shells to run those commands, maybe?

examples/bridge_echo/shared/src/app.rs
```diff
@@ -4,17 +4,19 @@ use crux_core::{
     render::{render, RenderOperation},
     Command,
 };
+use facet::Facet;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+#[derive(Facet, Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
```
charypar, line 11: How much do we want to use facet in the examples, while we're still not quite recommending to use it? Or are we recommending it?

examples/bridge_echo/shared/src/ffi.rs
charypar, line 1: Similar question - are we defaulting examples to the middleware FFI when we're considering it optional?

examples/bridge_echo/web-remix/app/routes/_index.tsx
```diff
@@ -18,6 +18,18 @@ export const meta = () => {
...
+  useEffect(
+    () => {
+      // There may be a nicer way using https://react.dev/reference/react/useSyncExternalStore
+      if (core.current === null) {
+        core.current = new Core(setView);
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    /*once*/ []
+  );
```
charypar, line 31: We could look at making a `useCrux` hook that returns an `update` function and a `view` which is promised to be up to date. 

A future nice to have though.

examples/bridge_echo/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# package our shared library for iOS, including bindgen for Swift
+mkdir -p shared/generated/swift
+pushd shared/generated/swift
+cargo swift package --name Shared --platforms ios
+rm -rf generated
+popd
+
+# run typegen for Swift/Kotlin/TypeScript and bindgen for Kotlin
+RUST_LOG=info cargo run --package shared --bin codegen --features cli,facet_typegen
```
charypar: Related to the above comment - wonder if we should move the relevant commands into the build automation of the shells?

examples/bridge_echo/shared/src/ffi.rs
StuartHarris: Have updated this PR to use `crux_core::bridge::Bridge` rather than  `crux_core::middleware::Bridge`, much simpler

examples/bridge_echo/build.sh
StuartHarris: I added a `Justfile` for each shell, so that you can just `cd` into the relevant directory and run `just dev` to get going. This obvs requires `brew install just` but I think it's worth it to get going _really_ quickly.

examples/bridge_echo/Android/shared/build.gradle.kts
StuartHarris: we did have bindgen and typegen in here before. Reluctant to put them back. Not a fan of jumping through all the extra hoops for gradle (and ios) builds. Would rather do that outside (e.g. the `Justfile` is way clearer in its intent). The actual library build for android can probably stay in here (at least for now), although that is also now outside the XCode build for iOS.

examples/bridge_echo/shared/src/app.rs
StuartHarris: I think it should become the new default (it's the only choice if you want Kotlin, rather than Java). I think we just need to make sure that the docs are clear about still being able to use the `serde_reflection` based typegen.

examples/bridge_echo/web-remix/app/routes/_index.tsx
StuartHarris: Yeah nice idea. Deffo for a future PR. Need to think about what the story will be for shell-side code distribution.

crux_cli/src/bindgen/mod.rs
```diff
@@ -25,6 +25,7 @@ pub(crate) fn bindgen(args: &BindgenArgs) -> Result<()> {
     let config_supplier = CrateConfigSupplier::from(metadata);
 
     if let Some(out_dir) = &args.languages.kotlin {
+        println!("{}", out_dir.display());
```
charypar: Leftover?

crux_core/tests/middleware.rs
```diff
@@ -548,6 +548,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "TODO: occasionally fails probably dues to effect ordering, fix this"]
```
charypar, line 553: I shall 🫡

examples/bridge_echo/Android/shared/build.gradle.kts
charypar: I suppose this is a perspective thing - are we favouring the perspective of the core, Rust developer (then Justfile should drive), or the perspective of the shell, Kotlin developer (then gradle should drive). 

Hard to say, happy with favouring the Rust developer.

examples/bridge_echo/shared/src/app.rs
charypar: If we're ready to commit to (more or less) backwards compatibility (in both API and code output) from here on out, then lets consider it a default.

examples/bridge_echo/shared/src/ffi.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[cfg_attr(
+        feature = "wasm_bindgen",
+        wasm_bindgen::prelude::wasm_bindgen(constructor)
+    )]
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            core: Bridge::new(Core::new()),
+        }
+    }
```
charypar, line 34: So, we're going with a middleground then as default: No callback or middleware, but shell owns the core instance. Happy with that, just checking my understading.

examples/bridge_echo/Android/shared/build.gradle.kts
StuartHarris: Yeah me too. We are Rustaceans 🦀 . A shell developer can call the relevant Just task from their build if they want to.

crux_cli/src/bindgen/mod.rs
StuartHarris: thanks

examples/bridge_echo/shared/src/app.rs
StuartHarris: I _think_ so. It's certainly better for us to have the control over the typegen.

PR #452, opened 2025-10-30
Simple Counter macOS example

### LINKED ISSUE #248: Crux with macOS and SwiftUI
Is it possible to create macOS apps using SwiftUI and crux. If so could we get a simple example?

graemeasher:
Hi Stuart & Viktor - I have created an example of a MacOS implementation of the Simple Counter example (.examples/simple_counter/macOS). I chatted with Viktor on Zulip and he said to submit a PR. I see an existing issue related to this (Crux with macOS and SwiftUI #248) in case this Closes #248.

I followed your existing docs (latest_master) using the iOS - Swfit and SwifUI - using XcodeGen to build this macOS example. There are only two difference from the iOS instructions:

1. Within the project.yml I changed the bundleIdPrefix slightly to avoid conflict with the iOS example. I also removed a handful of properties that do not pertain to macOS. 
2. Within the SimpleCounter.xcodeproj > General > Supported Destinations - I updated this to only use the Mac + macOS SDK as the destination.

I have tested using deploymentTargets of 15 (matches existing iOS example) as well as 26 and everything worked fine.  One issue I did encounter was that I could not have both the iOS and macOS projects open at the same time in Xcode as they both reference the same shared binary and this would cause a build error for whichever of the two projects was opened second in Xcode. Quitting Xcode and reopening one project at a time resulted in clean builds.


graemeasher:
@StuartHarris - Sounds good. I saw the new codegen being used in the counter-next example but built this using the previous method as the rest of the simple-counter used that. I look forward to the docs on the updated approach as it looks like a nice evolution for Crux.

PR #455, opened 2025-11-13
Update cat_facts example for Crux 0.17, and update APIs for crux_kv, crux_platform and crux_time

StuartHarris:
_note: these are breaking changes_

## `crux_kv`, `crux_platform`, `crux_time`

- [x] Add `#[derive(Facet)]` where missing
- [x] Move all protocol types to their own module, to make it clear that these are the only types that need to implement `serde` and `facet` traits.

## `cat_facts` example

- [x] removes shared types lib
- [x] adds `CoreFFI`
    - [x] uses `crux_core::bridge::Bridge`
    - [x]  introduces new features for `uniffi` and `wasm_bindgen`, which means that `web-yew` and `cli` shells no longer include a bridge
- [x] add codegen binary, for facet based typegen of Swift, Kotlin, TypeScript, and bindgen for Android
- [x] use `resolver = "3"`
- [x] Use `#[facet(opaque)]` to allow us to use types that don't implement the `Facet` trait, in event variants that don't leave the shell (depends on the newly released `facet_generate` v0.12.1)
- [x] update `iOS` app, uses `uniffi` feature
- [x] update `Android` app (including deps, moving from gradle to kts), uses `uniffi` feature
- [x] update `web-nextjs` (including wasm package in generated folder), uses `wasm_bindgen` feature
- [x] update `web-yew`, doesn't use `uniffi` or `wasm_bindgen` features, so no bridge
- [x] update cli, doesn't use `uniffi` or `wasm_bindgen` features, so no bridge
- [x] adds a `Justfile` for each shell, so that you can just `cd` into the relevant directory and run `just dev` to get going.


crux_kv/src/lib.rs
```diff
@@ -87,26 +85,28 @@ impl std::fmt::Debug for KeyValueOperation {
...
-    Set { previous: Value },
+    Set(Option<Vec<u8>>),
     /// Response to a `KeyValueOperation::Delete`,
     /// returning the value that was previously stored under the key, may be empty
-    Delete { previous: Value },
+    Delete(Option<Vec<u8>>),
     /// Response to a `KeyValueOperation::Exists`,
     /// returning whether the key is present in the store
-    Exists { is_present: bool },
+    Exists(bool),
```
charypar: I'm reasonably sure we're going back to what we once had, which we changed to be easier to work with in serialized form (esp. in JSON) and more explicit about what the carried values are, especially for the `previous` case. 

These types are the FFI protocol types, they are not meant to be used directly by a Rust engineer, ergonomics don't matter.

crux_kv/src/command.rs
```diff
@@ -24,38 +20,34 @@ where
     /// Read a value under `key`
     pub fn get(
         key: impl Into<String>,
-    ) -> RequestBuilder<Effect, Event, impl Future<Output = DataResult>> {
+    ) -> RequestBuilder<Effect, Event, impl Future<Output = KeyValueResult>> {
```
charypar: I really don't like this. Not only is it a breaking change, but it's in the core facing API, which should be as rust-native as it can possibly be. `KeyValueResult` is the protocol type, designed to serialize well, but it's not a rust `Result`, can't be mapped, `?` won't work on it....

crux_kv/src/value.rs
```diff
@@ -1,37 +0,0 @@
...
-/// The value stored under a key.
-///
-/// `Value::None` is used to represent the absence of a value.
-///
-/// Note: we can't use `Option` here because generics are not currently
-/// supported across the FFI boundary, when using the builtin typegen.
-#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
-pub enum Value {
-    None,
-    Bytes(#[serde(with = "serde_bytes")] Vec<u8>),
```
charypar, line 12: This was important for two reasons: 

1. signalling not found
2. serializing bytes better with formats that have special handling for them.

crux_time/src/command.rs
```diff
@@ -47,13 +43,13 @@ where
     /// # Panics
     /// Panics if the response is not `TimeResponse::Now`.
     #[must_use]
-    pub fn now() -> RequestBuilder<Effect, Event, impl Future<Output = SystemTime>> {
+    pub fn now() -> RequestBuilder<Effect, Event, impl Future<Output = Instant>> {
```
charypar: Same idea here - the distinction between the protocol and API types was intentional and the types were chosen to be easier to use from Rust

crux_kv/src/lib.rs
StuartHarris: Fair point, although they do matter, just not as much. :-)

charypar:
Yea, we do need to own the protocol types, and they also need to serialize well into various formats (esp. JSON which will probably be common). That's also what then makes them kinda nasty to use in normal Rust code, they are really more of an FFI format...

StuartHarris:
The confusion was due to inward-facing types needing to implement `Facet` in order to be part of the `Event` enum. I've since updated `facet_generate` to support `#[facet(opaque)]` correctly, so that now fields in structs and in struct or tuple variants can reference types that don't need to implement `Facet`. This means we can use the capabilities correctly (although I moved all the protocol types to a protocol module, and implemented `Facet` on them where missing).

crux_platform/src/lib.rs
charypar, line 1: Since we're making breaking changes... is it maybe time to sunset crux_platform? 🤔

examples/cat_facts/shared/src/app.rs
```diff
@@ -54,34 +55,43 @@ impl Default for CatImage {
...
-    Restore, // restore state
+    LoadState,
 
-    // events local to the core
+    // events local to the core:
+    // we can skip serialization using `#[serde(skip)]`
+    // we can skip typegen using `#[facet(skip)]`
+    // we can use types that do not implement `Facet`, by marking them with `#[facet(opaque)]`
     #[serde(skip)]
+    #[facet(skip)]
```
charypar, line 81: This is a good example, where it would be super helpful if facet_generate could read and respect the serde attributes - skips, renames, etc. 

For as long as we're using facet with serde, we should have a level of compatibility, so we don't have to implement half of serde's attributes and force people to double annotate. 

If facet won't let us read serdes attribute, could we maybe make a shim that translates them? We could then `#[derive(FacetGenerate)]`, which would, in pricniple at least, run our macro to translate serde attributes into facet ones, then derive Facet?

charypar (approved):
Amazing, thanks for doing all that! ❤️

crux_platform/src/lib.rs
StuartHarris: I would love to. It does nothing lol

examples/cat_facts/shared/src/app.rs
StuartHarris: Huh. I hadn't thought of that but a good idea if we can't read them directly. I'll have another look at this.

crux_platform/src/lib.rs
charypar: Hmm, shall we retire it alongside cat facts, notes and tap to pay, once we have a few more shells for Weather...?

examples/cat_facts/shared/src/app.rs
charypar: Don't hold this PR for it though, that's ready to go in, IMO. More of an idea for how to approach the ... facet-serde duality.

Issue #444, opened 2025-10-22
Document the middleware support in FFI


charypar:
The middleware base FFI should be treated as an opt-in, advanced alternative to the standard Bridge, as it's more complex to set up and use.

charypar:
Cancelling this in favour of #525 (which should include relevant docs).

PR #462, opened 2025-11-30
New documentation for 0.17 and beyond

charypar:
This is a full revamp of the doc book. This supersedes #352.

I've restructured it into two main parts:

- Part one for the basic building blocks and the FFI setup, demonstrated on a really basic Counter app
- Part two covering more real-world scenarios on the Weather example.

There is also a part three for other advanced topics and four for explaining internals.

## In this PR I'm aiming to 

- [x] write the Rust side of Part 1
- [x] complete Part 1 for iOS 
- [ ] complete Part 1 for the other three key platforms
- [x] write Part 2
   - [x] intro
   - [x] Elm architecture
   - [x] Structuring larger apps
   - [x] Managed effects
   - [x] Testing effects
   - [x] Building capabilities
- [x] complete Part 2 for iOS 
- [ ] complete Part 2 for Android  (as that's the current extent of the Weather example)
- [ ] update the advanced topics and cover middleware (#444)
- [ ] make sure internals are still up to date

This may take me a while.

@StuartHarris I've embedded bits of code directly in the docs now, but would like to just reference the simple counter, once it's updated to the new setup for FFI. The code _should_ just work as written in the docs. I'm happy to take on the updating of the simple counter as well, but it will take me longer 😅 If you get there first, I'd appreciate the help ❤️

### DIFF
--- a/docs/src/part-3/middleware.md
+++ b/docs/src/part-3/middleware.md
@@ -1,13 +1,33 @@
 # Middleware
 
-Middleware is a relatively new, and still somewhat experimental support
-for split effect handling – handling some effects in the shell, and some
+Middleware is a relatively new, and somewhat advanced feature for split
+effect handling, i.e. handling some effects in the shell, and some
 still in the core, but outside of the Crux state loop.
 
-This is useful when you have an existing 3rd party library written in Rust
+Middleware can be useful when you have an existing 3rd party library written in Rust
 which you want to use, but it isn't written in a sans-I/O way with managed
-effects or otherwise isn't compatible with Crux. This is most libraries
+effects or otherwise isn't compatible with Crux. This is sadly most libraries
 with side effects.
+
+It is quite likely most apps will never need to use middleware. Before reaching for middleware,
+we encourage you to consider:
+
+- Implementing the side-effect in each Shell using native, platform SDKs. Shared libraries give a
+  productivity boost at first, but for the same reason Crux uses Capabilities, they can't always
+  be the best platform citizens, and often rely on very low-level system APIs which compromise the
+  experience, don't collaborate well with platform security measures, etc.
+- Moving coordination logic from the Rust implementation into a custom capability in the core
+  and implementing it on top of lower level capabilities, e.g. HTTP. This would be the case for
+  HTTP API SDK type libraries, but may well not be practical at first
+
+Only if neither of these is a good option, reach for a middleware. The cost of using it is
+that the effect handling becomes less straightforward, which may cause some headaches
+debugging effect ordering, etc.
+
+We are also still learning how middleware operates in the wild, and the API may change more
+than the rest of Crux tends to.
+
+All that said, the feature is used in production with success today and should work well.
 
 ```admonish info
 This chapter is not finished, we're working on it. For examples of middleware use you can


mhedgpeth:
Great to see this! Let me know when it's time to look it over

charypar:
Okay! This is ready for review.

I've written the main journey through parts 1 and 2 including iOS shells. What still needs doing is:
* Other shells in part 1 and 2
* Revise parts 3 and 4 for updates in 0.17
* Document middleware in full, not just by refering to code

For reviewers, I'd specifically like feedback on:
* Understandability and learning curve
* Do instructions actually work as provided

And of course any other general feedback!

docs/src/part-1/shell/ios/index.md
```diff
@@ -0,0 +1,141 @@
...
+[`cargo-swift`](https://crates.io/crates/cargo-swift) to generate an
+swift package for our shared library, which we can add in Xcode.
+
+You can install it with
+
+```bash
+cargo install cargo-swift
+```
+
+To run the variou steps, we'll also use the [Just]() task runner.
```
mhedgpeth: `various`

docs/src/part-1/shell/ios/index.md
```diff
@@ -0,0 +1,141 @@
...
+
+The shared types are generated by Crux as a Swift package, which we can add to
+our iOS project as a dependency. The Swift code to serialize and deserialize
+these types across the boundary is also generated by Crux as Swift packages.
+
+![build flow](./flow.svg)
+
+## Compile our Rust shared library
+
+When we build our iOS app, we also want to build the Rust core as a static
```
mhedgpeth, line 26: One thing that threw me off at first was that this I believe is a one-time thing. You need to generate the project over and over again with xcodegen but after today you never need cargo swift again. You might consider making that clearer.

mhedgpeth: Oh as I read below perhaps this has changed? Or I have forgotten? Leaving this here though so you understand a potential confusion point.

docs/src/part-1/shell/ios/index.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+And finally, make sure `iOS/SimpleCounter/SimpleCounterApp.swift` looks like this to use
+the `ContentView`:
+
+```swift
+{{#include ../../../../../examples/simple_counter/iOS/SimpleCounter/SimpleCounterApp.swift}}
+```
+
+The one interesting part of this is the `@ObservedObject var core: Core`. Since the `Core` is
```
mhedgpeth, line 130: You might also include that swift is smart enough to know to only re-render the views that match elements _within_ the `ViewModel` that are updated, so you don't get a full refresh with this model.

docs/src/part-2/shell/android.md
```diff
@@ -0,0 +1,3 @@
+# Android
+
+TODO
```
mhedgpeth: Do you want a different message here? Perhaps:

```
We're working on a full Android example, if you want to contribute and speak Kotlin, feel free to create a PR. We hope you can fill in this blank with the other examples.
```

docs/src/part-2/capabilities.md
```diff
@@ -0,0 +1,93 @@
+# Building capabilities
+
+The final piece of the puzzle we should look at in our exploration
+of the Weather app before we move to the Shell is Capabilities.
+
+We looked at effects a fair bit and explored the `Command`s and `CommandBuilder`s,
+but in practice, it's quite rare that you'd interact with those directly from
+your app.
+
+Typically, you'll be working with effects using Capabilities.
```
mhedgpeth: You might define them, since you capitalize this term but I don't believe they exist anymore explicitly in code, you could say:

```
Capabilities are a set of commands that form a group which allows you to reason about your effects from a higher level, for example to express "this is a time-management effect" rather than "start a timer". This makes the app much easier to reason with.
```

Another thing you could mention is some examples of capabilities:

1. Http - `crux_http`
2. Timers/Time management - `crux_time`
3. Secrets - write your own, accessing the key store
4. Environment Variables - tune your app based on environment variable values
5. Navigation to urls - ask your shell to navigate to a url
6. Render

This is from my app, and it could illustrate the kinds of things that capabilities end up being perhaps?

docs/src/part-2/capabilities.md
```diff
@@ -0,0 +1,93 @@
...
+which says that the Effect type needs to be able to convert from a location Request, or
+in other words - we need to be able to wrap a `Request<LocationOperation>` into the app's
+Effect type. All Effect types generated with the `#[effect]` macro already do this.
+
+Other than that, the APIs just create command builds and return them. Those types are also
+somewhat gnarly, but it's mostly the `impl Future<Output = [value]>`, that's interesting.
+Notice that the Output types are not `LocationResult`, they are the specific convenient
+type the Core developer wants.
+
+And that's all Capabilities do - they provide a convenient API for creating `CommandBuilder`s,
```
mhedgpeth, line 92: Now that I read this, a restatement of this definition might help toward the top

docs/src/part-3/middleware.md
```diff
@@ -0,0 +1,17 @@
+# Middleware
+
+Middleware is a relatively new, and still somewhat experimental support
```
mhedgpeth: It's in production on my app, happily!

Perhaps we list out the gaps?

Oh I see that you're deferring this for later, which makes sense.

If it were me I'd say:

```
Middleware is a relatively new feature that exists to allows all of your shells to delegate effects processing to a shared rust struct that does it for you.

It's useful when you want to share code, or when the complexity of your shell processing makes duplication in each shell untenable. However, we encourage you to think about making effects very dumb and to run everything in the model when at all possible.

The basic feature exists now, and we are hardening it in upcoming releases with fixes edge cases based on load and heavy usage. It remains an advanced feature that most users likely will never need.

For examples on middleware you can ... (admonish part)
```

docs/src/part-1/shell/ios/index.md
charypar: I'm pretty sure cargo swift is needed every time to produce the latest version of the shared package (with the binary in). Cargo xcode is technically probably _not_ needed every time, but given how stable xcode project files are, we're probably better off doing it every time anyway.

I'm not sure whether discussing that level of detail in this tutorial is really necessary – it will probably confuse people more than help 🤔 I think of this as a "just follow these steps and don't ask too many questions, and you're gonna have a good time" 😄

docs/src/part-1/shell/ios/index.md
charypar: I'm not 100% sure how much that is true - there is definite nuance in there about which SwiftUI views _rerun_ in response to changes, and what actually ends up updated on screen (because the view end up returning an identical struct), and I don't feel confident enough to make any claims here tbh 😅 

What I do know is that the observability doesn't automatically nest. Basically the view gets told that the `viewModel` property has changed (through the Combine version of KVO) and the rest is based on comparing the props on Views, I think 🤔 I'm also not 100% sure on what kind of equality Swift will use by default... I think `Equatable`? 

In any case, much nuance here, wouldn't wanna open that can of worms

docs/src/part-2/shell/android.md
charypar: I'm secretly hoping someone will step up and copy/paste/adapt the iOS version very quickly 👀 but yes, an invitation to do so is probably a good idea. Will add.

docs/src/part-3/middleware.md
charypar: Let me see about rewording it without sounding too dramatic 👍🏻

charypar: okay, done in 81e3efa

docs/src/part-2/capabilities.md
charypar: had a go at a better intro in f0f025c

docs/src/part-2/shell/android.md
charypar: Mde a slightly more inviting version in  ef2ada0

docs/src/part-2/capabilities.md
mhedgpeth: Great job I like that new description

docs/src/part-3/middleware.md
mhedgpeth: Really great description

mhedgpeth:
This looks ready to me, thanks again @charypar

PR #463, opened 2025-12-01
Update `simple_counter` for 0.17

StuartHarris:
Stacked on top of #460 

- [x] web shells
- [x] Android
- [x] iOS


examples/simple_counter/Android/shared/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
...
+    // `rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android`
+    targets = listOf("arm", "arm64", "x86", "x86_64")
+    extraCargoBuildArguments = listOf(
+        "--package", "shared",
+        "--features", "uniffi"
+    )
+    cargoCommand = System.getProperty("user.home") + "/.cargo/bin/cargo"
+    rustcCommand = System.getProperty("user.home") + "/.cargo/bin/rustc"
+    pythonCommand = "python3"
+}
```
charypar, line 89: Does this build the core as part of the Android build? Didn't we say at some point that we don't want to do that and instead want to drive this build from the Justfile? I can't really remember what we decided 😓

examples/simple_counter/shared/src/app.rs
```diff
@@ -24,16 +26,16 @@ pub struct Model {
...
 
-#[derive(Serialize, Deserialize, Clone, Default)]
+#[derive(Facet, Serialize, Deserialize, Clone, Default)]
 pub struct ViewModel {
     pub count: String,
 }
 
 #[derive(Default)]
-pub struct Counter;
+pub struct App;
```
charypar: Moving away from naming the app types meaningfully? I've probably got a minor preference for `Counter` or `Weather` over just `App`, but no strong opinions.

examples/simple_counter/shared/src/bin/codegen.rs
```diff
@@ -0,0 +1,76 @@
...
+                    .build(),
+            )?;
+
+            info!("Bindgen for Kotlin");
+            bindgen(
+                &BindgenArgsBuilder::default()
+                    .crate_name(env!("CARGO_PKG_NAME").to_string())
+                    .kotlin(&args.output_dir)
+                    .build()?,
+            )?;
```
charypar, line 62: How come Kotlin is the only one that needs bindings built? 👀

examples/simple_counter/web-nextjs/src/app/core.ts
```diff
@@ -1,45 +1,96 @@
...
-  event.serialize(serializer);
+export type Timers = {
+  [key: number]: number;
+};
 
-  const effects = process_event(serializer.getBytes());
+export type SyncMessage = {
+  kind: "change" | "reset";
+  data?: number[];
+};
```
charypar: Is that used?

examples/simple_counter/Android/shared/build.gradle.kts
StuartHarris: Yeah, good point. Android is the only one that still does it this way. I'll fix that

examples/simple_counter/shared/src/app.rs
StuartHarris: I don't really have a preference so will rename it ;-)

examples/simple_counter/shared/src/bin/codegen.rs
StuartHarris: Only Swift and Kotlin need uniffi bindings and the Swift bindings are created by `cargo swift`

examples/simple_counter/Android/shared/build.gradle.kts
StuartHarris: Gonna leave this for now, as the rust-android-gradle-plugin needs to sit in the gradle build file for now.

PR #473, opened 2025-12-18
`crux_http`, remove need for fork of http_types

StuartHarris:
Because we can now stop Facet needing to recurse down into types from the `http_types` crate, we no longer need our temporary fork. (_edit: we still need it for emscripten, as the fork changes a u128 to u64 in the. trace id_)

Also updates to `facet` v0.31 and `facet_generate` v0.13.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,11 +22,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    if: |
-      ${{ github.event_name != 'pull_request' ||
-      github.event.pull_request.head.repo.full_name !=
-      github.event.pull_request.base.repo.full_name }}
-
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


ManevilleF:
Small note, why do we have `=0.31` for `facet` ? Strict versions prevent dependency patching and force forking or having duplicate depdendencies with different versions

StuartHarris:
> Small note, why do we have `=0.31` for `facet` ? Strict versions prevent dependency patching and force forking or having duplicate depdendencies with different versions

As I understand it, by not specifying the patch version, it is not included in the pin. When I do a `cargo tree` of e.g. the `counter_next` example I see `facet v0.31.8`. 

I like it as it stops e.g. `cargo upgrade --incompatible allow` from bumping the minor version, and also stops squiggle warnings in my IDE.

charypar:
Wasn't the temporary fork also to avoid a `u128` somewhere inside http types to support emscripten? Or am I misremembering? 

Either way, we need to switch to `http` as soon as we can.

Issue #474, opened 2025-12-25
Counter example fails to compile for web-dioxus


nilswloewen:
The `examples/counter/web-dioxus/Cargo.toml` file is missing the get_random dependency. `dx serve --web` gives this error:

`➜  web-dioxus git:(master) dx serve --web
-------- Start:  9:16:53pm --------

21:16:54 [dev] -----------------------------------------------------------------
                Serving your app: web-dioxus! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
                Learn more at https://dioxuslabs.com/learn/0.7/getting_started
               ---------------------------------------------------------------- 
21:16:56 [cargo] error: The "wasm_js" backend requires the `wasm_js` feature for `getrandom`. For more information see: https://docs.rs/getrandom/0.3.4/#webassembly-support
  --> /home/nils/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.3.4/src/backends.rs:40:17
   |
40 | /                 compile_error!(concat!(
41 | |                     "The \"wasm_js\" backend requires the `wasm_js` feature \
42 | |                     for `getrandom`. For more information see: \
43 | |                     https://docs.rs/getrandom/", env!("CARGO_PKG_VERSION"), "/#webassembly-support"
44 | |                 ));
   | |__________________^

21:16:56 [cargo] error[E0425]: cannot find function `fill_inner` in module `backends`
  --> /home/nils/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.3.4/src/lib.rs:99:19
   |
99 |         backends::fill_inner(dest)?;
   |                   ^^^^^^^^^^ not found in `backends`

21:16:56 [cargo] error[E0425]: cannot find function `inner_u32` in module `backends`
   --> /home/nils/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.3.4/src/lib.rs:123:15
    |
123 |     backends::inner_u32()
    |               ^^^^^^^^^ not found in `backends`
    |
help: consider importing this function
    |
 33 + use crate::util::inner_u32;
    |
help: if you import `inner_u32`, refer to it directly
    |
123 -     backends::inner_u32()
123 +     inner_u32()
    |

[...]

PR #475, opened 2025-12-25
Issue #474 - Add get_random dep for web-dioxus and update README.md

nilswloewen:
Solves issue https://github.com/redbadger/crux/issues/474

Adds get_random dep for web-dioxus.

Also, README.md has been updated with latest dx version and proper `dx serve` command. 

I noticed that other crates in `counter` example use `get_random`. Perhaps a better fix would be to add it to `examples/counter/Cargo.toml` as a workspace dependency?


charypar (approved):
Thank you!

mhedgpeth:
Thanks again for the contribution! I see that the build failed for unrelated reasons, so I am rerunning it. I'll also update the branch to match what's on main.

PR #479, opened 2026-01-18
refactor: Simplify `CatFacts::view`

mgeisler:
There doesn’t seem to be any reason to use the fully qualified syntax here: the `App` trait is already in scope, so we can call `view()` directly.


charypar (approved):
Nice catch, thank you!

PR #483, opened 2026-01-21
refactor: remove unnecessary `Arc` around `Sender`

mgeisler:
The send handle on a Crossbeam channel is already cheaply clonable. So wrapping it in an `Arc` is an anti-pattern.

Normally, I would `clone()` the send handle and pass it around as an owned value. But here, the functions are small and synchronous, so I just passed a reference.


charypar (approved):
Makes perfect sense 👍🏻

PR #488, opened 2026-02-02
feat: Support passing App and Model object to construct the Core

### LINKED ISSUE #487: Crux doesn't allow custom App/Model constructor
Crux doesn't currently allow the use of a custom App or Model constructor relying on the usage of Default. 

In cases where the Model needs to have some init data from the Core::new() call site it is currently impossible to pass it without resorting to global variables or other workarounds. 

Having a new `new_with(App, App::Model)` and not require `Default` bonds if the `Core::new` isn't used would solve that.

ccgauche:
Previously it was impossible to pass a custom constructor for app or model. Meaning you can't pass init params to the Model a constructor. 

This MR keeps back-compatibility by still supporting `new` if `App` and `App::Model` have `Default` implemented. But don't require those bonds to be present on `App` and `App::Model`. In which, case `map_with` can be called with an already constructed `App` and `App::Model`. 

Closes #487

### DIFF
--- a/crux_core/src/command/tests/integration.rs
+++ b/crux_core/src/command/tests/integration.rs
@@ -4,9 +4,9 @@ use serde::{Deserialize, Serialize};
 use crate::{Command, Request};
 
 // The future version of the app trait
-pub trait App: Default {
+pub trait App {
     type Event: Send + 'static;
-    type Model: Default;
+    type Model;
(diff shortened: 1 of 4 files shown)


StuartHarris (approved):
Thanks for this! I tested it with the `simple_counter` example and all works fine.

PR #491, opened 2026-02-05
feat: NativeBridge for typed FFI without serialization

fraschm1998:
## Summary

This PR adds `NativeBridge` - a new bridge that delivers effects as typed values via callbacks instead of serializing to bytes. This provides a zero-copy path using UniFFI annotations for typed foreign function interfaces.

## Binding Generators

The UniFFI annotations work with multiple binding generators:

- **uniffi-bindgen** (standard): Generates Swift, Kotlin, Python, Ruby bindings for native iOS/Android apps
- [**uniffi-bindgen-react-native**](https://github.com/jhugman/uniffi-bindgen-react-native): Generates TypeScript bindings for React Native (iOS/Android via JSI) and Web (WASM)

### crux_core
- Add `NativeBridge` struct with typed effect callback
- Add `NativeResolveRegistry` for tracking native effect resolutions
- Add `EffectNative` trait and `ResolveNative` type
- Add `UnitOutput` wrapper type for `()` (UniFFI doesn't support unit types in enum variants)
- Update `#[effect]` macro to support `native_bridge` attribute
- Add `native_bridge` feature flag

### crux_macros
- Added `uniffi::Enum` derive to `EffectOutput` when using `facet_typegen`
- Added `uniffi::Record` derive to `NativeRequest` when using `facet_typegen`

### Capabilities with native_bridge support
- **crux_http**: HTTP capability with UniFFI annotations
- **crux_kv**: Key-value storage capability with UniFFI annotations
- **crux_time**: Time capability with UniFFI annotations
- **crux_platform**: Platform capability with UniFFI annotations

### counter-next-native example
New example demonstrating NativeBridge with typed FFI:
- Uses `NativeBridge` for typed effect delivery (no byte serialization)
- Shell receives typed `EffectFfi` enum instead of `Vec<u8>`
- Resolution uses typed `EffectOutput` instead of bytes
- View returns typed `ViewModel` directly
- Includes iOS and Android shell scaffolding

## NativeBridge vs Bridge

| Aspect | Bridge | NativeBridge |
|--------|--------|--------------|
| FFI approach | Byte serialization (Bincode) | Typed FFI (UniFFI) |
| Effect delivery | `Vec<u8>` | Typed `EffectFfi` enum |
| Resolution | `resolve(id, &[u8])` | `resolve(id, EffectOutput)` |
| View | `view() -> Vec<u8>` | `view() -> ViewModel` |
| Shell callback | `\|bytes: Vec<u8>\|` | `\|id, effect: EffectFfi\|` |

## Test plan

- [x] `cargo test -p shared` in counter-next-native (6 tests pass)
- [x] `cargo test --workspace` (all tests pass)
- [x] `cargo clippy` (no warnings)


charypar:
Sorry it took a month, but I finally drafted an RFC which aims to support the use this bridge was targetting, but in hopefully a simpler way. Would love feedback on https://github.com/redbadger/crux/pull/514

It does unfortunately mean this work would get rejected, but it should be a lot easier to achieve the native/opaque FFI with that.

mhedgpeth:
It looks like this is going through in #514 , so closing this. Thanks so much @fraschm1998 for your contribution! It helped move this forward, even if we're not going to use this exact code.

PR #493, opened 2026-02-13
fix: prevent stack overflow when middleware resolves effects synchronously

### LINKED ISSUE #492: Stack overflow when working with middlewares
Hi,

```rust
#[derive(Default)]
pub struct App;

impl crux_core::App for App {
    type Effect = Effect;
    type Event = Event;
    type Model = Model;
    type ViewModel = ViewModel;

        Command::new(async |ctx| {
            for _ in 0..4 {
            }
        })
    }

        ViewModel
    }
}

/* ... */

pub struct PingMiddleware;

where
{
    type Op = PingOperation;

    fn try_process_effect_with(
        &self,
        effect: Effect,
    ) -> Result<(), Effect> {

        println!("resolve() begin");

        println!("resolve() end");

        Ok(())
    }
}
```
... prints:
```
resolve() begin
resolve() begin
resolve() begin
resolve() begin
resolve() end
resolve() end
resolve() end
resolve() end
```
```rust
for _ in 0..100_000 {
}
```
```
abort trap: 6
```

Reproduction:

fraschm1998:
## Summary

Fixes #492

When an `EffectMiddleware` resolves effects synchronously inside `process_effect`, the resolve chain recurses unboundedly, causing a stack overflow with enough sequential requests.

This PR prevents the issue at the API level by introducing an `EffectResolver<Output>` type with a runtime guard. If `resolve()` is called while `process_effect` is still on the stack, it panics with a clear message. This follows the same pattern as `RefCell` — catch misuse immediately rather than allowing silent stack overflow.

### Changes

**New `EffectResolver<Output>` type** — wraps the resolve callback + request handle behind an opaque struct with an `AtomicBool` guard. Set to `true` before calling `process_effect`, cleared to `false` after it returns. `resolve()` asserts the flag is `false`.

**Simplified `EffectMiddleware` trait** — no longer generic over `Effect`. The framework handles `TryInto<Request<Op>>` and `Request::split()` internally. Middleware implementations just receive `(operation, resolver)`:

```rust
pub trait EffectMiddleware: Send + Sync {
    type Op: Operation;

    fn process_effect(
        &self,
        operation: Self::Op,
        resolver: EffectResolver<<Self::Op as Operation>::Output>,
    );
}
```

**Removed trampoline** — the deferred queue, drain loop, and race-condition handling from the previous iteration are all gone.

### Why runtime, not compile-time?

Rust's type system can't express "must be called from a different thread" or "must be called after this function returns." Temporal ordering across threads isn't something the borrow checker can model. A session-type state machine would add significant API complexity for something already well-documented as a requirement.

### Breaking change

This is a breaking change for existing `EffectMiddleware` implementations. The trait signature changed — implementations now receive `(operation, resolver)` instead of the full effect + raw callback.

## Test plan

- [x] `#[should_panic]` test verifies synchronous resolution panics with expected message
- [x] All existing middleware tests pass (one dice, three dice, HTTP, remote trigger, late trigger, effect narrowing, bridge variants)
- [x] `cargo test -p crux_core` — 60 passed, 0 failed, 11 ignored (all pre-existing)
- [x] `cargo clippy -p crux_core -- -D warnings` clean
- [x] `counter-next` example compiles with updated middleware

### DIFF
--- a/crux_core/src/middleware/effect_handling.rs
+++ b/crux_core/src/middleware/effect_handling.rs
@@ -1,6 +1,9 @@
-use std::sync::{
-    Arc, Weak,
-    atomic::{AtomicBool, Ordering},
+use std::{
+    sync::{
+        Arc, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
+    thread::{self, ThreadId},
 };
 
 use crate::{Request, RequestHandle, Resolvable, ResolveError, capability::Operation};
@@ -25,6 +28,8 @@ pub struct EffectResolver<Output: Send + 'static> {
     resolve_fn: ResolveFn<Output>,
     /// `true` while `try_process_effect` is executing on the call stack.
     active: Arc<AtomicBool>,
+    /// The thread that called `try_process_effect`.
+    calling_thread: ThreadId,
 }
 
 impl<Output: Send + 'static> EffectResolver<Output> {
@@ -43,7 +48,7 @@ impl<Output: Send + 'static> EffectResolver<Output> {
     /// See <https://github.com/redbadger/crux/issues/492>
     pub fn resolve(&mut self, output: Output) {
         assert!(
-            !self.active.load(Ordering::Acquire),
+            !(self.active.load(Ordering::Acquire) && thread::current().id() == self.calling_thread),
             "EffectMiddleware::try_process_effect must not call resolve() synchronously. \
              Dispatch work asynchronously (thread, spawn_local, channel, etc.). \
              See https://github.com/redbadger/crux/issues/492"
@@ -326,6 +331,7 @@ where
                     handle,
                     resolve_fn: Box::new(resolve_fn),
                     active: active.clone(),
+                    calling_thread: thread::current().id(),
                 };
 
                 // Call the middleware. resolve() will panic if called during


charypar:
On closer consideration, I'm afraid the answer is more likely to be neither 😓 See https://github.com/redbadger/crux/issues/492#issuecomment-3916967113 for the reasoning

crux_core/src/middleware/effect_handling.rs
```diff
@@ -1,82 +1,149 @@
...
 
 impl<Next, EM> Layer for HandleEffectLayer<Next, EM>
 where
-    // Next layer down, core being at the bottom
     Next: Layer,
-    // Effect has to try_into the operation which the middleware handles
-    Next::Effect: TryInto<Request<EM::Op>, Error = Next::Effect>,
-    // The actual middleware effect handling implementation
-    EM: EffectMiddleware<Next::Effect> + Send + Sync + 'static,
+    Next::Effect: TryInto<Request<EM::Op>, Error = Next::Effect> + Send,
```
charypar: The `Send` bound doesn't appear necessary - I could remove it from here and the `impl<Next, EM> HandleEffectLayer<Next, EM>` without issues.

crux_core/src/middleware/mod.rs
```diff
@@ -108,8 +108,8 @@ pub trait Layer: Send + Sync + Sized {
     /// must implement the [`EffectMiddleware`] trait.
     fn handle_effects_using<EM>(self, middleware: EM) -> HandleEffectLayer<Self, EM>
     where
-        EM: EffectMiddleware<Self::Effect> + Send + Sync + 'static,
-        Self::Effect: TryInto<Request<EM::Op>, Error = Self::Effect>,
+        EM: EffectMiddleware + 'static,
+        Self::Effect: TryInto<Request<EM::Op>, Error = Self::Effect> + Send,
```
charypar: Similar here, `Send` doesn't look to be required.

crux_core/src/middleware/effect_handling.rs
```diff
@@ -1,82 +1,149 @@
...
-    /// The expected error type is the same as the input Effect type, allowing the conversion to be attempted
-    /// non-destructively.
-    fn try_process_effect_with(
+    /// The framework has already extracted the operation from the effect enum.
+    /// Use the [`EffectResolver`] to send the result back. The resolver **must
+    /// not** be called before this method returns — dispatch the work
+    /// asynchronously (e.g. `std::thread::spawn` on native, `spawn_local` on
+    /// WASM, or a channel send) and call [`EffectResolver::resolve`] from
+    /// there.
+    fn process_effect(
```
charypar: Could we keep the naming (or at least the `try_` part)? It helps hint at the intended operation for the method to do - process effect if it can, return it in `Err` otherwise. It's not a super intuitive thing, I was hoping to help it with the name 😅

charypar (approved):
Thank so much, this not only stops the problem, but even makes the API a lot nicer!

Bar the small comments and rust fmt config, lets get this merged!

crux_core/src/middleware/effect_handling.rs
```diff
@@ -1,82 +1,154 @@
...
+    ///
+    /// Panics if called synchronously from within
+    /// [`EffectMiddleware::try_process_effect`]. Middleware must dispatch work
+    /// asynchronously (e.g. `std::thread::spawn`, `spawn_local`, or a channel)
+    /// and call `resolve` from there.
+    ///
+    /// See <https://github.com/redbadger/crux/issues/492>
+    pub fn resolve(&mut self, output: Output) {
+        assert!(
+            !(self.active.load(Ordering::Acquire) && thread::current().id() == self.calling_thread),
```
Patryk27, line 51: How does it work on WASM?

I think all `std::thread` functions are just stubs there - and even if `std::thread::current().id()` happens to work, I'd expect it to always return the same id.

charypar: Oh, the thread part well might!

fraschm1998: Both conditions must hold: active == true AND same thread. On WASM, spawn_local doesn't run the closure immediately, it schedules it for later. By the time resolve() actually executes, process_effect has already returned and active is false, so the thread ID is irrelevant.

Added a test at https://github.com/redbadger/crux/pull/497

PR #494, opened 2026-02-14
fix: Do not recurse again after a middleware roundtrip.

### LINKED ISSUE #492: Stack overflow when working with middlewares

o0Ignition0o:
Fixes #492

This changeset adds a worker and a queue to ensure middleware and core back and forth calls via callback won't eventually cause a stack overflow.

This probably needs to be iterated on as I've:
- Added a `Send` bound to `Effect`
- cloned a lot of things
- agressively used SeqCst ordering which might not be that needed.

There is probably room for improvement, happy to iterate if you have any ideas

Edit: The `Send` bound is probably needed as I'm sending effects across the queue. I would assume it not being completely breaking since I cannot find a usecase for `Effects` not being `Send`, they're meant to make it to middlewares, and / or to the shell eventually 🤔.

Please correct me if I'm wrong here!

### DIFF
--- a/crux_core/src/middleware/effect_handling.rs
+++ b/crux_core/src/middleware/effect_handling.rs
@@ -1,4 +1,6 @@
-use std::sync::{Arc, Weak};
+use std::sync::{Arc, Weak, atomic::AtomicBool};
+
+use crossbeam_channel::{Receiver, Sender, unbounded};
 
 use crate::{Request, RequestHandle, Resolvable, ResolveError, capability::Operation};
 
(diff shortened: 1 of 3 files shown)


o0Ignition0o:
Closing this one in favor of https://github.com/redbadger/crux/pull/496, I hope we can work out a compile time error, although I'm afraid there might not be any solution for that.

PR #495, opened 2026-02-14
Tui examples

StuartHarris:
Adds TUI shells (using [ratatui)](https://ratatui.rs/) to `simple_counter` and `counter`. The `counter` TUI has a debug panel to show events and effects.

### DIFF
--- a/examples/counter/tui/src/core.rs
+++ b/examples/counter/tui/src/core.rs
@@ -1,7 +1,7 @@
-use anyhow::{Result, anyhow};
-use crossbeam_channel::Sender;
+use anyhow::Result;
 use futures::TryStreamExt;
 use std::collections::VecDeque;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use tokio::spawn;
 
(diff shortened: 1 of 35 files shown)


examples/counter/tui/src/main.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use crate::core::EventLog;
+use crate::counter_widget::CounterWidget;
+use crate::debug_panel::DebugPanel;
+
+/// owns all state and drives the event loop
+struct App {
+    core: core::Core,
+    effect_tx: Sender<Effect>,
+    effect_rx: Receiver<Effect>,
```
charypar: This is a mildly misleading naming - should probably be `render_tx/rx` because that's the only use.

This channel is drained further down, without any level of skipping render, so I'm not sure whether it in fact does anything - draw runs in a loop anyway...?

charypar: And the view model is read in the main widget `render` call. 

I suppose we could set a flag and run that call conditionally, so we don't waste `view` calls?

examples/counter/tui/src/core.rs
```diff
@@ -0,0 +1,94 @@
...
+pub type Core = Arc<shared::Core<Counter>>;
+pub type EventLog = Arc<Mutex<VecDeque<String>>>;
+
+pub fn new() -> Core {
+    Arc::new(shared::Core::new())
+}
+
+pub fn new_log() -> EventLog {
+    Arc::new(Mutex::new(VecDeque::new()))
+}
```
charypar, line 31: I suspect there's a nicer design without sharing the core across taks (something something mpsc channel for effect responses), but it's alright anyway.

examples/counter/tui/src/main.rs
StuartHarris: I replaced the channel with a render flag, and cached the view when the flag is set. Overall a simpler pattern

PR #496, opened 2026-02-18
fix: Make sure middlewares cannot run synchronously.

### LINKED ISSUE #492: Stack overflow when working with middlewares

o0Ignition0o:
Fixes #492

This change set prevents middlewares from returning synchronously, and hints that implementers must either:
- run the middleware logic in a separate task to prevent stalling the crux runtime
- run the sync logic in the capability definition, and not in the middleware

### DIFF
--- a/crux_core/src/middleware/effect_handling.rs
+++ b/crux_core/src/middleware/effect_handling.rs
@@ -1,8 +1,10 @@
-use std::sync::{Arc, Weak};
+use std::sync::{Arc, Weak, atomic::AtomicBool, atomic::Ordering};
 
 use crate::{Request, RequestHandle, Resolvable, ResolveError, capability::Operation};
 
 use super::Layer;
+
+static BLOCKING_MIDDLEWARE_ERROR: &'static str = "Blocking middleware detected. Middlewares must not block the crux runtime. Make sure the middleware runs in a separate task.";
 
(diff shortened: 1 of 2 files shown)


charypar (approved):
Excellent! Thanks so much for the change!

charypar:
Finally had a chance to look at #493 properly, I think I agree @o0Ignition0o, it is better. Closing this one in favour of it, but thank you regardless!

PR #498, opened 2026-02-23
Expose `AbortHandle`

Patryk27:
`Command::abort_handle()` returns `AbortHandle` that's currently private - this commit exposes that type, so you can store `AbortHandle` on the application's side.


charypar (approved):
Good catch! Thank you

PR #501, opened 2026-03-07
Update to facet_generate v0.16

StuartHarris:


mhedgpeth:
@StuartHarris the C# examples being proper are waiting for this, but I think you weren't aligned with @charypar in the last community meeting about not yet being aligned on the API for the new facet version namespace yet. Or are we settled on that?

If we're aligned I'd like to:

1. Merge examples
2. Update this to update
3. Add C# to examples

In rapid succession

StuartHarris:
Hey @mhedgpeth this is a good idea. Viktor suggested `gen` as a prefix rather than `fg`, which I also prefer. I need to understand whether that's a breaking change or not (as in can you can already use `facet_generate` as a prefix so can you use any alias? Not sure. Will find out today and see if I can move this forward.)

mhedgpeth:
That seems simple enough, do you consider this a breaking change that would warrant `0.18`? If so we can get the examples merged, then bump the version and even create an rc1 of that perhaps?

StuartHarris:
So I did a bit of experimentation and it doesn't matter what prefix is used. We can't use `gen` though as that's a reserved keyword in Rust (for marking generator functions that yield). So I updated this PR to use `#[facet(typegen::namespace = "...")]` instead and that works. I think we could use the `typegen` prefix in our code, examples and docs, for now, and I'll update `facet_generate` and `facet-generate-attrs` so that it's there from the next release, even though we're not dependent on that.

mhedgpeth (approved):
Looks great Stuart

Issue #508, opened 2026-03-21
Proposal: Add Renovatebot for automated dependency management


mhedgpeth:
## Problem

Each example supports multiple shells (Leptos, Next.js, Yew, Dioxus, Tauri, Android/Kotlin, SwiftUI) and each shell brings its own dependency tree across Cargo, npm/pnpm, and Gradle. With ~30 Cargo.toml files, ~7 package.json files, and Gradle build files, the explosion of unrelated dependencies makes it nearly impossible to keep things current manually.

Today, dependency updates are handled by fish scripts in `scripts/` that cover Rust (`cargo update` / `cargo upgrade`) and pnpm across the root workspace and examples. However:

1. `cargo upgrade --incompatible allow` blindly upgrades everything, ignoring the many exact-pinned versions in the project. There's no way to express "upgrade everything except X because of Y."
2. Gradle (Android), GitHub Actions, and Swift dependencies aren't covered by any script.
3. We have many exact-pinned dependencies (see below) but the reasons aren't captured anywhere. When it's time to unpin, the context is lost.
4. Someone has to remember to run the scripts. There's no notification when a dependency gets a security fix.

## Proposal

Add [Renovatebot](https://docs.renovatebot.com/) to automate dependency updates with a conservative, low-noise configuration:

- Once a week — scheduled for a single window (Wednesday morning UTC) to avoid constant PR churn
- Grouped PRs — dependencies grouped by area (core crates, each example, GitHub Actions) so we get a handful of PRs, not dozens
- Concurrent PR limit — cap at 3 open PRs at a time
- Minimum release age — 7 days, so new releases have a full week for the community to surface issues before Renovate proposes them
- Declarative exception documentation — every pinned or excluded dependency gets a `description` field explaining why and what needs to happen before it can be unpinned

### What Renovate would manage

| Ecosystem | Count | Location |
|---|---|---|
| Cargo (Rust) | ~30 Cargo.toml files | Root workspace + 5 example workspaces |
| npm/pnpm | ~7 package.json files | Web frontends (Next.js, React Router, Tauri) |
| Gradle | Android build files | Example Android shells |
| GitHub Actions | 3 workflow files | `.github/workflows/` |

### Expected PR volume

With the grouping rules, the first run would produce roughly 6-10 PRs (one per group with updates available), but only 3 will be open at a time. A maintainer merges or closes one before the next opens. Ongoing, expect a few PRs per week.

## Documenting pinned dependency exceptions

[...]

PR #509, opened 2026-04-01
chore: introduce renovatebot config

mhedgpeth:
This addresses #508 

When this is merged to `master` the rest of the renovatebot process will kick in where there is an issue that drives everything.

Before we merge, we should align on the policy:

1. Automatically merge patch-level (last-digit) dependencies and all dev-dependencies
2. Pinned versions with what I think are the reasons they're pinned.

Then, once we merge and have the rest in place, next Wednesday we'll get some PRs to update the rest and will sort those out. Only a max of 3 PRs at a time are allowed.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,114 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":dependencyDashboard",
+    ":semanticCommitTypeAll(chore)"
+  ],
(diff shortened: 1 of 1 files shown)


mhedgpeth:
@StuartHarris this is ready for merging, it should help with the dependency management of the repo

StuartHarris (approved):
Amazing. Thanks @mhedgpeth and apologies it's taken me so long to get to this. I'll update the pinned versions when we merge the other PRs.

Issue #510, opened 2026-04-03
Incorporate API Secret into Weather App


mhedgpeth:
## Problem

The weather app requires an `OPENWEATHER_API_KEY` key in order to run properly. This complicates the setup and running of it.

The user should be the one to provide that API key, not the developer. This simplifies the example to be "just run it" and then also shows off a capability to read secrets.

## Proposal

1. Introduce state machine at the model level with `Uninitialized`, `Initializing`, `Configuration` and `Active` variants, where the current model is inside the `Active` variant
2. When shells start they send `Event::Start` to the model, initializing the model to fetch secrets
3. Introduce a new `secret` effect/capability to fetch secrets (based on Michael's internal project where this already works fine)
4. When starting, the model fetches secrets in `Initializing`, if the secret is not fetched, the model transitions to `Configuration` where they configure the secret, otherwise it transitions to `Active` with that secret, where everything it does today happens, with an API
5. The additional elements of the build system can be simplified: `justfile`, the `github actions` and `github secret` can be simplified to exclude the `.env` and reference to `OPENWEATHER_API_KEY`
6. The book would need to update to explain the state machine, but this is a good thing

mhedgpeth:
This is completed :)

PR #512, opened 2026-04-07
Incorporate API Secret into Weather example

mhedgpeth:
This addresses #510 

When I refactored the examples I encountered two issues with the weather examples:

1. It relies on a developer-provided API key from Open Weather which really could be user-provided
2. When it fails to either get that API key or the API key fails, it just freezes at "waiting for connection" leaving the user (and example runner) very confused.

This PR is a full walkthrough of the example that starts with the premise of solving the above problems, but also as I solve the problems, adopts the patterns I use in v2 of my app, People Work, namely:

1. Use of an `Outcome` object on an `update` return, where `update` takes `self` - I find this much more rust-idiomatic and easy to reason about when managing a state machine with state transitions, which crux ends up being in apps like this.
2. Incorporating a hierarchy of events that attach to the mutation of composed objects, which makes the above state transitions easier to reason about at lower levels
3. Using smaller modules, keeping files smaller, to make it clear the exact roles of everything, which makes the whole thing easier to reason about as it grows

The following elements are outstanding before this can become a mergeable PR:

- [x] Add sufficient documentation, I did this as I went but I need to go through one more pass
- [x] Update the book to explain it properly, as a real example

Overall this adds some complexity to the weather example, but for me it gives a user a more realistic view of what a production application might look like and so I think it's worth the change/update.


StuartHarris (approved):
I love this, Thank you for all your work on it. I rebased onto the new master, and _think_ the conflict resolution is correct, but you might want to sense check that everything is included correctly. I ran all the shells and everything appears to work (and it looks great!). ❤️

mhedgpeth:
@StuartHarris thanks for your feedback on this. I updated the book, examples, and aligned this more with my internal learnings since I made this a couple of weeks ago, for example on how Leptos handles signals. This should be ready to merge, and the book ready to update.

StuartHarris:
@mhedgpeth I read through the book. Thanks for being so thorough with this!

PR #513, opened 2026-04-07
fix: bundle `typegen_extensions/` in library

mgeisler:
This fixes a problem we faced when migrating to Bazel: the type generation code relied on being able to read files from where the code was originally compiled. This does not work when compilation happens in a sandbox, possibly one which is ephemeral and which doesn’t exist on the machine running the `codegen` binary.

Instead, the templates are simply included as binary data in the library, and thus also in any downstream `codegen` binaries.

Note that this removes the possibility of customizing type generation by adding a `typegen_extensions/` directory in `$PWD` when executing the `codegen` binary. This thus breaks backwards compatibility. The customization possibilities were documented for Serde, but not for Facet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -769,6 +769,7 @@ dependencies = [
  "facet 0.44.1",
  "facet_generate",
  "futures",
+ "include_dir 0.7.4",
  "log",
  "rand 0.10.0",
  "serde",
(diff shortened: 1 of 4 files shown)


StuartHarris (approved):
Thanks for this, and apologies for being slow to review.

mgeisler:
> Thanks for this, and apologies for being slow to review.

No problem at all! It turns out that it's easy to carry local patches in a Bazel workspace – they are applied automatically before the code is built. This makes it really convenient to test out this kind of fixes.

PR #514, opened 2026-04-20
Effect Routing

charypar:
This is an RFC for a replacement for effect middleware, and a proof of concept test demonstrating how it would work. Not every detail is covered, but hopefully enough for feedback.

The more feedback the better!


charypar:
@StuartHarris this hopefully matches our conversation a while back. It's turned out reasonably nice, other than having to deal with the annoying reference cycle between the core/router and the core-side effect handlers.

I've managed to keep the serialized FFI "lane" broadly the same, but an argument could be made that we should have one typed resolver per effect even in the serialized case and fully avoid the possibility of resolving effect A with a value for effect B.

StuartHarris (approved):
Let's do it!! ;-)

crux_core/tests/effect_router_prototype.rs
```diff
@@ -0,0 +1,708 @@
...
+
+        /// Resolve an effect sent over the serialized lane.
+        pub(crate) fn resolve_serialized(&self, id: u32, data: &[u8]) {
+            self.serialized_registry
+                .resume(EffectId(id), data)
+                .expect("serialized resolve should work");
+
+            // FIXME: adjust the APIs such that it's not possible to forget this call
+            self.router.process();
+        }
```
charypar, line 380: I'll take ideas for how we might change the API such that the `.process()` call isn't independent.

One idea I can think of is that the registries require a core-like (core or router) in order to resume an effect, purely so that they can call the `process`. Since `Router` is from crux_core, it can call the `pub(crate) fn process` method on Core. 

This is a difficulty of the remote unblocking of the effect futures, something then needs to re-enter the core to progress anything waiting...

crux_core/src/bridge/mod.rs
```diff
@@ -10,7 +10,9 @@ use thiserror::Error;
 use crate::{App, Core, core::ResolveError};
 pub use formats::{BincodeFfiFormat, JsonFfiFormat};
 pub use registry::EffectId;
-pub(crate) use registry::ResolveRegistry;
+// FIXME(effect-router-rfc): revisit this public re-export once router/serialized-lane
+// integration settles and we can provide a smaller dedicated API.
+pub use registry::ResolveRegistry;
```
charypar: The registries feel like they now need a tidy up... We're effectively in a situation where:

- Bridge is public, hides a serialized ResolveRegistry inside of it
- The router API exposes both this serialized Registry and a pure storage one, as helpers for the shell-side effect handlers

The supported FFI models are basically:
1. Core -> Bridge -> Serialized FFI
2. Core -> Router -> Custom effect handling, including serialized FFI and not serialized FFI, each with a resolve registry.

It all feels a bit ad-hoc and in need of a good tidy 🤔

charypar:
@StuartHarris I had a go at solving both of those problems by reifying the idea of lanes in code (I call them routes now). This both hides both the registry implementations (bridge one and new one), so we can change and optimise them how we like, _and_ makes it impossible to resolve effects without running the effect pump. 

Each route looks a bit like a core in its API, but it can be anything we find useful. The user is responsible for defining the route set, but we create it and give it the router weak ref. The dispatch function then gets the complete, ready to use route set, and the FFI can access it through the router.

mhedgpeth:
@charypar it sounds like this is starting to be the implementation; if so could you change the title? Is it ready? If not could you make this a draft?

charypar:
@mhedgpeth I'd say it's ready for review and feedback now – it's gone from just a description to an actual proposed implementation. 

We can either do the feedback on this PR, or I can merge the core changes under a feature flag and we can collect feedback some other way? 

I do feel we perhaps need a better process of proposing and validating these larger changes 🤔 any ideas?

crux_core/src/effects/registry.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn resolve(&self, id: u32, output: Op::Output) -> Result<(), ResolveError> {
+        let mut requests = self.requests.lock().expect("registry lock poisoned");
+        let Some(mut request) = requests.remove(&id) else {
+            panic!("missing pending handle for id {id}");
+        };
+        drop(requests);
+
+        request.resolve(output)?;
+
+        if !matches!(request.handle, RequestHandle::Never) {
```
StuartHarris: should this be `if matches(request.handle, RequestHandle::Many)`?

examples/counter-routing/shared/src/ffi.rs
```diff
@@ -0,0 +1,281 @@
...
+                .routes
+                .serialized
+                .view()
+                .expect("view model should serialize")
+        }
+    }
+}
+
+#[cfg(feature = "wasm_bindgen")]
+pub mod wasm_bindgen {
```
StuartHarris, line 123: we should also move `wasm_bindgen` and `wit_bindgen` from middleware to routes

crux_core/src/effects/registry.rs
```diff
@@ -0,0 +1,80 @@
...
+
+pub struct Registry<Op: Operation> {
+    next_id: AtomicU32,
+    requests: Mutex<HashMap<u32, Request<Op>>>,
+}
+
+impl<Op: Operation> Default for Registry<Op> {
+    fn default() -> Self {
+        Self {
+            next_id: AtomicU32::new(0),
```
StuartHarris: slab?

StuartHarris (changes requested):
loving this. Especially when we codegen some of it. ❤️

examples/counter-routing/shared/src/ffi.rs
charypar: Yes, yes we should.

crux_core/src/effects/registry.rs
charypar: Yea, I only cared about the API of this, it's just what fell out of GPT, sorry 😅

mhedgpeth:
> @mhedgpeth I'd say it's ready for review and feedback now – it's gone from just a description to an actual proposed implementation.
> 
> We can either do the feedback on this PR, or I can merge the core changes under a feature flag and we can collect feedback some other way?
> 
> I do feel we perhaps need a better process of proposing and validating these larger changes 🤔 any ideas?

I think this is great, if it's ready let's keep it, I'll rename it and create an issue that describes it. Implementations are better than markdown files :)

Thanks!

I do think we should get 0.18 out, then think about this as a part of 0.19

mhedgpeth:
Created #525 to track this

crux_core/src/effects/registry.rs
charypar: Yes, definitely that.

crux_core/src/effects/registry.rs
charypar: Made a pass at a more production-grade version

charypar:
@StuartHarris I've folded your feedback in. I also updated the registry to have typed effect IDs, which are now also typed by output – should be impossible to resolve with the wrong type – and they use generational IDs so that reusing slots in the Slab can't accidentally lead to resolving a newer effect with a stale effect ID.

I've also expanded the test a little to make the core-side effect a streaming one, just to make sure nothing prevents that.

StuartHarris (approved):
Love it, thanks @charypar ❤️ 
If you're happy with #541 then we can merge that into this, and this into master?

StuartHarris:
@charypar I added doc comments. Have a look at https://github.com/redbadger/crux/pull/514/commits/f71d8b9d20a3d2b1c4cb29dd60ac846c297e35a2 and then feel free to merge! 

(also, the new example doesn't run now and is excluded in CI, see #541)

PR #515, opened 2026-04-20
Resurrect lost RFCs

charypar:
Somehow in the process of new docs, the RFCs got lost, this brings them back.


mhedgpeth:
Sorry about the removal, when I went through the docs, I thought to remove them because it was confusing to me whether they were aspirational or implemented, so since they were implemented, I incorporated them in the docs and removed the RFCs. I think if we add the RFCs when they're aspirational, it's perfect. It just was confusing after that aspiration is implemented. I hope that makes sense, and if you want to re-incorporate these, not a problem.

charypar:
I see what you mean - maybe we can add a status to them then? like... proposed / implemented

mhedgpeth (approved):

PR #517, opened 2026-04-21
Support C# with counter example

mhedgpeth:
This PR adds support for C# through facet-generate and it shows that support through a windows shell in the counter example.

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -134,7 +134,7 @@ jobs:
       - uses: taiki-e/install-action@just
       - uses: taiki-e/install-action@v2
         with:
-          tool: dioxus-cli
+          tool: dioxus-cli@0.7.5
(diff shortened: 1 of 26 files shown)


mhedgpeth:
@StuartHarris I also need to push to `examples.yaml` a windows runner that runs the examples, but I can't due to not having workflow permissions.

That change is, examples.yaml starting on 224:
[examples.yaml](https://github.com/user-attachments/files/26949702/examples.yaml)

If you can provide me that access or else add this file yourself we'll have this build in place. I'm getting the error: `! [remote rejected]   michael/csharp-support -> michael/csharp-support (refusing to allow an OAuth App to create or update workflow `.github/workflows/examples.yaml` without `workflow` scope)`

StuartHarris (approved):
Amazing! Thanks ❤️

PR #523, opened 2026-05-02
Fluent Command Assertions

mhedgpeth:
The current way to assert effects and resolve them to events is verbose, and that leaves me not testing my app as much as I would like.

```rust
// today, asserting one HTTP request, resolving it, then feeding the event back
let mut request = cmd.expect_one_effect().expect_http();
assert_eq!(request.operation, HttpRequest::get("...").build());
request.resolve(HttpResult::Ok(...)).unwrap();
let event = cmd.expect_one_event();
let cmd = local.update(event, &api_key());
```

This change introduces a fluent way to do this, which would make testing more
accessible while writing it, leading to better tests.

There are three variants of this:

**Quick render assertion**: single effect, no payload check:

```rust
// was:
let mut cmd = app.update(Event::Reset, &mut model);
cmd.expect_one_effect().expect_render();

// now:
app.update(Event::Reset, &mut model).expect_only_render();
```

**Multiple chained assertions**: assert each effect's variant and operation, in order:

```rust
// was:
let secret_request = cmd.expect_effect().expect_secret();
assert_eq!(secret_request.operation, SecretRequest::Fetch(API_KEY_NAME.to_string()));
let kv_request = cmd.expect_one_effect().expect_key_value();
assert!(matches!(kv_request.operation, KeyValueOperation::Get { .. }));

// now:
cmd.expect_secret_with(|op| {
       assert_eq!(op, &SecretRequest::Fetch(API_KEY_NAME.to_string()));
   })
   .expect_key_value_with(|op| {
       assert!(matches!(op, KeyValueOperation::Get { .. }));
   });
```

**Resolve to an event**: assert, resolve, and hand back the event for `update()`:

```rust
// was:
let mut request = cmd.expect_one_effect().expect_http();
request.resolve(HttpResult::Ok(HttpResponse::ok().body(weather_json).build())).unwrap();
let event = cmd.expect_one_event();
let cmd = local.update(event, &api_key());

// now:
let event = cmd
    .resolve_http(|op| {
        assert_eq!(op, &expected_request);
        HttpResult::Ok(HttpResponse::ok().body(weather_json).build())
    })
    .expect_event();
let cmd = local.update(event, &api_key());
```

The closure does double duty: assert on `&op`, then return the response payload the validator hands to `request.resolve(...)` for you. The next chain step (or `.expect_event()`) sees whatever the resolved task emits.

Since the current API still allows for edge-case testing and since I don't want to introduce breaking changes, it's untouched.

The examples are updated with the fluent style, as well as the book. New panic-case tests in `crux_core/tests/effect_test_ext.rs` cover empty queues, wrong variants, missing events, and `expect_only_*` strictness.


mhedgpeth:
@StuartHarris @charypar I'm open to changing the ergonomics of this, I just made a judgement call on what looked best to me here, happy to rename anything if you find it confusing.

This has been something I've been wanting that I've built internally but just makes life easier while writing tests.

crux_macros/src/effect/macro_impl.rs
```diff
@@ -152,6 +152,149 @@ pub fn effect_impl(args: Option<Ident>, input: ItemEnum) -> TokenStream {
...
+            where
+                F: ::core::ops::FnOnce(&Event),
+            {
+                let ev = self.events().next()
+                    .unwrap_or_else(|| panic!("expected an event but got none"));
+                f(&ev);
+                self
+            }
+        }
+    };
```
charypar, line 335: Can we make sure we only compile these in `#[cfg(test)]`?  Should be as easy as annotating the trait definition and implementation as such.

docs/src/part-1/testing.md
```diff
@@ -23,17 +23,19 @@ mod test {
...
 
 We create an instance of the app, and an instance of the model. Then we call update with the `Event::Reset` event.
-As you may remember we get back a `Command`, which we expect to carry a request for a render operation. Using the
-expectation helper API of the Command type, we check we got one effect, and that the effect is a render. Both methods will panic if they don't succeed (they are also `#[cfg(test)]` only, don't use them outside of tests).
+As you may remember we get back a `Command`, which we expect to carry a request for a render operation.
+The `#[effect]` macro on the `Effect` enum we declared earlier generates an `EffectTestExt` extension trait
+that adds chainable test-helper methods to `Command`. One of them is `expect_only_render`, which asserts
+"the next effect is a Render and there are no others." It panics if either condition fails.
+The trait is generated alongside the `Effect` declaration, so `use super::*;` brings it into scope automatically.
+By convention these `expect_*` helpers are intended for test use only.
```
charypar:
```suggestion
As you may remember we get back a `Command`, which we expect to carry a request for a render operation.
The `#[effect]` macro on the `Effect` enum we declared earlier generates chainable test-helper methods for `Command` (technically, this is implemented a a trait called `EffectTestExt`, which needs to be in scope). One of them is `expect_only_render`, which asserts
"the next effect is a Render and there are no others." It panics if either condition fails.
The trait is generated alongside the `Effect` declaration, so `use super::*;` brings it into scope automatically.
```

docs/src/part-2/testing-effects.md
```diff
@@ -53,19 +53,19 @@ cargo nextest run
 
 ## The test steps
 
-Crux provides test APIs to make the tests a bit more readable, but it's still up to the test to drive the event → update → effect → resolve cycle by hand.
+Crux gives us a chainable test API that keeps assertions concise, but you still drive the event → update → effect → resolve cycle yourself. You'll see methods like `.expect_only_render()` and `.resolve_http(|op| ...)` on `Command` throughout this chapter. They come from an `EffectTestExt` trait the `#[effect]` macro generates next to your `Effect` enum.
```
charypar:
```suggestion
Crux gives us a chainable test API that keeps assertions concise, but you still drive the event → update → effect → resolve cycle yourself. You'll see methods like `.expect_only_render()` and `.resolve_http(|op| ...)` on `Command` throughout this chapter. They come from an `EffectTestExt` trait the `#[effect]` macro generates next to your `Effect` enum, and auto-implements for `Command<Effect, E>` where `E` is an event.
```

charypar (approved):
This is _super_ nice! 👏🏻 I left some minor suggestions, but it reads really nicely and makes the tests more readable still. Great! ❤️

StuartHarris:
I love it, thanks @mhedgpeth 🚀

docs/src/part-1/testing.md
StuartHarris: typo: `this is implemented a a trait` (first "a" should be "by"?)

StuartHarris (approved):
Fantastic!

docs/src/part-1/testing.md
charypar: I meant "as" 🤦🏻‍♂️

crux_macros/src/effect/macro_impl.rs
mhedgpeth: @charypar that would make them only available in in the same crate that declared `Effect` and would shut out integration tests or other crates. The current assertions don't have `cfg(test)`, I believe, for this reason.

So I think we leave it as is; does that work for you?

charypar: Is that right? I'd have thought that the test binaries built from `tests/` modules would automatically build their parent crate (which declares the `Effect`) in `cfg(test)` as well 🤔 

I'd really like us to avoid generating test-only helpers and code in people's production binaries, for both compile times and wasm bundle sizes.

StuartHarris: I've gated as many test helpers behind `#[cfg(test)]` as I can. The others I've left. e.g. `Effect::is_*()` and `Effect::into_*()` are more general purpose and the others are used cross-crate currently (although I'll see if we can fix that too)

StuartHarris: Actually I'm going to put these in a stacked PR

mhedgpeth:
@StuartHarris I'm going to make sure the new declaration in `dev-dependencies` is included in the book real quick before we decide to merge.

mhedgpeth:
@StuartHarris I made three changes after merging from your change:

1. The `EffectTestExt` wasn't gated by the `testing` feature, so I added that
2. Updated the book to include these details in the introduction to testing
3. Slightly improved the effect testing language to "expecting A, got B" so they're easier to troubleshoot.

I think these are ready, let me know and we can merge. It would be great to include this in 0.18.

StuartHarris (approved):
This is great. Thank you. It's a shame the tokens are now emitted without formatting, but I don't suppose that matters. The extra compile time to pretty print them is probably not warranted seeing as few if anyone needs to see the expanded content here. I'll merge and look at getting 0.18 out

PR #524, opened 2026-05-03
Make macro-generated test helpers available to integration tests

StuartHarris:
### Motivation

The previous commit on this branch gated the `#[effect]`-generated test helpers behind `#[cfg(test)]`. This works for unit tests defined inside the same crate (in `src/`), but **silently breaks for integration tests** defined in a crate's `tests/` directory.

The reason is a subtlety of how Rust compiles integration tests: a file in `tests/` is compiled as a *separate crate* that depends on the library under test. When the library is compiled for that integration-test binary, `cfg(test)` is **not** set for the library — only for the test crate itself. So any items the macro emitted under `#[cfg(test)]` simply don't exist when an integration test tries to use them, producing confusing "method not found" or "trait not in scope" errors with no obvious fix.

#### `crux_macros` — `#[cfg(test)]` replaced with `#[doc(hidden)]`

Two groups of macro-generated items that were previously gated with `#[cfg(test)]` are now annotated with `#[doc(hidden)]` instead:

- **`Effect::expect_*()`** — the per-variant panicking assertion methods on the enum itself (e.g. `Effect::expect_render(self)`). These are used internally by the `EffectTestExt` impl, and occasionally called directly in tests.
- **`EffectTestExt` trait + `impl`** — the fluent `Command`-level helpers (`cmd.expect_render()`, `cmd.resolve_http(…)`, `cmd.then_event(…)`, `cmd.expect_only_render()`, etc.).

`#[doc(hidden)]` achieves the right balance:

- Items are **always compiled**, so integration tests (and any other downstream test crate) can use them without any feature flag or build-script workaround.
- Items are **excluded from rustdoc** output, so they don't clutter the public API docs.
- Items are **deprioritised by IDEs** in autocomplete, so they don't get in the way of normal development.
- The **compile-time and binary-size cost is negligible** — the trait and its blanket impl introduce no vtables, and the generic methods are only monomorphised when actually called. A production build that never calls these helpers pays nothing for them.

The non-panicking `is_*()` and `into_*()` filter methods on the enum remain fully public and unconditional — they are general-purpose utilities appropriate for production effect-dispatch code.

#### Example integration test added

`examples/counter/shared/tests/example.rs` is a minimal integration test that exercises the `EffectTestExt` helpers from outside the `shared` crate, demonstrating the pattern that was previously broken:

```rust
use crux_core::App as _;
use shared::*;

#[test]
fn increments_count() {
    let app = Counter;
    let mut model = Model::default();

    app.update(Event::Increment, &mut model)
        .expect_only_render();

    let actual_view = app.view(&model).count;
    let expected_view = "Count is: 1";
    assert_eq!(actual_view, expected_view);
}
```

#### `crux_core` — `testing` module is now opt-in

`pub mod testing` (which contains `AppTester`, `Update`, `assert_effect!`, and the `Command::expect_*` family of methods) is now gated with:

```rust
#[cfg(any(test, feature = "testing"))]
pub mod testing;
```

A `testing = []` feature has been added to `crux_core/Cargo.toml`.

This cannot use `#[cfg(test)]` alone because `cfg(test)` does not propagate across crate boundaries — when `crux_core` is compiled as a dependency for another crate's tests, it is compiled as a plain library without `cfg(test)` set. Crates that call `Command::expect_one_effect()`, `AppTester`, `assert_effect!`, etc. directly in their test code must opt in by adding to their `[dev-dependencies]`:

```toml
crux_core = { version = "…", features = ["testing"] }
```

Cargo unifies `[dependencies]` and `[dev-dependencies]` feature sets only during test builds, so the `testing` feature is never active in production compilations of those crates.

The one doc test in `command/mod.rs` that used `expect_one_effect()` / `expect_one_event()` has been updated to use `effects().next().expect(…)` / `events().next().expect(…)` — more instructive in documentation and free of any feature dependency.

#### Crates updated to opt in

| Crate | Reason |
|---|---|
| `crux_core` (self) | Own integration tests in `tests/` compile `crux_core` as a dependency |
| `crux_http` | `tests/command_with_tester.rs` uses `AppTester` and `Update::expect_one_effect()` |
| `crux_kv` | `src/tests.rs` (`#[cfg(test)]`) uses `Command::expect_one_effect()` etc. |
| `crux_time` | `tests/time_test.rs` uses `Command::expect_one_effect()` etc. |
| `examples/counter/shared` | Integration test in `tests/` uses `EffectTestExt` and `crux_core::testing` |
| `examples/counter-http/shared` | `#[cfg(test)] mod tests` uses `Command::expect_one_effect()`, `expect_one_event()` |
| `examples/counter-middleware/shared` | Same |
| `examples/notes/shared` | `#[cfg(test)]` modules use `assert_effect!` |
| `examples/weather/shared` | `#[cfg(test)]` modules use `Command::expect_event()` |

Crates that only use `#[effect]` and test exclusively through the generated `EffectTestExt` fluent API (without importing anything from `crux_core::testing` directly) do **not** need the opt-in.

### Snapshot tests

All 10 generated-output snapshots in `crux_macros/src/effect/tests.rs` have been updated to reflect the `#[doc(hidden)]` annotations in place of `#[cfg(test)]`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -762,6 +762,7 @@ dependencies = [
  "bincode",
  "crossbeam-channel",
  "crux_cli",
+ "crux_core",
  "crux_http",
  "crux_macros",
  "crux_time",
--- a/crux_core/Cargo.toml
+++ b/crux_core/Cargo.toml
@@ -32,6 +32,7 @@ slab = "0.4"
 thiserror.workspace = true
 
 [dev-dependencies]
+crux_core = { path = ".", features = ["testing"] }
 assert_fs = "1.1.3"
 assert_matches = "1.5"
 async-channel = "2.5.0"
@@ -49,4 +50,10 @@ uuid = { version = "1.22.0", features = ["serde", "v4"] }
 default = ["crux_macros"]
 cli = ["dep:crux_cli"]
 facet_typegen = ["crux_macros/facet_typegen", "dep:facet_generate", "dep:log"]
-typegen = ["crux_macros/typegen", "dep:serde-generate", "dep:serde-reflection", "dep:include_dir"]
+testing = []
+typegen = [
+    "crux_macros/typegen",
+    "dep:serde-generate",
+    "dep:serde-reflection",
+    "dep:include_dir",
+]
--- a/crux_core/src/command/mod.rs
+++ b/crux_core/src/command/mod.rs
@@ -204,7 +204,7 @@
 //!     .then_send(Event::GotPost);
 //!
 //! // Check the effect is an HTTP request ...
-//! let effect = cmd.expect_one_effect();
+//! let effect = cmd.effects().next().expect("expected an effect");
 //! let Effect::Http(mut request) = effect else {
 //!     panic!("Expected a HTTP effect")
 //! };
@@ -229,7 +229,7 @@
 //!     .expect("Resolve should succeed");
 //!
 //! // Check the event is a GotPost event with the successful response
-//! let actual = cmd.expect_one_event();
+//! let actual = cmd.events().next().expect("expected an event");
 //! let expected = Event::GotPost(Ok(ResponseBuilder::ok().body(body).build()));
 //! assert_eq!(actual, expected);
 //!
--- a/crux_core/src/lib.rs
+++ b/crux_core/src/lib.rs
@@ -156,6 +156,7 @@ pub mod bridge;
 pub mod capability;
 pub mod command;
 pub mod middleware;
+#[cfg(any(test, feature = "testing"))]
 pub mod testing;
 #[cfg(any(feature = "typegen", feature = "facet_typegen"))]
 pub mod type_generation;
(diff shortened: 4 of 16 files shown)


charypar:
Are we missing the key case, which is integration tests in app crates? We should try this on an example I think 🤔 Ideally it would work without the `testing` feature flag. If that doesn't work, it's better with the ff, than with no conditional compilation, IMHO.

mhedgpeth:
> Are we missing the key case, which is integration tests in app crates? We should try this on an example I think 🤔 Ideally it would work without the `testing` feature flag. If that doesn't work, it's better with the ff, than with no conditional compilation, IMHO.

In my experience an integration test (and even a module) acts as a separate crate so I think the feature should work here. It would be nice to have a test case, thanks @StuartHarris for the update and cleaning it up! 🚀

mhedgpeth (approved):

StuartHarris:
@charypar you were right. I changed cfg(test) to doc(hidden) in the macro generated code (see PR desc). Interestingly there is no code emitted into production binaries as the functions are generic and not monomorphised if not called.

charypar:
> Interestingly there is no code emitted into production binaries as the functions are generic and not monomorphised if not called.

Oh! That's a good point I completely missed.

mhedgpeth (approved):

PR #527, opened 2026-05-07
chore(crux_core): Reduce Command async boilerplate

ManevilleF:
Hi everyone, 

As I dig in the `crux_core` code I noticed some possible improvements in the `command` building logic, following those [recommandations](https://tweedegolf.nl/en/blog/235/debloat-your-async-rust).

The following improvements were made:

- `Command::build` no longer clones the `CommandContext` and does not create an extra `async` block and the associated state machine
- `Command::then_send` utilizes `futures` crate combinator to avoid an extra `async` block
- `Command::notfiy_shell` and `Command::event` do not create an `async` block and returns a sync block with `futures::ready()`
- `Command::into_future` does not create an extra `async` block and returns the original future with no overhead
- `Command::then`, `Command::map_effect` and `Command::map_event` no longer add extra `async` block and utilize `futures` combinator patterns instead
- `StreamBuilder::then_send` and `StreamBuilder::build` no longer:
  -  `pin!` the stream
  -  clone the `CommandContext`
  - add an extra `async` block 
  
 I do not have benchmarks for these specific changes, but I expect small improvements in binary size, especially for WASM on large projects, and a slight reduction in CPU cycles as we get less async state machines.
 
 On a large project, replacing `async` "pass-through" methods where possible reduced the final bundle wasm optimized bundle size by 10kB. As `Command` are one of the foundation of the crux app, I think any optimization will scale
 
 > I did read `CONTRIBUTING.md` but I'm not sure opening an issue is relevant, as no logic is changed and this does not fix any bug

### DIFF
--- a/crux_core/src/command/builder.rs
+++ b/crux_core/src/command/builder.rs
@@ -6,7 +6,7 @@
 //! of the composition are unclear. If you need to compose streams, use the async
 //! API and tools from the `futures` crate.
 
-use std::{future::Future, pin::pin};
+use std::future::Future;
 
 use futures::{FutureExt, Stream, StreamExt};
 
@@ -42,9 +42,7 @@ where
 
     /// Convert the [`NotificationBuilder`] into a [`Command`] to use in an sync context
     pub fn build(self) -> Command<Effect, Event> {
-        Command::new(|ctx| async move {
-            self.into_future(ctx.clone()).await;
-        })
+        Command::new(move |ctx| self.into_future(ctx))
     }
 }
 
@@ -374,9 +372,9 @@ where
         E: FnOnce(T) -> Event + Send + 'static,
         Task: Future<Output = T> + Send + 'static,
     {
-        Command::new(|ctx| async move {
-            let out = self.into_future(ctx.clone()).await;
-            ctx.send_event(event(out));
+        Command::new(move |ctx| {
+            self.into_future(ctx.clone())
+                .map(move |out| ctx.send_event(event(out)))
         })
     }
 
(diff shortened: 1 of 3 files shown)


StuartHarris (approved):
Thanks for this Felix. I like the changes. There are two timing differences (`send_event` and `notify_shell`, which now happen during command construction rather than on first poll). Whilst these are technically different, I don't think they're observable by clients, as the internal channels are always drained through `run_until_settled()`. But @charypar let me know what you think?

ManevilleF:
Oh, that's right I didn't think of this change.
But now I think this might actually be a fix !

Correct me if I'm wrong, but AFAIK `send_event` will push it through a `mpsc` so that the event can be processed by the sending `Command` in `process()`.
I have been investigating some cases in my crux app where it seemed like `send_event` would do nothing until an `Event` was sent by the shell or an `EffectResponse` was sent by a shell or middleware. 
If we don't wait for the first `poll` wouldn't that make the processing more reactive ?

StuartHarris:
Huh, interesting. Yeah that does sound to me like it would be better now than it was before. @charypar will no doubt have an opinion ;-)

ManevilleF:
Maybe not actually, I usually call `CommandContext::send_event` directly, not through a `Command`.

Thinking about it, this change might be significant taking this snippet
```rust
let prepared = Command::all([
  Command::new(...),
  Command::event(MyEvent)
]);
// Stuff happens
prepared.into_future().await;
```
Without the changes, the `send_event` would only happen at the `await` point, but with my changes it would happen at the construction.

I think I should revert the changes for `send_event` and `notify_shell`

StuartHarris:
I don't think it makes a meaningful difference though. The event would go into the channel earlier, but it won't come out of the channel until the same moment (the poll).

StuartHarris:
I was a bit worried that the eager placement might change the event order, so I wrote a test to prove that, but it doesn't, so I think we're good!

PR #528, opened 2026-05-07
fix(commands): lazy event placement for immediate abort

StuartHarris:
#527 eagerly placed events into the command's internal queue under certain situations. This is not observable through the public API (they still get polled at the same time and in the same order). 

However, there was one very tiny edge case with respect to aborting a command _before_ it is first polled. This in practice would likely never be done (it probably doesn't make sense to create a command and cancel it straight away), but we want to preserve the old behaviour (before the change, the event would never have been placed in the queue because `run_until_settled()` short-circuits on abort, after the change the event is already in the queue, so it stays there).

This PR uses `futures::future::lazy` to regain the original lazy behaviour of only putting events in the queue when they are first polled, whilst keeping the ergonomics of the previous PR.

### DIFF
--- a/crux_core/src/command/mod.rs
+++ b/crux_core/src/command/mod.rs
@@ -384,10 +384,7 @@ where
     /// the event is not guaranteed to dispatch instantly - another `update` call which is
     /// already scheduled may happen first.
     pub fn event(event: Event) -> Self {
-        Command::new(|ctx| {
-            ctx.send_event(event);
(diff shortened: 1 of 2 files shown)


StuartHarris:
@ManevilleF we found a tiny edge case so made it lazy again, whilst keeping the cool things from your PR :-)

charypar (approved):
TIL about `future::lazy`. Neat!

PR #532, opened 2026-05-14
use classic `powershell.exe` instead of newer `pwsh.exe`

mknet:
@mhedgpeth 
To keep the barriers as low as possible, and because I don't see any need for the newer PowerShell, I would go with the more widely used executable.


mhedgpeth (approved):

mhedgpeth:
Thanks @mknet!

PR #533, opened 2026-05-18
Remove rustdoc-based typegen

StuartHarris:
This PR removes the `rustdoc` JSON typegen.

We were supporting 3 types of typegen (in order of preference): `facet_generate`, `serde_generate` and `rustdoc` JSON output.

The latter relied on Rust nightly and was therefore against a moving target. It also required a full `rustdoc` build which meant it was slow. The parsing of the JSON output and subsequent processing used logic programming and was complex and hard to maintain. Exploring through crate boundaries was really difficult 😅.
It was a useful experiment, but I don't think that it's worth pursuing, sadly. 

This is a precursor to removing `crux_cli` completely. The only surviving feature is bindgen for Kotlin, which we can move inside `crux_core` behind a gate.


charypar (approved):
🥲 rip. We can always bring it back later on when the whole rest of the stack stabilises, and we find something is impossible without it.

PR #534, opened 2026-05-18
Remove crux cli

StuartHarris:
The only remaining code in `crux_cli` was for Kotlin bindgen. This PR moves that functionality to `crux_core` behind the `bindgen` feature, and removes `crux_cli`.

When we publish next, we should publish a readme to https://crates.io/crates/crux_cli indicating that this crate is no more.


mhedgpeth:
Nice job! It threw me off that swift was removed but after some research I get why. From my research:

  - Swift — driven externally by `cargo-swift`. Look at `examples/counter-http/apple/Justfile:51` which calls `cargo swift package --features uniffi`, which under the hood invokes `uniffi-bindgen` and packages the result as an `.xcframework`
  with a `Package.swift`. So Crux doesn't need a Swift code path of its own; the toolchain already wraps it.
  - Kotlin — no equivalent off-the-shelf tool, so Crux invokes `uniffi_bindgen::library_mode::generate_bindings` with `KotlinBindingGenerator` itself, from `shared/src/bin/codegen.rs`. That's the function this PR moves into
  `crux_core::bindgen`.
  - TypeScript / Web — UniFFI not involved at all; bindings come from `wasm_bindgen`.
  - C# - is using FFI directly, not uniffi at the moment

mhedgpeth (approved):

PR #535, opened 2026-05-18
chore: clippy nursery

StuartHarris:
Fixes for `clippy::nursery`, including adding to CI

### DIFF
--- a/crux_http/src/testing/fake_shell.rs
+++ b/crux_http/src/testing/fake_shell.rs
@@ -9,7 +9,7 @@ use crate::protocol::{EffectSender, HttpRequest, HttpResponse, HttpResult};
 
 /// `FakeShell` implements `EffectSender` for use in our internal tests.
 #[derive(Clone, Default)]
-pub(crate) struct FakeShell {
+pub struct FakeShell {
     inner: Arc<Mutex<Inner>>,
 }
 
@@ -20,7 +20,7 @@ struct Inner {
 }
 
 impl FakeShell {
-    pub fn provide_response(&mut self, response: HttpResponse) {
+    pub fn provide_response(&self, response: HttpResponse) {
         self.inner
             .lock()
             .unwrap()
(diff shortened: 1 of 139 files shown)


ManevilleF:
Nice ! `clippy::nursery` is a gold mine of useful lints !

I would suggest adding the lints in the `Cargo.toml` directly and be able to selectively disable some, as lots in `nursery` (though less than in `pedantic`) can be false negatives.
I would also advise against using `missing_const_for_fn` or be particulary careful on which public functions become `const`, as rolling this back will be a breaking change

examples/notes/shared/src/app.rs
```diff
@@ -808,6 +808,7 @@ mod sync_tests {
                     .expect("should resolve");
 
                 if let Some(cmd) = self.command.as_mut() {
+                    #[allow(clippy::needless_collect)]
                     for event in cmd.events().collect::<Vec<_>>() {
```
charypar: I wonder whether something like `.cloned()` would be better and not trigger clippy...?

charypar (approved):
Not sure I agree with the push for `map_or_` – seems to hurt readability – but don't have a strong opinion (and forced consistency is nice.)

crux_http/src/testing/fake_shell.rs
```diff
@@ -9,7 +9,7 @@ use crate::protocol::{EffectSender, HttpRequest, HttpResponse, HttpResult};
 
 /// `FakeShell` implements `EffectSender` for use in our internal tests.
 #[derive(Clone, Default)]
-pub(crate) struct FakeShell {
+pub struct FakeShell {
```
mhedgpeth: I personally don't like this clippy, I think it's technically correct in that yes the parent shell's visibility is the real controlling factor, but it does express something here that the intent is that this be crate-level. On my projects I have this ignored.

mhedgpeth (approved):

examples/notes/shared/src/app.rs
StuartHarris: It would, but `Event` is not `Clone`. We can't avoid the collect, because we're borrowing `self` through `cmd`, so we then can't mutably borrow it again inside the loop. But I can convince clippy that the collect is load-bearing by extracting to a variable. Gonna do that.

crux_http/src/testing/fake_shell.rs
StuartHarris: More than happy to revert these and add a crate-wide exception.

ManevilleF:
A typical way would be to set the priority of the clippy lint group "nursery" to `-1` and then selectively disable some lints from that group like `map_or` or the crate visibility lints

PR #541, opened 2026-06-02
Migrate `counter-routing` from UniFFI to BoltFFI

StuartHarris:
## Migrate `counter-routing` from UniFFI to BoltFFI

The `counter-middleware` example was already using BoltFFI. This PR brings `counter-routing` in line with it, and fixes a wasm/Leptos rendering bug discovered in **both** examples along the way.

### What changed in `counter-routing`

**Shared crate (`shared/`)**

- Replaced the three FFI submodules (`uniffi`, `wasm_bindgen`, `wit_bindgen`) in `ffi.rs` with a single flat BoltFFI module.
- The routing pattern (`EffectRouter` / `Routes` / `RngHandler`) is preserved for the native path — `Random` is still handled entirely inside Rust and never reaches the native shells.
- The wasm path uses `Bridge<Core<Counter>>` directly and returns effects synchronously from `update`/`resolve`.
- Added `boltffi.toml` with target output paths and the `getrandom_backend` cfg flag for boltffi wasm builds.
- `codegen.rs`: removed `uniffi` and `bindgen_kotlin` dependencies; now matches `counter-middleware`.
- Removed `uniffi`, `wasm-bindgen`, `wit-bindgen`, `js-sys` dependencies; `rand` and `getrandom` moved to target-specific deps.
- Root `Justfile`: added `set -euo pipefail` to `ci` and `dev` so failures abort early.

**Android**

- Replaced rust-android-gradle plugin + JNA with `boltffi pack android` (pre-built `.so` files placed in `generated/jniLibs`).
- Removed `jna`, `rustAndroid` from `libs.versions.toml`.
- Cleaned up `shared/build.gradle.kts` and root `build.gradle.kts`.

**Apple**

- Replaced `cargo swift package --features uniffi` (+ header workaround) with `boltffi pack apple`.
- `core.swift`: updated type name `CoreFfi` → `CoreFFI`, constructor `CoreFfi(handler)` → `CoreFFI(shell: handler)`, method calls to BoltFFI labelled-parameter style (`data:`, `effectId:data:`).

**web-nextjs**

- Build: `wasm-pack build --features wasm_bindgen` → `boltffi pack wasm`.
- `core.ts`: constructor, `update`, and `resolve` updated for the BoltFFI API (passes a `processEffects` shell to `CoreFFI.new(...)` and consumes the effect bytes returned from `update`/`resolve`).
- `package.json`: added `@boltffi/runtime`, aligned package versions with `counter-middleware`.
- `tsconfig.json`: added `ignoreDeprecations: "6.0"`, excluded `generated/pkg/**/*.ts` from type-checking (works around a `number[]` vs `Uint8Array` type annotation issue in the generated BoltFFI TypeScript).
- `page.tsx`: updated wasm init pattern from `import init_core from "shared/shared"` to `import * as sharedWasm from "shared"` with `.initialized` promise.

**web-leptos**

- Bumped `wasm-bindgen`, `leptos`, `gloo-net`, `js-sys` to match `counter-middleware`.

---

### Wasm/Leptos fix applied to both `counter-routing` and `counter-middleware`

Both examples rendered a blank page in the Leptos shell. The root cause: `#[boltffi::export]` on the `CruxShell` trait generates callback stubs (`__boltffi_callback_crux_shell_process_effects`, `__boltffi_callback_crux_shell_free`) that are emitted into the wasm binary as imports from the `"env"` module. The BoltFFI JS runtime (used by web-nextjs) provides these; trunk, which links the `shared` crate directly into the Leptos wasm, does not.

**Why the obvious fix doesn't work.** The natural approach is to `#[cfg(not(target_family = "wasm"))]`-gate `CruxShell` and give the wasm `CoreFFI` a different, callback-free signature. But `boltffi`'s binding generator parses the Rust source and **does not evaluate `#[cfg]` at all** — it indexes every `#[boltffi::export]` item regardless of target. With two cfg-divergent `impl CoreFFI` blocks (`new(shell)` vs `new()`), bindgen emits a single, cfg-blind binding for the **wasm** variant, so the generated Kotlin/Swift bindings no longer match the native shells (or the native `.so`/`.dylib` ABI). This is what produced the `CoreFfi` "`Long` was expected" Kotlin compile error. Splitting the export is therefore a dead end.

**The fix (two parts):**

1. **One cfg-independent FFI signature.** `CoreFFI` is a single `#[boltffi::export] impl` with `new(shell)`, `update -> Vec<u8>`, `resolve -> Vec<u8>`, `view -> Vec<u8>`; `#[cfg]` appears only *inside* method bodies and on struct fields, never on export signatures. `CruxShell` is a single, un-gated export. This guarantees the generated bindings match what every target compiles:
   - **native** drives the `EffectRouter` (`Random` handled in Rust; effects delivered asynchronously via `CruxShell.processEffects`; `update`/`resolve` have no synchronous effects to return);
   - **wasm** drives `Bridge<Core<Counter>>` and returns effects synchronously from `update`/`resolve` (the callback is wired for API parity but, with no background Rust tasks on wasm, is effectively unused). `Random` is exposed to the JS shell.

2. **Feature-gate the FFI surface out of the Leptos build.** A new default `ffi` feature owns the (now `optional`) `boltffi` dependency, the `ffi` module, and the FFI-only `rng_handler`/`middleware` modules. Android/Apple and web-nextjs build with `ffi` on (symbols present; `env` imports satisfied by the BoltFFI JS runtime). `web-leptos` depends on `shared` with `default-features = false`, so it links no `boltffi` and emits no `env` imports. Because bindgen reads source (ignoring features as well as cfg), the generated bindings are unaffected by the gate.

The native shells are unchanged in behaviour: they still receive all effects through `CruxShell.processEffects`.

---

### Verification

- `cargo clippy -p shared` clean for both examples, native **and** `wasm32-unknown-unknown` (pedantic + warnings; nursery for `counter-middleware`).
- `shared` compiles both with `--no-default-features` (Leptos config) and with defaults (FFI on).
- Leptos: `cargo clippy -p web-leptos --target wasm32-unknown-unknown -- -D warnings` clean; the built Leptos wasm contains **zero** `boltffi`/`env` callback imports.
- Android: regenerated bindings expose `CoreFfi(shell: CruxShell)` / `boltffi_core_f_f_i_new(BoltFFICallbackHandle shell)`; `:app:assembleDebug` succeeds for both examples.
- Apple `boltffi pack apple` and the web shells verified locally.
- `cargo nextest run -p shared` → 6/6 for both examples.


StuartHarris:
This wasn't as easy as I hoped it would be. 😢  It's interesting that BoltFFI is making our simple examples simpler, but it's making our complex examples more complicated, especially where we have different mechanisms in play (apple/android vs web/ts vs web/rust).

StuartHarris:
But we can upstream a fix for BoltFFI to honour target cfg

Issue #357, opened 2025-05-02
Investigate gradual migration path of crux_http to `http`


charypar:
The HTTP capability currently uses `http-types`, which does not look actively maintained, and probably isn't the first choice for common HTTP types today. We also rely on a custom fork to avoid compatibility issues with emscripten (see https://github.com/redbadger/crux/pull/195).

From a quick look, it may be possible to switch to `http`, but maintain backwards compatibilty with `http-types` temporarily under a feature flag. That way we wouldn't force people to migrate or stay behind. It's not easy to see whether this is possible and how hard it might be without trying, which is what this issue is about. 

It is quite likely to make the code very annoying to read for the duration of maintaining both versions.

Sketch implementation:
* introduce a `http-types` feature flag to `crux_http`
* guard all uses of the types from the crate with a feature flag and... 
* switch to either an equivalent from `http` or to our own type if an equivalent doesn't exist (I think this is the case for the `Body` type, which is just a generic argument in `http`)

If this becomes far too messy, we will have to just bite the bullet and make people migrate the types as a step change, providing helpful conversion traits where we can, etc.

charypar:
A few hours investigation suggests this should be reasonably possible, but will require a few days work. 

I'll open a PR some time in the following days or weeks.

charypar:
Just for anyone watching this - this turned out really quite painful, and coupled with the other changes to HTTP, the more likely outcome is a hard switch from http-types to http - the impact might be quite high for people who use lots of details of HTTP, but I'm hopeful most have a central HTTP client for this case, so it's at least localised, and not scattered around in a hundred places.

In case you depend on the re-exported http-types heavily, please get in touch.

Issue #285, opened 2024-11-04
Migrate from `crux_http` to standard `http` crate types


m00nwtchr:
I'm using several other crates that are meant to interoperate with the types provided by the `http` crate, which makes using them with `crux_http` difficult. Using the semi-standard Response/Request/etc. types from the `http` crate in crux would rectify that. Alternatively, `Into`/`From` implementations could provide a good enough experience via conversions (ideally non-copying).

charypar:
This is definitely on our to-do list, but it would be quite a breaking change to `crux_http`, and will make providing a smooth migration path quite tricky. 

It's likely we will first release a `http` compatible alternative capability, and then migrate crux_http.

On the `Into` / `From` front, that sounds like good short term patch. Would you be interested in trying? We'd welcome a PR.

m00nwtchr:
Sure.

m00nwtchr:
Just noticed that the "http" name in `crux_http` is actually an alias of `http-types`, which is a different http "base-crate" from `http`. To be clear, do you mean that you want to eventually switch to `http` (and so I should focus on bringing in compatibility with that), or are you going to more fully embrace `http-types`?

(My original problem requires compatibility with `http`, but `http-types` has conversion implementations for `http` already, so I can go `crux_http` -> `http-types` -> `http`)

m00nwtchr:
Side Note: Looks like converting `http-types` to `http` requires `async` code in the core.... is there any way to achieve that?

m00nwtchr:
@charypar

charypar:
I've had another look at how we go about this and logged an investigation for actually migrating and dropping the dependncy on `http-types` https://github.com/redbadger/crux/issues/357

mhedgpeth:
Looks like this got released!

PR #543, opened 2026-06-09
RFC: Decouple `crux_http` from `http-types`

StuartHarris:
## 1. RFC to propose migrating `crux_http` to the `http` crate (with backwards compatibility behind a `http-types` feature).

Related issues:

- [#285 — Migrate from `crux_http` to standard `http` crate types](https://github.com/redbadger/crux/issues/285) (milestone 0.19)
- [#357 — Investigate gradual migration path of `crux_http` to `http`](https://github.com/redbadger/crux/issues/357)
- [#195 — emscripten compatibility (reason for the `http-types` fork)](https://github.com/redbadger/crux/issues/195)

## 2. Implementation of "Decouple `crux_http` from `http-types`" ([RFC](docs/src/rfcs/decouple-from-http-types.md)) + header API overhaul

Implements the full RFC. The headline change is that **`crux_http::http` now always re-exports the upstream [`http`](https://docs.rs/http) crate (v1.4)** instead of the internal `http-types-red-badger-temporary-fork`. The fork is removed from the default build entirely and is only pulled in when the new `http-types` feature is enabled.

With the `http` crate in place, all of the header-manipulation APIs were audited for consistency and idiomaticity. The result is a header API that mirrors `http::HeaderMap` directly — same parameter types, same return types, no silent failures.

---

#### `crux_http::http` is now the real `http` crate

Types previously sourced from `http_types` are replaced throughout:

| Before | After |
|---|---|
| `http_types::Method::Get` | `http::Method::GET` |
| `http_types::StatusCode::Unauthorized` | `http::StatusCode::UNAUTHORIZED` |
| `http_types::Headers` | `http::HeaderMap` |
| `http_types::Version` | `http::Version` |
| `http_types::Body` | `crux_http::Body` |
| `http_types::Mime` | `mime::Mime` (via `crux_http::mime`) |

`HttpError::Http { code, .. }` — `code` is now a plain `u16`. Compare directly (`code == 401`) or via `StatusCode::UNAUTHORIZED.as_u16()`.

#### New `crux_http::Body` type

A simple, synchronous, in-memory request body — `Vec<u8>` backed with an optional `mime::Mime` — replaces the async, streaming `http_types::Body`. The Crux effect protocol already collapses bodies to bytes before they reach the core, so there was never real streaming work to do there.

#### `crux_http::mime` re-export

The `mime` crate (v0.3) is now re-exported directly as `crux_http::mime`. Use `mime::APPLICATION_JSON`, `mime::TEXT_HTML`, etc. — no extra dependency needed.

#### `http-compat` feature removed; native conversions are unconditional

`From<http::Request<Body>> for crux_http::Request`, `From<crux_http::Request> for http::Request<Body>`, and `TryFrom<crux_http::Response<T>> for http::Response<T>` are now always available — no feature flag required. Remove `features = ["http-compat"]` from your `Cargo.toml`.

#### New `http-types` compat feature

For code that still bridges to `http_types::Request` / `http_types::Response` (e.g. existing middleware or shell code), add `features = ["http-types"]` to get `From`/`Into` conversion impls. This pulls in the fork only for that transition period.

#### `insert_header` and `append_header` now take `http::HeaderValue` directly

The low-level header mutation methods on `Request`, `RawResponse`, and `Response<T>` now mirror `http::HeaderMap` exactly:

- The value parameter changes from `impl AsRef<str>` to `HeaderValue`. This eliminates silent failure — the old signatures silently discarded any value that failed string validation, with no indication to the caller.
- `insert_header` returns `Option<HeaderValue>` (the evicted previous value, if any), matching `HeaderMap::insert`.
- `append_header` returns `bool` (whether a prior value already existed for that name), matching `HeaderMap::append`.

The high-level builder `.header(name, value)` on `RequestBuilder` and `ResponseBuilder` is **unchanged** — it still accepts `impl AsRef<str>` for convenience, and panics on invalid values.

```rust
// Before
req.insert_header("authorization", format!("Bearer {token}"));
res.append_header("set-cookie", "sid=abc");

// After
use crux_http::http::HeaderValue;
req.insert_header(
    "authorization",
    HeaderValue::from_str(&format!("Bearer {token}")).expect("valid token"),
);
res.append_header("set-cookie", HeaderValue::from_static("sid=abc"));
```

#### `header_all(name)` on `Request`, `RawResponse`, and `Response<T>`

All three types now expose `header_all`, returning all values for a given header name as an `http::header::GetAll` iterator. Previously only the response types had this — `Request` had no way to read multiple values for the same header name, which was a gap for middleware. The single-value `header(name)` still works and returns `Option<&HeaderValue>`.

#### `ResponseBuilder::append_header(name, value)`

The test helper now provides `append_header` alongside the existing `header` (which replaces). Use it when building mock responses with multiple values for the same header name — e.g. multiple `Set-Cookie` headers — without losing earlier values.

#### `Request::set_header` deprecated

`set_header` was an `http-types`-era alias that did nothing but call `insert_header`. It is now `#[deprecated]`. Replace all uses with `insert_header`.

#### `Config::add_header` name parameter is now `impl IntoHeaderName`

Previously `impl AsRef<str>`, which was inconsistent with every other header-setting method in the library. Now `impl http::header::IntoHeaderName`. Callers using `&'static str` literals are unaffected; callers building names from runtime strings should construct a `HeaderName` first.

#### `RawResponse` (was `ResponseAsync`) body methods are now synchronous

`body_bytes()`, `body_string()`, `body_json()`, and `body_form()` on `RawResponse` drop the `async`/`.await` — the body is already fully in memory, so the futures were always no-ops. Middleware and test helpers that were `async fn` only because of these calls can now be plain `fn`.

#### `Request::set_content_type` takes `&mime::Mime`

The low-level method now borrows rather than takes ownership. The builder-level `.content_type(…)` methods are unchanged.

#### Streaming bodies: use the `Chunk`/`Done` capability pattern

`AsyncRead` on `ResponseAsync` is removed. For genuinely streaming responses (large downloads, SSE, chunked transfer), use the `Chunk`/`Done` capability pattern shown in `examples/counter-http/shared/src/sse.rs` — the shell delivers bytes progressively, and the core processes a `Stream<Item = Vec<u8>>`. This is architecturally correct: I/O stays in the shell, the core stays pure.

---

### Bug fixes

- **Multi-value headers silently clobbered in all three `http-types` compat conversions.** `From<Request> for http_types::Request`, `From<http_types::Request> for Request`, and `From<http_types::Response> for RawResponse` were all calling `insert` inside the per-value loop, leaving only the last value for any header with multiple entries. Fixed to use `append` throughout.

- **Non-ASCII `Location` header in redirect middleware no longer causes a silent loop.** `HeaderValue::to_str()` rejects opaque bytes (≥ 0x80 — e.g. UTF-8 multi-byte characters). Previously the middleware called `unwrap_or("")`, turning a non-ASCII location into an empty string and re-requesting the same URL indefinitely. Now it returns `HttpError::Io` with a clear message.

- **Non-ASCII header values no longer forwarded as empty strings in protocol requests.** `From<Request> for HttpRequest` was mapping opaque header values to `""` via `to_str().unwrap_or("")`. The empty string is misleading — the header name is preserved but the value is lost. Fixed to `filter_map` such values out entirely, so the shell never receives a header with a silently corrupted value.

- **`Config::headers` are now actually applied to every outgoing request.** The per-client default headers set via `Config::add_header` were stored but never inserted into requests in `Client::send`. Fixed with default-wins-if-absent semantics: config headers are appended only when the request doesn't already carry a value for that name, so per-request headers always take precedence.

---

### Breaking changes summary

Most apps using `crux_http` through the high-level API (`Http::get(…)`, `RequestBuilder`, `Response<T>`) will compile with minimal changes. The full migration checklist is in the new **[migration guide](docs/src/guide/migrate-crux-http.md)**.

Quick summary for middleware and shell authors:

| What changed | Action |
|---|---|
| `crux_http::http` is now the real `http` crate | See migration guide — most high-level code compiles unchanged |
| `insert_header` / `append_header` take `HeaderValue`, return `Option<HeaderValue>` / `bool` | Construct `HeaderValue` at call site; use return value if needed |
| `ResponseAsync` → `RawResponse` | Update imports and type annotations |
| `RawResponse` body methods are synchronous | Remove `.await` |
| `Config::add_header` name is `impl IntoHeaderName` | `&'static str` literals unchanged; dynamic names need `HeaderName::from_bytes(…)?` |
| `Request::set_content_type` borrows `&mime::Mime` | Add `&` |
| `Request::set_header` deprecated | Replace with `insert_header` |
| `http-compat` feature removed | Remove feature flag from `Cargo.toml` |

---

### Tests

- All existing unit, integration, and doc tests pass on the default (`http`) path.
- `protocol.rs` insta snapshots are byte-for-byte unchanged — confirming the wire format between shell and core is unaffected.
- New round-trip integration tests:
  - `http::Request<Body>` → `crux_http::Request` → `HttpRequest`
  - `HttpResponse` → `crux_http::RawResponse` → `crux_http::Response<Vec<u8>>` → `http::Response<Vec<u8>>`
- `header_all_returns_multiple_values` now actually asserts both values are present (the old test only checked the API compiled).
- CI now runs `cargo nextest run -p crux_http --features http-types` as a separate job to keep the compat path green.
- `counter-http`, `counter-middleware`, `weather`, and `counter-routing` examples all build on both feature configurations.
- New regression tests for the bug fixes above:
  - `compat`: `request_to_http_types_preserves_multi_value_headers`, `http_types_request_to_request_preserves_multi_value_headers`
  - `redirect`: `follows_absolute_redirect`, `follows_relative_redirect`, `non_ascii_location_header_returns_io_error`, `stops_after_max_attempts`
  - `protocol`: `non_ascii_header_values_are_omitted_from_protocol_request`
  - `client`: `config_headers_are_sent_with_every_request`, `per_request_header_takes_precedence_over_config_header`

## 3. Rename `ResponseAsync` to `RawResponse`

This no longer has any async in it so it was misleading to call it this. May as well change it now while we're making breaking changes. 😓


crux_http/CHANGELOG.md
```diff
@@ -8,6 +8,104 @@ and this project adheres to
...
+  header.
+
+### ⚙️ Miscellaneous Tasks
+
+- `http_types` (the temporary fork) is no longer a default dependency; it is pulled in
+  only when the `http-types` feature is enabled, reducing the default dependency footprint.
+- `into_protocol_request` is now a synchronous function (was `async`); the previous
+  `await` on `http_types::Body::into_bytes()` is gone.
+- `ResponseAsync` renamed to `RawResponse` throughout (file remains `response_async.rs`
+  internally; the public name is what changed).
```
StuartHarris: The file has been renamed so this can go

crux_http/src/request_builder.rs
```diff
@@ -76,7 +75,11 @@ where
     /// ```
     /// # Panics
     /// Panics if the `RequestBuilder` has not been initialized.
-    pub fn header(mut self, key: impl Into<HeaderName>, value: impl ToHeaderValues) -> Self {
+    pub fn header(
+        mut self,
+        key: impl http::header::IntoHeaderName,
```
StuartHarris: Note to self. We use AsRef elsewhere. Check this

crux_http/src/request_builder.rs
```diff
@@ -120,16 +123,16 @@ where
...
     ///     .body(json!({"any": "Into<Body>"}))
-    ///     .content_type(mime::HTML)
+    ///     .content_type(mime::TEXT_HTML)
     ///     .build()
     ///     .then_send(Event::ReceiveResponse);
     /// ```
     /// # Panics
     /// Panics if the `RequestBuilder` has not been initialized.
-    pub fn body(mut self, body: impl Into<Body>) -> Self {
+    pub fn body(mut self, body: impl Into<crate::body::Body>) -> Self {
```
StuartHarris: There's a bunch of places where we could have a use instead

crux_http/src/response/response.rs
```diff
@@ -82,76 +85,86 @@ impl<Body> Response<Body> {
     /// #   .build();
     /// assert!(res.header("Content-Length").is_some());
     /// ```
-    pub fn header(&self, name: impl Into<HeaderName>) -> Option<&HeaderValues> {
+    pub fn header(&self, name: impl http::header::AsHeaderName) -> Option<&HeaderValue> {
```
StuartHarris, line 88: See above comment about unifying the header name and value parameters

crux_http/src/response/response.rs
```diff
@@ -364,98 +334,168 @@ where
...
+        let res = ResponseBuilder::ok().header("x-custom", "hello").build();
+        let map: &HeaderMap = res.as_ref();
+        assert_eq!(map["x-custom"], "hello");
+    }
+
+    #[test]
+    fn header_all_returns_multiple_values() {
+        // header_all() API compiles and returns values even when only one is present
+        // (ResponseBuilder uses insert_header which replaces).
+        let res = ResponseBuilder::ok()
```
StuartHarris, line 384: Need to think a bit more about multiple headers with the same name

mknet:
Looks awesome to me.

Two things I’d like clarity on:

1) Emscripten / http-types
Good that the fork stays behind http-types ([#195](https://github.com/redbadger/crux/issues/195)). Is emscripten still a first-class supported target, or mainly legacy? The default http path is what most WASM users need (unknown-unknown); emscripten feels rare today. Would you document emscripten as best-effort / http-types only, and confirm CI covers default WASM without that feature?

2) Status codes via Ok
Since the protocol snapshots are unchanged: 404/500/etc. still arrive as HttpResult::Ok with status set, not Err. Worth a prominent migration-guide note or example? Initially, I expected `Err` for “HTTP errors” but the idea seems to be only transport failures use `Err(Url|Io|Timeout)`.


Update:

3) Builder panics vs. protocol
Worth a migration-guide note that `.header()` panics on invalid values only on the internal Request API, while `protocol::HttpRequest` remains unvalidated strings? We want to avoid shell authors assuming header checks happen at the FFI boundary.

StuartHarris:
> Looks awesome to me.
> 
> Two things I’d like clarity on:
> 
> 1. Emscripten / http-types
>    Good that the fork stays behind http-types ([#195](https://github.com/redbadger/crux/issues/195)). Is emscripten still a first-class supported target, or mainly legacy? The default http path is what most WASM users need (unknown-unknown); emscripten feels rare today. Would you document emscripten as best-effort / http-types only, and confirm CI covers default WASM without that feature?
> 2. Status codes via Ok
>    Since the protocol snapshots are unchanged: 404/500/etc. still arrive as HttpResult::Ok with status set, not Err. Worth a prominent migration-guide note or example? Initially, I expected `Err` for “HTTP errors” but the idea seems to be only transport failures use `Err(Url|Io|Timeout)`.
> 
> Update:
> 
> 3. Builder panics vs. protocol
>    Worth a migration-guide note that `.header()` panics on invalid values only on the internal Request API, while `protocol::HttpRequest` remains unvalidated strings? We want to avoid shell authors assuming header checks happen at the FFI boundary.

Thanks @mknet, much appreciated. I added 3 commits, one for each.

1. Now we're using `http` as the default, users should be able to compile to any wasm target (there was a 128bit trace_id in `http-types` which we changed to 64bit in our fork to support users on Emscripten because they have  C++ modules, but this shouldn't be an issue anymore). So Emscripten should still work, but we don't need to do anything special anymore for the default use case. I added a CI check for the regular wasm target, but nothing really changes here.
2. Yeah, I added a changelog note, and doc comments. Thanks
3. Added a changelog note, doc comments and a tweak to the migration guide.

StuartHarris:
@charypar @mhedgpeth I want to merge this and publish as we have folks waiting for it (#547). I've reviewed it several times, added tests, fixed bugs and I'm confident it's OK to merge. Last chance if you want to lay your 👀 on it :-)

PR #544, opened 2026-06-17
Flutter: counter example

mknet:
A simple example how to use the Flutter Rust Bridge to integrate a Crux core a Flutter app for several OS'.


mknet:
Has anyone got already a chance to check the example?

PR #546, opened 2026-06-19
fix: build for examples with MacOS shells

StuartHarris:
Temporary workaround for [this problem](https://github.com/boltffi/boltffi/issues/560) when building MacOS shells in examples

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -871,9 +871,9 @@ dependencies = [
 
 [[package]]
 name = "facet_generate"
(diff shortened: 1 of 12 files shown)


mhedgpeth:
apologies for the delay on this @StuartHarris I'm able to reproduce this and it's now blocking my own development. I'm going to try to get this released as a hot fix today on the `boltffi` side as a part of the issue you reported there, and will either close this or update it depending on what happens.

PR #549, opened 2026-07-05
Clarify building instructions for Android

ganyuss:
When following the Android shell tutorial. I got a bit confused at what to do when reading this line: 

> If you now build your project, BoltFFI and the type generator populate the`Android/generated` folder.

This PR aims to clarify that step by explicitly writing the shell commands to build when they are needed. 

---
PS: Thank you for this tool (framework?), I am only getting started, but it is already really exciting!


StuartHarris (approved):
Thank you for this, and glad you're enjoying Crux.

PR #554, opened 2026-07-27
fix(http): make a rejection's body reachable, and stop tests asserting a state apps can't receive

StuartHarris:
## Why

`crux_http` converts every 4xx/5xx into `Err(HttpError::Http { code, message, body })`
before an app sees it, so an `Ok(Response)` can never carry an error status. That
invariant was undocumented, and two parts of the API actively pointed away from it:

- `Response::status()` is public, prominent, and documented with
  `assert_eq!(res.status(), 200)` — nothing said the status on the `Ok` side is always
  non-error, so checking it for failure reads like sensible defensive code.
- `ResponseBuilder::with_status(409)` let a **test** construct a `Response` with an error
  status — a value a feature can never receive.

In a downstream app (an HR service with a Crux UI), those two facts together produced
**eight independent call sites** shaped like this:

```rust
match result {                                   // crux_http::Result<Response<Vec<u8>>>
    Ok(mut response) => {
        if (200_u16..300).contains(&response.status().into()) {
            /* success */
        } else {
            ctx.error(api::error_reason(&mut response));  // read {"error": …} from the body
        }
    }
    Err(error) => ctx.fail("Failed to …", &error),
}
```

The `else` branch is unreachable. Every one of those eight sites showed users
`"Failed to …: HTTP error 409: 409 Conflict"` where the service had sent a sentence
written specifically for them — *"assignment would create a management cycle"*, *"Those
dates overlap a request you already have from 3 to 7 April 2026"*.

The part that makes this a library problem rather than eight authorial mistakes: **all
seven pre-existing sites had passing tests for their rejection message.** Each test
hand-built the impossible state with
`ResponseBuilder::with_status(409).body(…)` and fabricated `Ok(that_response)`, so the
tests asserted the dead branch and left the live path uncovered. Code review passed all
eight.

## What changed

Three commits, in the order they build on each other:

**1. `HttpError::code`, `HttpError::body`, `HttpError::body_json` + `testing::rejection`**

The rejection body was already preserved — it just wasn't reachable without
destructuring:

```rust
// before, hand-rolled in every app
let HttpError::Http { body: Some(body), .. } = error else { return None };
serde_json::from_slice::<Value>(body).ok()
    .and_then(|b| b["error"].as_str().map(String::from))

// after
error.body_json::<Value>().ok().and_then(|b| b["error"].as_str().map(String::from))
```

`body_json` deserializes into whatever shape the service uses — an app's own envelope,
an RFC 7807 `problem+json` struct, or `serde_json::Value` — so **no `problem+json` type
is added**; generic access is the requirement. An empty body reads as `None`, because
every real 4xx arrives with `Some(bytes)` and `Some(&[])` would make the obvious
`if let Some(body)` idiom display blanks.

`crux_http::testing::rejection(status, body)` builds the `crux_http::Result<Response<Body>>`
a feature genuinely receives for a rejection:

```rust
let event = Event::Saved(crux_http::testing::rejection(409, r#"{"error":"…"}"#));
```

It runs the *real* conversion (protocol response → `RawResponse` → `Response::new`)
rather than assembling the error by hand, and a test pins that its output is equal to
what the machinery produces. It panics for a non-error status, so it can't be misused
for the success case.

**2. Documentation of the invariant**

Stated on `Response` (with the correct match shape as a compiled example), on
`Response::status`, on `Response::new` where the invariant is established, on
`HttpError` / `HttpError::Http`, on the new `testing` module docs, and in the crate root.
`HttpError::Http`'s docs also spell out that `Display` shows only the status, which is
what makes `error.to_string()` such a lossy thing to show a user.

**3. `ResponseBuilder::with_status` now panics for 4xx/5xx** (breaking, test-only)

This is the item that would actually have caught the downstream bug: it turns those
seven green-but-dead tests red, with a message naming `testing::rejection` as the
replacement. Migration is mechanical:

```rust
// before — asserts a state the app can never observe
let result = Ok(ResponseBuilder::with_status(409).body(body).build());
// after — what the app really receives
let result = crux_http::testing::rejection(409, body);
```

Nothing about the runtime behaviour of a request changes. Non-error statuses, including
non-standard ones, are unaffected. Two of `crux_http`'s own tests built a 499 `Response`;
one is now covered by the panic test, and the status-serialisation round-trip uses 299
(still non-standard, but a status a `Response` can actually hold).

## Tests

- `crux_http/tests/rejections.rs` (new) drives a real `Command` resolved with a real
  `HttpResult`, and pins: a rejection arrives as `Err` with the server's message intact;
  the raw error body survives a `expect_json::<T>()` request (the expectation decode never
  runs on an error status, because `and_then` short-circuits); `testing::rejection`'s value
  equals what the machinery delivers.
- `ResponseBuilder` tests: builds any non-error status, panics for 409 and 503. **These two
  panic tests fail before this change** — that is the point.
- `HttpError` accessor tests, and `testing::rejection` tests including equality with the
  shell path.
- Doc examples throughout are compiled.

## Deliberately not done

- **No reqwest-style semantics** (`Ok` for 4xx plus an opt-in `error_for_status()`).
  Arguably the better design, but the migration would be *silent*: every existing crux app
  that propagates with `?` would start treating a 409 as success, with no compile error
  anywhere. Not worth it without a much larger, deliberate migration.
- ~~**`HttpError::Http` keeps `body: Option<Vec<u8>>`; it does not carry the whole
  `Response<Vec<u8>>`.**~~ **Superseded by #555**, which adds the headers after all —
  they turned out to be unreachable by any route, which the reasoning below missed. Kept
  for the record: Headers would be nice (to tell `problem+json` from an app's own
  envelope), and the variant is `#[serde(skip)] #[facet(skip)]` so this isn't FFI-breaking.

  One obstacle I initially thought was decisive turned out not to be: `#[facet(skip)]` on a
  variant does **not** exempt its field types from `Facet` (adding a `Response<Vec<u8>>`
  field to a skipped variant fails with "the trait `Facet<'_>` is not implemented for
  `Response<Vec<u8>>`") — but `#[facet(opaque)]` on the field waives exactly that, which I
  verified compiles, and which is what `HttpError` carried for its `StatusCode` before
  0.19. The derive is not a blocker.

  What decided it is cost against demonstrated need:
  1. Two of the three construction sites have no response to supply
     (`Response::body_bytes`'s "body had no bytes", and `From<http_types::Error>`), so the
     field would be `Option<Response<…>>` sitting alongside `code`/`message` — three fields
     where two overlap.
  2. It breaks every downstream `match` on the variant's fields **and** every construction
     of it, including app tests that build one, for a benefit nothing has asked for yet.
     `body_json::<T>()` is content-type-agnostic, so `problem+json` is already readable
     without headers.

  Adding the accessors first is the hedge: the shape to move to is
  `response: Option<Response<Vec<u8>>>` with `#[facet(opaque)]` (or `Option<HttpResponse>`,
  the protocol type, which is `Facet` already), and readers of `body()` / `body_json()`
  won't notice the change.

## CI

`cargo fmt --check`, `cargo check`, `cargo clippy --all-targets -Dpedantic -Dnursery`,
`cargo build`, `cargo nextest run` and `cargo test --doc` are all green for the root
workspace (all features). No insta snapshots changed.

`just ci` cannot complete on my machine: `just examples/check` fails in
`examples/counter/web-nextjs` at `pnpm install`, because
`examples/counter/web-nextjs/generated/pkg` doesn't exist until the BoltFFI wasm package
is built. That failure is environmental and is hit before any `crux_http`-dependent
example. To compensate I ran, per example, `cargo clippy --all-targets --all-features
-Dpedantic -Dnursery` and the test suites for `counter`, `counter-http`,
`counter-middleware`, `counter-routing`, `notes` and `weather` shared crates: all tests
pass. Two pre-existing failures unrelated to this PR, for the record:
`counter-routing/shared` has four clippy pedantic/nursery findings in `app.rs` and
`capabilities/sse.rs`, and `notes/shared`'s `starts_a_timer_after_an_edit` fails under
`cargo test` (a `TimerId` counter shared between tests) but passes under `cargo nextest`,
which is what CI uses.

## Review follow-up

- `HttpError::code`'s doc no longer claims that processing errors have no status.
  `Response::body_bytes` reports an already-taken body as an `Http` error carrying that
  response's own (non-error) status, so a `Some` is not by itself proof of a rejection.
  Giving that case a variant of its own would be the better fix, and a third breaking
  change — left for its own PR.

### DIFF
--- a/crux_http/CHANGELOG.md
+++ b/crux_http/CHANGELOG.md
@@ -10,6 +10,46 @@ and this project adheres to
 
 ### Added
 
+- **`HttpError::body`, `HttpError::body_json` and `HttpError::code`** — read what the
+  server said when it rejected a request, without destructuring the variant.
+
+  A 4xx/5xx arrives as `Err(HttpError::Http { code, message, body })`, and `body` has
+  always held the server's own explanation. But `Display` shows only the status
+  (`"HTTP error 409: 409 Conflict"`), so getting at the message meant hand-rolling
+  this in every app:
+
+  ```rust
+  // before
+  let crux_http::HttpError::Http { body: Some(body), .. } = error else { return None };
+  serde_json::from_slice::<serde_json::Value>(body)
+      .ok()
+      .and_then(|b| b.get("error").and_then(|e| e.as_str()).map(String::from))
+
+  // after
+  error.body_json::<serde_json::Value>().ok()
+      .and_then(|b| b["error"].as_str().map(String::from))
+  ```
+
+  `body_json` deserializes into whatever shape your API uses — your own envelope, an
+  RFC 7807 `problem+json` struct, or `serde_json::Value`. Note that body decoding is
+  skipped for an error status, so the raw error body survives even for a request
+  built with `expect_json::<T>()`; there is now a test pinning that.
+
+- **`crux_http::testing::rejection(status, body)`** — builds the
+  `crux_http::Result<Response<Body>>` a feature receives when the server rejects a
+  request, via the same conversion a real shell response takes:
+
+  ```rust
+  let event = Event::Saved(crux_http::testing::rejection(409, r#"{"error":"…"}"#));
+  ```
+
+  Use it wherever you would previously have fabricated `Ok(response_with_409)` (see
+  the breaking change below).
+
+- `crux_http::testing` now has module docs, stating the invariant the two builders
+  divide: `ResponseBuilder` for the `Ok(Response)` of a successful exchange,
+  `rejection` for the `Err` of a rejection. There is no third case.
+
 - `HttpRequestBuilder::body_json`, which sets a JSON body **and**
   `content-type: application/json`, mirroring what
   `command::RequestBuilder::body_json` puts on the wire.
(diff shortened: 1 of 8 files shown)


StuartHarris:
Related: #556 — while checking whether a rejection's headers were reachable any other way, I found that `crux_http` middleware never runs on the command API (`command::RequestBuilder::middleware` is silently discarded). Neither of these PRs fixes it.

mhedgpeth (approved):
Nice work

PR #555, opened 2026-07-27
feat(http)!: carry the rejected response's headers on HttpError::Http

StuartHarris:
> **Stacked on #554** — base it on `http-rejection-ergonomics`, or merge #554 first and this
> becomes three commits against `master` (one feature, two docs). The accessors #554 adds
> (`HttpError::body`, `body_json`, `code`, and `testing::rejection`) are what make this
> change small.

## Why

#554 deliberately left `HttpError::Http` alone, on the grounds that nothing had asked for a
rejection's headers. That was wrong, and checking it turned up why: **the headers are
currently unreachable by any route.**

- `Response::new` discards the `HeaderMap` on the error branch — 4xx/5xx keeps only
  `code`, `message` and `body`.
- `crux_http`'s own middleware, which *does* see the full `RawResponse` (that is how
  `Redirect` reads `Location`), **never runs in the command API**: `command::RequestBuilder::build`
  goes straight `into_protocol_request()` → `request_from_shell`, and the only executor of a
  middleware stack is `Client::send`, reached solely from the deprecated
  `request_builder.rs`. `command::RequestBuilder::middleware(…)` is accepted and silently
  dropped. (Separate bug — #556. Not fixed here, but no longer undocumented; see below.)

So for an app on the current API, everything a server says in the headers of a rejection is
gone. That is not cosmetic:

| Header | Why it can't be recovered otherwise |
| --- | --- |
| `Retry-After` (429, 503) | Not in the status, not in the body. An app cannot back off politely or say "try again in 5 minutes". |
| `WWW-Authenticate` (401) | `Bearer error="invalid_token"` vs insufficient scope decides silent refresh vs re-login. |
| `X-RateLimit-*` | Same. |
| `Content-Type` | Tells an app envelope from an RFC 7807 document from a proxy's HTML page — i.e. whether the body is worth deserializing, and whether it is safe to show as text. |

## What changed

**`HttpError::Http` grows a `headers` field**, read through three accessors that sit
alongside the `body`/`body_json` pair from #554:

```rust
if let Some(retry_after) = error.header("Retry-After") { /* back off */ }   // case-insensitive
match error.content_type() { … }
error.headers()   // escape hatch: multi-value, iteration, logging
```

The field is `Option<Box<HeaderMap>>`:

- `Option`, so `None` means "there was no response to describe" (a transport failure, or a
  status code that wasn't valid HTTP) as distinct from a response that sent no headers.
- `Box`, because a bare `HeaderMap` is **96 bytes**, which took `HttpError` to 160 and
  tripped `clippy::result_large_err` in `config.rs`, `expect.rs` and `error.rs` — this type
  is the `Err` of nearly every function in the crate. Boxed, it's 8 bytes and `HttpError` is
  back to 72. (Measured, not guessed.)

I chose `HeaderMap` over the protocol's `Vec<HttpHeader>` (which would have needed no
`#[facet(opaque)]` and no box) so that case-insensitive lookup and multi-value handling are
**inherited rather than re-implemented** — this crate has already been bitten by multi-value
header bugs, per the 0.19 changelog — and so that `header()`/`content_type()` have exactly
the same signatures as they do on `Response`.

**The variant is now `#[non_exhaustive]`.** Since this PR is already breaking construction of
it, this is the moment to make that the *last* time: only `crux_http` constructs the variant,
so what a rejection carries can keep growing behind the accessors. It also pushes callers
onto `{ code, .. }` plus `header()`/`body()`, which is the shape both PRs are arguing for.

**`testing::rejection_from(HttpResponse)`** is the header-carrying test helper, taking the
same protocol response you would resolve a request with, so both styles of test describe a
rejection identically. `rejection(status, body)` is unchanged and is now sugar over it —
with a test pinning that equivalence.

**Two docs commits ride along:**

- `command::RequestBuilder::middleware` gets a `# Warning` that nothing executes the stack —
  `Redirect` included, so redirects are not followed — pointing at #556. Its example
  previously implied otherwise. Documents existing behaviour; fixes nothing.
- A new book page, [Handling `crux_http` rejections](docs/src/guide/http-rejections.md),
  covering the invariant, the dead-`else` trap, the accessors, and the two test builders,
  linked from `SUMMARY.md` and from `part-2/capabilities.md` where the weather example's
  `Result<Response, HttpError>` is first explained.

## Migration

`match` arms that name only the fields they use are unaffected — including the pattern in
this repo's own weather example:

```rust
Err(HttpError::Http { code, .. }) if code == 401 => { … }   // still compiles
```

Code that *constructs* the variant can no longer do so at all (`#[non_exhaustive]`). In
tests, that was the wrong move anyway:

```rust
// before
HttpError::Http { code: 409, message: "409 Conflict".into(), body: Some(body) }
// after
crux_http::testing::rejection::<Vec<u8>>(409, body).unwrap_err()
```

One subtlety worth knowing if you have tests comparing whole errors: `PartialEq` now
includes the headers, and `rejection(409, body)` carries `Some(<empty map>)`, which does not
equal a hand-built `headers: None`. Compare against `rejection` / `rejection_from` rather
than a value assembled by hand.

## Tests

- `tests/rejections.rs`: a 503 resolved through a real `Command` — `Retry-After`,
  `Content-Type` and body all reach the app. This is the test that fails without the change.
- `testing::rejection` still equals what the shell path produces, now also for a
  header-carrying rejection; `rejection` proven equal to `rejection_from` with no headers.
- Accessor unit tests: case-insensitive lookup, `content_type` parsing, `problem+json`,
  `WWW-Authenticate`, and the three-way distinction between no response, an empty header
  map, and a populated one.
- Both helpers' panics assert the *calling* function's name, so `rejection_from` can't
  report itself as `rejection`.

## Review follow-ups

Applied after review, each in the commit that introduced the problem:

- `Response::body_bytes` no longer clones the `HeaderMap` on the success path. The
  `ok_or_else` closure had to capture the clone, so every successful `body_bytes` /
  `body_string` / `body_json` paid for a map it threw away; it's a `match` now, with a
  comment saying why.
- `HttpError::headers`' doc said "empty for errors with no response behind them" when it
  returns `None` — which contradicted the deliberate `None` vs `Some(&empty)` distinction.
- `HttpError::code`'s doc (in #554) claimed processing errors have no status. Not true:
  `Response::body_bytes` reports an already-taken body as an `Http` error carrying that
  response's own non-error status, so `Some` isn't proof of a rejection. Now says so.
  Giving that case its own variant would be a better fix and a third breaking change —
  left for its own PR.
- The shared panic path named `rejection` even when `rejection_from` was called.

## Not done here

- **The middleware no-op** (#556). It needs `build()` to route through `Client::send`, which
  needs an `EffectSender` backed by the command context — doable (the trait is the right
  shape), but it's a behaviour change (redirects would start being followed) and a separate
  discussion about whether `crux_http` middleware survives the capability deprecation. This
  PR only stops the doc example promising redirect-following that does not happen.
- **A separate variant for "body had no bytes"**, which is not a rejection but now presents
  as one, complete with headers. Third breaking change; wants its own PR.
- No RFC 7807 type. `body_json::<T>()` plus `content_type()` covers it without the crate
  taking a position on the shape.

## CI

`cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -Dpedantic
-Dnursery`, `cargo nextest run --all-features` (239 workspace tests, 110 in `crux_http`),
`cargo test --doc --all-features` (75 `crux_http` doc tests), `cargo doc` with
`-D rustdoc::broken_intra_doc_links`, and `mdbook build` with the linkcheck backend are all
green locally, as is #554's tip on its own. Same caveat as #554: `just examples/check` can't
complete on my machine because `examples/counter/web-nextjs` needs a BoltFFI-generated wasm
package before `pnpm install` will resolve; the per-example shared-crate clippy and test runs
pass.

### DIFF
--- a/crux_http/CHANGELOG.md
+++ b/crux_http/CHANGELOG.md
@@ -181,6 +181,12 @@ and this project adheres to
   rejection can only be handled on the `Err` side. The `Response` docs carry the
   correct match shape as a compiled example.
 
+- `command::RequestBuilder::middleware` now warns that it is a **no-op**: nothing on the
+  command API executes a middleware stack, so middleware pushed there — `Redirect`
+  included — is accepted and silently ignored. Its example previously implied redirects
+  were followed. This documents existing behaviour; the underlying gap is
+  [#556](https://github.com/redbadger/crux/issues/556).
+
 ## [0.19.0](https://github.com/redbadger/crux/compare/crux_http-v0.18.0...crux_http-v0.19.0) - 2026-07-06
 
 > **📖 See the [Migrating `crux_http` to native `http` types](https://redbadger.github.io/crux/guide/migrate-crux-http.html)
--- a/crux_http/CHANGELOG.md
+++ b/crux_http/CHANGELOG.md
@@ -35,6 +35,22 @@ and this project adheres to
   skipped for an error status, so the raw error body survives even for a request
   built with `expect_json::<T>()`; there is now a test pinning that.
 
+- **`HttpError::header`, `HttpError::headers` and `HttpError::content_type`** — read the
+  headers of the response that was rejected.
+
+  A rejection's *policy* often lives only in its headers, and none of it is recoverable
+  from the status or the body: `Retry-After` on a 429 or 503, `WWW-Authenticate` on a 401
+  (expired token vs insufficient scope), rate-limit headers, or the `Content-Type` that
+  says whether an error body is JSON, an RFC 7807 document, or a proxy's HTML page.
+  `Response::new` used to drop the `HeaderMap` on the error branch, and since `crux_http`
+  middleware does not run in the command API, an app had no way to see any of it.
+
+  ```rust
+  if let Some(retry_after) = error.header("retry-after") { /* back off politely */ }
+  ```
+
+  This adds a `headers` field to `HttpError::Http` — see the breaking change below.
+
 - **`crux_http::testing::rejection(status, body)`** — builds the
   `crux_http::Result<Response<Body>>` a feature receives when the server rejects a
   request, via the same conversion a real shell response takes:
@@ -46,9 +62,19 @@ and this project adheres to
   Use it wherever you would previously have fabricated `Ok(response_with_409)` (see
   the breaking change below).
 
-- `crux_http::testing` now has module docs, stating the invariant the two builders
-  divide: `ResponseBuilder` for the `Ok(Response)` of a successful exchange,
-  `rejection` for the `Err` of a rejection. There is no third case.
+- **`crux_http::testing::rejection_from(HttpResponse)`** — the header-carrying form,
+  taking the same protocol response you would resolve a request with in an end-to-end
+  test, so both styles of test describe a rejection the same way:
+
+  ```rust
+  let result = rejection_from(HttpResponse::status(429).header("retry-after", "30").build());
+  ```
+
+  `rejection(status, body)` is unchanged, and is now sugar over it.
+
+- `crux_http::testing` now has module docs, stating the invariant the builders divide:
+  `ResponseBuilder` for the `Ok(Response)` of a successful exchange, `rejection` /
+  `rejection_from` for the `Err` of a rejection. There is no third case.
 
 - `HttpRequestBuilder::body_json`, which sets a JSON body **and**
   `content-type: application/json`, mirroring what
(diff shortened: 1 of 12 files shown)


StuartHarris:
Related: #556 — while checking whether a rejection's headers were reachable any other way, I found that `crux_http` middleware never runs on the command API (`command::RequestBuilder::middleware` is silently discarded). Neither of these PRs fixes it.

mhedgpeth (approved):
Nice work Stuart!

PR #561, opened 2026-08-04
Update BoltFFI to 0.29.3

### LINKED ISSUE #558: chore: update BoltFFI to 0.29.1
I want to update Crux to BoltFFI 0.29.1, which includes arm64 support for Windows and a refactoring of the C# approach to BoltFFI's new model.

Update BoltFFI references consistently across the repository, including Rust dependencies, CLI instructions, examples, docs, web runtime packages, lockfiles, and Renovate configuration.

Done when:

- All BoltFFI dependencies and CLI installs use 0.29.1.
- No active references remain pinned to 0.25.2.
- The counter Windows example uses the new C# model, supports arm64 Windows, and builds its C# bindings successfully.
- Relevant checks pass.

### LINKED ISSUE #560: bug: Apple example packaging fails during BoltFFI Swift generation
## Summary

On a fresh checkout of `master`, the `counter` and `weather` Apple examples fail during `boltffi pack apple` after the Apple targets build successfully.

The failure is:

```
error: command failed: generate swift: no BoltFFI binding metadata found in compiled artifacts:
[.../target/debug/libshared.dylib, .../libshared.rlib, .../libshared.a]
```

## Reproduction

1. Check out `master`.
2. Install the documented prerequisites and `boltffi_cli`.
3. Confirm `just doctor` reports all checks green.
4. Run the Apple example packaging recipe for `counter` or `weather`.

## Expected behavior

The examples should generate Swift bindings and complete Apple packaging.

## Investigation

Both examples fail identically and use `#[boltffi::export]`. The Apple targets themselves build successfully, so the failure occurs during BoltFFI metadata-based Swift generation.

`master` pins the Rust BoltFFI crate and documented CLI to `0.25.2`, but `just doctor` only verifies that `boltffi` exists on `PATH`; it does not verify the CLI version matches the Rust dependency. A coordinated BoltFFI upgrade is tracked in #558.

mhedgpeth:
Closes #558.
Closes #560.

I updated Crux to BoltFFI 0.29.3. This adds Windows ARM64 support, moves the counter to BoltFFI's new C# model, and includes the Windows metadata-path and `BufFromBytes` fixes.

Changes include:

- updating Rust crates, CLI pins, web runtime packages, documentation, lockfiles, and Renovate configuration
- updating the counter FFI and C# integration for the generated `CoreFfi` API
- packaging native libraries for Windows x64 and ARM64
- removing the temporary explicit `crux_core` path workaround
- making `just doctor` verify that the installed BoltFFI CLI is version 0.29.3

Validation:

- all six locked Rust workspaces pass
- counter shared tests pass
- formatting and Justfile formatting pass
- BoltFFI 0.29.3 generates and compiles the C# package successfully
- the NuGet package contains both x64 and ARM64 native libraries
- the generated `BufFromBytes` P/Invoke is present
- doctor rejects BoltFFI 0.27.0 and accepts 0.29.3

Apple packaging was not run locally because this validation was performed on Windows.


mhedgpeth:
@StuartHarris this is the required boltffi upgrade, and also includes the fix for the problem that was reported this morning.

Can you tag me on the PRs that I need to review? I can do that first thing in the morning...

StuartHarris:
Thanks Michael! Looking great. I added a few commits to rectify a few issues Claude found — I've reviewed them and am happy. Please feel free to merge if you agree.

The two substantive ones:

**`&[u8]` at the FFI boundary** (9f1a687) — BoltFFI 0.29.3's C# backend supports slice parameters, so the owned `Vec<u8>` isn't needed any more. I checked by generating both ways and diffing: the output is byte-identical (`public byte[] Update(byte[] data)` either way), so your Windows validation still holds exactly. The owned-buffer requirement was real for 0.25.2 — that's why the `update_bytes`/`resolve_bytes` shims existed, and removing those was the right call. Reverting just the signature drops two `#[allow(clippy::needless_pass_by_value)]`, restores the zero-copy shape, and keeps C#-specific detail out of `part-1/shell.md`, which includes this file verbatim.

**`CoreFFI` → `CoreFfi`** (caed34a) — 0.29.3 tightened C# casing, so one Rust name produced two: Swift/Kotlin/TS echo the Rust spelling, while C# rebuilds it from name parts. Rather than report it upstream, the fix is on our side — `CoreFFI` was never idiomatic Rust (RFC 430: acronyms count as one word, `Uuid` not `UUID`), and the same rule holds for Kotlin and .NET. Renaming the struct makes all four backends agree, verified with `boltffi generate`. The C ABI symbol is unchanged, so it's a pure source-level rename. It also makes the three mermaid diagrams in the book correct, which already said `CoreFfi`.

Plus some smaller ones:

- **770d5b0** — `just build platform=ARM64` doesn't work; just only takes variable overrides *before* the recipe name, so it errors with ``justfile does not contain recipe `platform=ARM64` ``. Now documents `just platform=ARM64 build`.
- **c7e9ca5** — the PowerShell `boltffi --version` check needed the `Select-Object -First 1` equivalent of the bash helper's `head -n 1`, otherwise multi-line output makes `-eq` filter rather than compare.
- **e71dd2f** — the Binaryen ≥123 requirement only lived in CI. I hit the exact failure locally (`Unknown option '--enable-bulk-memory-opt'` on wasm-opt 116), so there's now a version check in `check-web-wasm` alongside `_check-java-version`, and the book says so.
- **394ed04** — the Renovate pin only matched the cargo manager, leaving `@boltffi/runtime` free to drift past the crates in seven `package.json` files.
- **df60ea6** — the version appears in 24 places. `lib.just` is now the single source of truth and `just check-versions` holds the rest to it (runs inside `just check`, so the existing `build` job covers it free).

Tested the Apple and Android counter shells by hand — both good. I also ran counter's web-nextjs CI locally in full, and confirmed all four backends emit `CoreFfi` against their callers.

One gap worth flagging rather than fixing: **the ARM64 app build is never exercised.** CI produces the `win-arm64` native lib via `boltffi pack`, but `just windows/ci` builds with the default `platform=x64`, so the new sln/csproj ARM64 configurations and the `Platform=ARM64` → `RuntimeIdentifier=win-arm64` mapping ship untested. I don't have a Windows machine to check it — happy to leave as-is if you've verified locally, otherwise a build-only ARM64 pass on the x64 runner would be cheap insurance.

StuartHarris (approved):
Amazing! I added a few things (see my comment in the PR conversation). But good to go from my side.

mhedgpeth:
These changes look great Stuart thanks so much I'm going to squash/merge & close.

PR #562, opened 2026-08-04
feat(http)!: split non-rejection errors out of HttpError::Http

StuartHarris:
> **Third of three stacked PRs — merge last.** Based on `http-rejection-headers` (#555),
> which is based on `http-rejection-ergonomics` (#554). Merge #554, then #555, then this.

## Why

`HttpError::Http` is documented as "the exchange completed, but the server answered with a
4xx or 5xx" — the one place an app handles a rejection. It wasn't only that. Two other
things constructed it, and after #555 gave the variant headers, both presented as
fully-formed rejections:

| Site | What it really is | What it claimed |
| --- | --- | --- |
| `Response::body_bytes` on an already-taken body | A caller took the body twice | `Http { code: <the success status>, headers: <that response's>, .. }` |
| `RawResponse::try_from` with a status outside 100–999 | The shell is misbehaving | `Http { code: 999, .. }` |

So `error.code()` could return `Some(200)`, and `matches!(err, HttpError::Http { .. })` was
not a reliable test for "the server rejected this". #554 papered over the first case with an
honest-but-awkward `code()` doc — *"`Some` is not by itself proof of a rejection"* — which is
the sort of caveat that means the type is wrong.

Doing it now rather than later: 0.20 is unreleased and #555 already breaks this variant, so
one release absorbs the lot. Adding a variant is breaking regardless, so waiting means a
second break.

## What changed

Two new variants, and `Http` now means a server rejection and nothing else:

```rust
/// The response body had already been taken.
HttpError::BodyAlreadyTaken,
/// The shell sent a status code that is not valid HTTP (outside 100–999).
HttpError::InvalidStatusCode(u16),
```

`code()` therefore gets its invariant back, and it's worth stating plainly: **`Some` if and
only if the server rejected the request.** Nothing else in the crate carries a status.
`InvalidStatusCode` keeps its number, but not via `code()` — it isn't a status any server
answered with.

`BodyAlreadyTaken` deliberately carries nothing. The caller still holds the `Response` (the
method takes `&mut self`), so the status and headers the old error reported were already in
hand — and they described a *successful* response, which is exactly what made them
misleading.

A side benefit: `Response::body_bytes` goes back to one line. #555 had to restructure it to
avoid cloning the `HeaderMap` on every successful read; with no headers to attach there is
nothing to clone.

```rust
self.body.take().ok_or(HttpError::BodyAlreadyTaken)
```

**`From<http_types::Error>` is removed** (second commit, `http-types` feature only). It
mapped a middleware error onto `Http`, and nothing in the crate consumes it — `compat.rs`
converts `Request` and `Response`, not `Error`. It is not rehomed onto a new variant because
the `http-types` feature is slated for removal in full; `http` is the only HTTP type system
`crux_http` will keep. **`mod compat` and the `pub use http_types` re-export deliberately
stay in this PR** and come out in their own change — they're not an oversight.

## Third commit — the payoff

Making `Response::new` the variant's only constructor retires both of its `Option`s: it
always supplied `Some` headers and `Some` body, so those `Option`s described states that
could not arise.

```rust
-    headers: Option<Box<HeaderMap>>,
-    body: Option<Vec<u8>>,
+    headers: Box<HeaderMap>,
+    body: Vec<u8>,
```

This is what the variant-level `#[non_exhaustive]` from #555 was *for*: nothing outside the
crate can construct the variant, so nobody downstream can observe the change. The accessors
keep their signatures, because the other variants still have neither headers nor a body.

`headers()` gets the better contract out of it. It used to need a paragraph explaining that
`None` meant "no response behind this error" while `Some(&empty_map)` meant "a response that
sent no headers" — a distinction that existed only because non-rejections were sharing the
variant. Now `None` means one thing: not a rejection. `body()` is unchanged, still mapping an
empty body to `None`.

Free at runtime, too: `Box` and `Vec` are both niche-optimised, so `HttpError` stays 72 bytes.
The `Option`s were costing nothing and claiming something false.

## The enum stays exhaustive

I had proposed `#[non_exhaustive]` on `HttpError` and talked myself out of it. `crux_http` is
pre-1.0, so every breaking change is already a minor bump — the attribute would buy the
freedom to add variants without a major, which is freedom we already have. What it would cost
is the compile error an app *wants*: if `HttpError` grows a failure mode, code mapping it
onto its own error type should be made to decide what the user sees, not have it land
silently in a wildcard. And after this split the taxonomy looks complete — transport
(`Url`/`Io`/`Timeout`), rejection (`Http`), decode (`Json`), caller error
(`BodyAlreadyTaken`), malformed shell response (`InvalidStatusCode`) — so it would be
flexibility for growth I can't name. Revisit at 1.0, where it has to be settled anyway.

The variant-level `#[non_exhaustive]` on `Http` from #555 is a different question and stays:
it governs *fields*, where patterns already need `..` so no useful compile error is lost, and
where growth is concrete.

## Migration

```rust
// an exhaustive match gains two arms
match error {
    HttpError::Http { code, .. } => …,
    HttpError::Json(msg) => …,
    HttpError::BodyAlreadyTaken => …,          // new
    HttpError::InvalidStatusCode(code) => …,   // new
    HttpError::Url(_) | HttpError::Io(_) | HttpError::Timeout => …,
}

// code that tested Http to mean "rejected" is now correct rather than nearly correct
if let Some(code) = error.code() { /* the server said no */ }
```

Anything already using a wildcard — including this repo's weather example — is unaffected.

The field-type change needs no migration for the same reason `#[non_exhaustive]` exists: you
cannot construct the variant. The one thing that does break is code that *destructures* the
fields and then unwraps them — `HttpError::Http { body, .. } => body.unwrap()` becomes
`=> body`. Prefer `error.body()`, which is unchanged.

## Tests

- `only_a_rejection_has_a_code` pins the invariant across every variant.
- `body_bytes_returns_error_when_body_already_taken` asserts `BodyAlreadyTaken` **and**
  `code() == None`, with a comment recording that it used to report `Http { code: 200 }`.
- `from_http_response_rejects_out_of_range_status` asserts `InvalidStatusCode(code)` for
  `0`, `99`, `1000`, `u16::MAX`.
- `there_are_no_headers_without_a_response` rewritten: the `code: 999` case isn't an `Http`
  at all any more, which is the point.
- `the_new_variants_display_usefully` pins both `Display` strings.

## CI

Green locally: `cargo fmt --check`; `cargo clippy --workspace --all-targets --all-features
-Dpedantic -Dnursery`; `cargo nextest run --workspace --all-features` (241 tests, 112 in
`crux_http`); `cargo test --doc --all-features` (75 doc tests); `cargo doc` with
`-D rustdoc::broken_intra_doc_links`; `mdbook build` with the linkcheck backend; and
`cargo check` for `--features http-types`, `--no-default-features` and
`--target wasm32-unknown-unknown` — the last three because removing that `From` impl is the
kind of change that only breaks under one feature combination.

**FFI surface verified, not assumed.** Both new variants are `#[serde(skip)]` +
`#[facet(skip)]` like `Http`/`Json`, so shells are untouched. I ran the real generator
(`examples/counter-http/shared --bin codegen --language typescript`) and the emitted type is
still exactly:

```ts
export type HttpError =
    | { kind: "Url"; value: str }
    | { kind: "Io"; value: str }
    | { kind: "Timeout" };
```

`size_of::<HttpError>()` is unchanged at 72 bytes, so `clippy::result_large_err` stays quiet
(the crate has no `allow` for it anywhere).

### DIFF
--- a/crux_http/CHANGELOG.md
+++ b/crux_http/CHANGELOG.md
@@ -130,6 +130,16 @@ and this project adheres to
   **not** `#[non_exhaustive]`, so a new failure mode is a compile error you get to think
   about, rather than something that silently lands in a wildcard.
 
+- **`From<http_types::Error> for HttpError` is gone** (`http-types` feature only). It mapped
+  a middleware error onto `HttpError::Http`, which now means only a server rejection, and it
+  has no callers inside the crate. `http-types` middleware must map its own errors
+  explicitly.
+
+  It isn't rehomed onto a new variant because **the `http-types` feature is slated for
+  removal in full** — `http` is the only HTTP type system `crux_http` will keep. This is the
+  first piece to go; `crux_http::compat` and the `pub use http_types` re-export follow in
+  their own release. If you still depend on the feature, now is the time to say so.
+
 - **`HttpError::Http` has a new `headers` field**, so that a rejection's headers reach the
   app (see `HttpError::header` above). It is `Option<Box<HeaderMap>>`, boxed only because a
   bare `HeaderMap` is 96 bytes and this type is the
--- a/crux_http/CHANGELOG.md
+++ b/crux_http/CHANGELOG.md
@@ -141,10 +141,17 @@ and this project adheres to
   their own release. If you still depend on the feature, now is the time to say so.
 
 - **`HttpError::Http` has a new `headers` field**, so that a rejection's headers reach the
-  app (see `HttpError::header` above). It is `Option<Box<HeaderMap>>`, boxed only because a
-  bare `HeaderMap` is 96 bytes and this type is the
-  `Err` of nearly every function in the crate — inline, it pushed `HttpError` to 160 bytes
-  and tripped `clippy::result_large_err` across the crate.
+  app (see `HttpError::header` above). It is `Box<HeaderMap>`, boxed only because a bare
+  `HeaderMap` is 96 bytes and this type is the `Err` of nearly every function in the crate —
+  inline, it pushed `HttpError` to 160 bytes and tripped `clippy::result_large_err` across
+  the crate.
+
+  Neither `headers` nor `body` is optional, and `body` is now a plain `Vec<u8>` rather than
+  `Option<Vec<u8>>`. `Response::new` is the variant's only constructor and always supplies
+  both, so the `Option`s described states that could not arise. Both types are
+  niche-optimised, so this costs nothing at runtime — `HttpError` stays 72 bytes — and the
+  accessors are unchanged: `body()` still returns `None` for an empty body, and `headers()`
+  still returns `None`, but now that means simply "not a rejection".
 
   `match` arms that name only the fields they use are unaffected:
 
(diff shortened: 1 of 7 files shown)


mhedgpeth:
This looks like a good positive improvement, nice work @StuartHarris

mhedgpeth (approved):

Issue #566, opened 2026-08-07
justfiles in examples aren't runnable on windows


tswast:
I'm on Windows and I would like to use crux for Windows, CLI, and Android development. The examples seem to make some assumptions that we're running on a Mac or Linux, but with a little help from an AI agent I was able to get the Android Counter and Tauri examples running.

mhedgpeth:
Hey there thanks for the submission, I have a way to run this in a polyglot environment and will update the examples accordingly.

mhedgpeth:
I have this on my radar now, going to wait for 0.20 to be released, then will go through the examples and update this to what I have in my local development on Windows.

mhedgpeth:
Correction: I'm waiting for `0.21` I had my version numbers off.

PR #567, opened 2026-08-07
docs: update counter examples for Android, Tauri, and Tui to build on Windows

tswast:
Regarding AI usage, I did generate these with Gemini and Antigravity, but I tested both the Android and Tauri apps manually.

Towards issue #566


StuartHarris (approved):
Amazing! Thank you for this. I don't have a windows machine, but it looks good to me. Thank you for testing it manually.

StuartHarris:
I had Claude do some reviewing and fixing of what it found. We did as much verification locally as we could, but the windows runner in CI will do more validation.

StuartHarris:
Thanks for this — the split into `[unix]`/`[windows]` variants is the right shape, and I'd not have got to the Windows side any time soon otherwise.

I've reviewed it and pushed some commits straight to this branch (hope that's OK — shout if you'd rather I'd left them as suggestions). Four were real bugs:

- **Failure propagation.** The bash recipes run under `set -euo pipefail`, so a failing child aborts the loop. The PowerShell twins had no equivalent — a failing native command neither stops the script nor sets its exit code, so `just ci` on Windows ran every child, printed the failures, and exited 0. Now `$ErrorActionPreference` plus a `$LASTEXITCODE` check after each native call. `doctor` deliberately keeps going on both platforms.
- **`✓`/`✗` would have been mojibake.** This is the one I'd never have caught by reading. PowerShell 5.1 decodes a BOM-less `.ps1` using the legacy ANSI codepage, and `just` writes `[script('powershell')]` bodies as BOM-less UTF-8 — so every check mark and em-dash in the `lib.just` helpers would have garbled on a default install. `powershell.exe` is always 5.1, so there was no way round it via the interpreter. They're now plain recipes under `windows-shell`, matching what `counter/windows/Justfile` already does: `just` passes the recipe as a process argument, so no file decoding happens. If it looked fine on your machine, you may have the UTF-8 beta codepage enabled — that would mask it.
- **`clean` was deeper than the bash version.** `Get-ChildItem -Include target -Recurse` matches at any depth, where `rm -rf */target */*/target` is depth-limited. It deleted a depth-3 `target` bash never touched, and walked every `node_modules` tree looking for one. Also dropped `-ErrorAction SilentlyContinue` from the removal, so locked files and `MAX_PATH` failures surface instead of `clean` claiming success.
- **`_check-java-version` required a dot** after the major version, so JDK GA releases printing `openjdk version "21" 2023-09-19` fell through to "Could not parse".

Plus some naming: `shell` was overloaded (a Crux shell vs the command interpreter, and one Crux shell is literally called `windows`), `shared` was being listed as a shell when it's the core, and the directory is `Android` not `android` — which meant the aggregate recipes could never have driven it on a case-sensitive filesystem.

There's now an `examples-windows.yaml` running `windows-justfiles` and `windows-tauri` on `windows-latest`, path-filtered to `lib.just` and `examples/counter`. Both pass. The justfiles job asserts the two things that can't be checked off Windows: that `doctor` output isn't mangled, and that the aggregate recipes really do fail when a child fails.

### The question — how did your Android build work?

I had to drop the Windows Android job, because `boltffi pack android` can't complete on `windows-latest` with boltffi 0.29.3. The cross-compilation itself succeeds — boltffi reports both "Building Android targets" and "Generating Kotlin bindings" — and then `scan_crate` hands `cargo metadata` a verbatim `\\?\...` manifest path, Cargo resolves `examples/counter/shared`'s workspace to the repo root instead of `examples/counter`, and `crux_core.workspace = true` fails to inherit. Filed as boltffi/boltffi#793; it's the third instance of that root cause after #383 and #781.

Which is why I'm curious: you said you tested the Android app manually. Rust's `canonicalize` always emits the `\\?\` prefix on Windows regardless of path length, so I'd expect `just Android/package` to have failed for you too. Any idea what was different? A few guesses:

- Did you run `just Android/build` (or `dev`/`ci`), or open it in Android Studio after an earlier successful `package`?
- Were you on 0.29.3, or an earlier boltffi?
- Any chance your checkout was somewhere that changed the path handling?

If it *did* work for you, that's genuinely useful information for #793 — it would mean the failure is conditional rather than universal, and I've told them it reproduces flatly. Happy to be corrected there.

Log, if it helps: https://github.com/redbadger/crux/actions/runs/31164459732/job/92822106806

### Still to do

The other four examples' `Android/Justfile`s are byte-identical to counter's old version, so they're still unported — I left them alone to keep this reviewable. `examples/Justfile`'s Windows recipes iterate all six examples, so `just ci` at that level will now fail loudly on Windows rather than silently, which is the right behaviour but means it won't go green until those are done. Happy to take that as a follow-up unless you'd rather roll it in here.

tswast:
> hope that's OK — shout if you'd rather I'd left them as suggestions

Thanks! Yeah, it's fine to write to this branch. I'll try these out!

> The question — how did your Android build work?

I ran `just build` to create the Kotlin bindings and then opened the folder in Android Studio. This was similar to the flow for the Windows folder, where it creates an sln file to open in Visual Studio.

> Were you on 0.29.3, or an earlier boltffi?

I suspect I was on the earlier one. I know I synced to main where it was upgraded, but I think that was after I had tested on Android. I'll try it out again after the latest changes.

tswast:
I just double-checked. Indeed with boltffi 0.29.3, the Android build is failing. I must have been using 0.25.2 before.

mhedgpeth:
I'm planning on working through the examples once 0.20 comes out and will include looking at this as a part of that.

PR #568, opened 2026-08-07
fix(core): don't reuse bridge effect ids, and don't retain fire-and-forget requests

StuartHarris:
Two bugs in the bridge's resolve registry, found while looking at the two
public types called `EffectId`. Both are in the same ~90-line file, and the
second is only fixable once the first is.

Intended for **0.20.0**, merged ahead of the release prep PR (#565) rather than
stacked on it, so it can be reviewed on its own. #565's changelog currently
promises that `crux_core`'s behaviour is unchanged from 0.19.0; that wording
needs correcting when it rebases onto this.

## 1. A resolved request's id was handed to a later request

`EffectId`s were slab indices, and the slab reused an index the moment its
request completed. A shell that resolved an id twice — a retry, a race, a bug —
had its second resolve land on whichever unrelated request had inherited the
slot. If the two outputs deserialized compatibly, which two HTTP requests
generally do, it succeeded **silently**, delivering one request's response to
another.

`tests/json_bridge.rs` demonstrates it. With an app whose HTTP resolution
settles without a follow-up effect, the second request is handed `EffectId(0)`
again, and resolving the first id a second time returns `Ok`:

```
assertion `left != right` failed: a resolved request's id was handed out again
  left: EffectId(0)
 right: EffectId(0)
```

Ids are now issued from an ascending counter, with outstanding requests in a
`HashMap<u32, _>` keyed by that id, so a completed request's id is a lookup miss
rather than a hit on its successor — `ResolveError::NotFound`, the error already
returned for an id that was never issued.

**Why not generational ids?** `effects::registry` already solves this that way:
a 32-bit generation beside a 32-bit index in an opaque `u64`, with
`Storage::take` rejecting a stale one. The bridge can't afford it — its id is a
`u32` on the wire. Splitting those bits (say 22 index, 10 generation) buys only
1024 reuses of a slot before the generation wraps; widening to `u64` changes the
FFI type every shell passes. Spending all 32 bits on identity and detecting
reuse by absence needs no wire change at all.

The counter wraps after `u32::MAX` requests and steps over any id still
outstanding, so a live request can't be displaced even then — unit-tested, since
no integration test can reach that path. The `EffectId overflow` panic is gone
too: ids are `u32` by construction rather than a `usize` slab index narrowed on
the way out.

## 2. Fire-and-forget requests were never released

Every effect was registered so the shell could resolve it later, but a
fire-and-forget request has no continuation, and only a resolve removed an
entry — which for these never comes. So each one held a registry slot for the
life of the process. `Render` is one of them, so that's a leaked entry **per UI
update, forever**.

It's visible in the ids: five `Trigger` events in a row produced render ids
`[0, 1, 2, 3, 4]` under the old slab. The slab always hands out the lowest
vacant index, so ids that never repeat prove nothing was ever freed. An app
rendering ten times a second leaks on the order of a megabyte a day, and marches
toward exhausting the id space.

Such a request now takes an id, so the shell still sees one on the `Request`,
but stores nothing.

### The one observable change

Resolving a fire-and-forget id used to report

> Attempted to resolve a request that is not expected to be resolved.

and now reports

> Request with id 0 not found.

There's no longer an entry to tell those cases apart. `ResolveError::Never`
stays in the enum and is still produced by the `Core` API. Only a shell
resolving an id it was never meant to resolve sees either message.

## Not fixed here

A streaming (`Many`) request keeps its entry after the stream ends, because
nothing tells the registry a stream has finished. That needs `ResolveSerialized`
to report completion — a change to the resolve plumbing rather than to storage
policy — so it belongs in its own PR.

## Verification

- `cargo nextest run --all-features`: 246 tests (242 + 4 new)
- clippy clean across the workspace with `--all-features --all-targets`
- doctests, `cargo fmt --check`
- `cargo nextest run --all-features` in all six example workspaces

Both new integration tests fail on `master` in exactly the way described above,
and the two unit tests cover the id policy including the wrap path.

### DIFF
--- a/crux_core/src/bridge/registry.rs
+++ b/crux_core/src/bridge/registry.rs
@@ -78,7 +78,12 @@ impl<T: FfiFormat> ResolveRegistry<T> {
             let mut outstanding = self.0.lock().expect("Registry Mutex poisoned.");
             let id = outstanding.issue_id();
 
-            outstanding.entries.insert(id, resolve);
+            // A request that cannot be resolved has nothing worth keeping: storing
+            // one would add an entry per fire-and-forget effect — every render, for
+            // the life of the process — that nothing would ever remove.
+            if !matches!(resolve, ResolveSerialized::Never) {
+                outstanding.entries.insert(id, resolve);
+            }
 
             id
         };
@@ -92,8 +97,9 @@ impl<T: FfiFormat> ResolveRegistry<T> {
 
     /// Resume a previously registered effect.
     ///
-    /// Fails with [`ResolveError::NotFound`] if `id` is not outstanding, either
-    /// because it was never issued or because it has already been resolved.
+    /// Fails with [`ResolveError::NotFound`] if `id` is not outstanding —
+    /// because it was never issued, because it has already been resolved, or
+    /// because it belongs to a request that never expected a response.
     ///
     /// # Errors
     ///
--- a/crux_core/src/bridge/registry.rs
+++ b/crux_core/src/bridge/registry.rs
@@ -1,24 +1,59 @@
+use std::collections::HashMap;
 use std::sync::Mutex;
 
 use facet::Facet;
 use serde::{Deserialize, Serialize};
-use slab::Slab;
 
 use super::{BridgeError, FfiFormat, Request};
 use crate::bridge::request_serde::ResolveSerialized;
 use crate::{EffectFFI, ResolveError};
 
+/// Identifies one request across the FFI boundary, for as long as anything
+/// could still refer to it.
+///
+/// Ids are issued in ascending order and are not reused when a request
+/// completes, so an id that has been resolved stays unusable rather than being
+/// handed to some unrelated later request.
 #[allow(clippy::unsafe_derive_deserialize)]
 #[derive(Facet, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(transparent)]
 #[facet(transparent)]
 pub struct EffectId(pub u32);
 
-pub struct ResolveRegistry<T: FfiFormat>(Mutex<Slab<ResolveSerialized<T>>>);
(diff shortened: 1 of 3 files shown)


crux_core/src/bridge/registry.rs
```diff
@@ -1,24 +1,59 @@
...
 pub struct EffectId(pub u32);
 
-pub struct ResolveRegistry<T: FfiFormat>(Mutex<Slab<ResolveSerialized<T>>>);
+pub struct ResolveRegistry<T: FfiFormat>(Mutex<Outstanding<T>>);
+
+/// The requests the shell could still resolve, keyed by the id it was given.
+struct Outstanding<T: FfiFormat> {
+    entries: HashMap<u32, ResolveSerialized<T>>,
+    next_id: u32,
+}
```
charypar, line 29: If the HashMap turned out to be an issue, we could technically have Outstanding be some kind of a growing ring buffer instead, where we remember the minimum id at the tail and subtract it from the requested id.

crux_core/src/bridge/registry.rs
```diff
@@ -31,31 +66,40 @@ impl<T: FfiFormat> ResolveRegistry<T> {
...
-            .lock()
-            .expect("Registry Mutex poisoned.")
-            .insert(resolve);
+        let id = {
+            let mut outstanding = self.0.lock().expect("Registry Mutex poisoned.");
+            let id = outstanding.issue_id();
+
+            // A request that cannot be resolved has nothing worth keeping: storing
+            // one would add an entry per fire-and-forget effect — every render, for
+            // the life of the process — that nothing would ever remove.
```
charypar, line 83: Good catch! All registries should follow that policy

charypar (approved):

crux_core/src/bridge/registry.rs
StuartHarris: _interesting_!

PR #571, opened 2026-08-07
docs: make the root README canonical and share it with crux_core

StuartHarris:
The GitHub landing page and the crates.io landing page for `crux_core` were two
files saying nearly the same thing, and they had drifted **in both directions**.

| | ahead on |
| --- | --- |
| root `README.md` | BoltFFI named as the FFI generator; iOS/Android described as BoltFFI packaging rather than Java Native Access; no stale [Surf](https://crates.io/crates/surf) reference for `crux_http`; the live `part-2/effects.html` book link; all six examples listed; a Talks & Articles section; prose fixes |
| `crux_core/README.md` | the Bridge API — the root still presented `process_event`/`handle_response` as "the core API interface", with a `uuid` rather than an `EffectId` |

`crux_core`'s also carried two dead book links: `guide/effects.html` and
`guide/capability_apis.html`. `docs/src/guide/` now holds only two http pages,
and `capability_apis` no longer exists in the book at all.

## The fix

The root README takes the better half of each, and there is now only **one**
README. `crux_core` points at it from its manifest:

```toml
readme = "../README.md"
```

cargo copies it into the package at publish time. Verified on the real crate,
not just in principle — extracting the built `.crate` gives a regular file:

```
-rw-r--r--  0 0  14760  crux_core-0.20.0/README.md
```

### Why not a symlink

A symlink works for cargo — I tested that too — but not for everyone checking
the repo out. Git for Windows leaves `core.symlinks=false` unless the user has
Developer Mode or admin rights, and materialises the link as a plain file
containing the text `../README.md`.

Nothing in the build reads the README, so Windows CI would have stayed green
(it did, on the symlink version of this branch: `windows` passed in 14m33s)
while Windows contributors saw a twelve-byte stub — and a `cargo publish` from
such a checkout would have shipped that stub to crates.io. The manifest path
has no such failure mode.

## Why they were two files

Link bases. Repo-relative links (`./examples`, `./crux_http`) resolve on GitHub
but not on crates.io, which is why the `crux_core` copy used absolute URLs — and
why the two could never simply be the same bytes.

So every link and image in the shared file is now absolute. Each of the 19
repo-pointing URLs was checked against the working tree; all resolve.

The cost: links point at `master` rather than the branch being viewed, so a PR's
README preview links to master's files. For a page that is mostly prose and
external links, that seems a fair price for making drift impossible rather than
merely discouraged.

## Bonus: a much smaller crate

The README no longer lives inside the crate, so the two SVGs it used move out of
`crux_core`. `architecture.svg` was byte-identical to
`docs/src/architecture.svg` and is simply dropped; `crux_logo.svg` moves to
`docs/src/images/`. Nothing else referenced either.

|  | before | after |
| --- | --- | --- |
| `crux_core` package | 765 KiB | **475 KiB** |
| compressed | 306 KiB | **109 KiB** |

## Verification

- `cargo package -p crux_core` **including** the verify step
- every repo-pointing URL in the README resolves to a path that exists
- `cargo clippy --all-targets -- --no-deps -Dclippy::pedantic -Dclippy::nursery -Dwarnings`
- `cargo fmt --check`, 246 tests

Stacked on #565 so it carries that PR's `crux_core` README work rather than
conflicting with it. Merge after the release.


charypar (approved):
What happens when this is checked out on Windows...? Or have they joined us in the 21st century now?

StuartHarris:
> What happens when this is checked out on Windows...? Or have they joined us in the 21st century now?

Great catch. We changed it to reference the Readme in the directory above. PR Desc updated

PR #572, opened 2026-08-10
docs: clean up RFC docs for the migration off of http-types

fbartho:
This RFC's admonish banner still said "proposal, not yet adopted," and the body still described the design in future/proposal tense throughout ("this document proposes," "we will need," "proposed design"), but the migration it describes has been implemented and shipped.

Issue #285 (migrate from crux_http to standard http crate types) is closed as completed, and #357 (investigate the gradual migration path) is also closed as completed, with the closing comment noting the investigation was well documented. PR #543 implements the RFC in full and merged on 2026-07-06, closing both issues.

This updates the banner and sweeps the body's self-describing status language to past/adopted tense, matching the treatment already given to the other two adopted RFCs in this directory, command.md and typegen.md, which each carry the same "adopted and implemented, kept for future reference" wording after their own migrations shipped. The design rationale and technical content are untouched.

While verifying adoption I diffed the page's technical claims against what actually shipped on master and found seven places where they've drifted. Happy to update the page's content to match the implementation, either in this PR or a follow-up, whichever you'd rather review. I didn't want to do that proactively because I wasn't sure if the RFC was intended as a to be kept as a historical document, or if it should better match what ended up in master.

<details><summary>Specific Documentation-Implementation Drift Instances</summary>

List of content-code divergences:
- the Body backing store is Vec<u8>, not the bytes::Bytes the "Resolved decisions" section names
- mime shipped as an unconditional dependency rather than gated behind the http-types feature
- Request carries an extra middleware field the design section doesn't mention
- the http_types compat conversions live in a separate compat.rs module rather than inline in request.rs/response.rs as described; the response/mod.rs "new_headers() hack" the plan describes replacing isn't in that file
- the design section and file checklist name response_async.rs and ResponseAsync throughout, but the shipped file and type are raw_response.rs and RawResponse
- HttpError::Http shipped non_exhaustive with an added headers field the design never discusses. 

</details>

Evidence:
- https://github.com/redbadger/crux/issues/285#issuecomment-4936245429
- https://github.com/redbadger/crux/issues/357#issuecomment-3941217880
- https://github.com/redbadger/crux/pull/543

### DIFF
--- a/docs/src/rfcs/decouple-from-http-types.md
+++ b/docs/src/rfcs/decouple-from-http-types.md
@@ -1,9 +1,9 @@
 # RFC: Decouple `crux_http` from `http-types`
 
 ```admonish
-This RFC is a **proposal** and has not yet been adopted. We are looking for
-feedback on the overall direction and the open questions, rather than the exact
-public API details.
+This RFC has been **adopted** and implemented, it's kept
+for future reference as additional context for the design
+choices.
 ```
 
 Related issues:
--- a/docs/src/rfcs/decouple-from-http-types.md
+++ b/docs/src/rfcs/decouple-from-http-types.md
@@ -14,13 +14,13 @@ Related issues:
 
 ## Summary
 
-`crux_http` is currently built on top of [`http-types`](https://docs.rs/http-types)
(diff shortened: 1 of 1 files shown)


StuartHarris (approved):
Thanks

Issue #525, opened 2026-05-06
Replace middleware with effect routing by effect type


mhedgpeth:
## Motivation

Middleware lets Crux apps handle effects somewhere other than the normal serialized shell bridge. That is useful, and it showed us that Crux needs an escape hatch for advanced effect handling.

The problem is that middleware does not describe the thing app developers are usually trying to do.

Most apps are not trying to build a stack of effect processors. They are trying to say: "this kind of effect belongs over here."

For example:

- `Render`, `Http`, and `Time` can keep using the normal serialized bridge.
- A camera effect may need a custom typed FFI because the shell returns an opaque platform handle.
- A file store, database, random number generator, or Rust SDK may be better handled inside Rust.
- A project may want to use a different FFI or serialization mechanism, such as `boltffi`, instead of the current UniFFI plus bindgen path.
- When any of those effects resolve, the follow-up effects still need to be handled by the same rules.

Middleware can be stretched to support some of this, but the shape is awkward. Each layer has to decide whether to handle or pass on an effect, composing multiple layers is hard to follow, and resolving an effect can produce more effects that need to re-enter the same handling logic.

We need a replacement for middleware that models effect handling as routing by effect type. The simple bridge should remain the default path, but advanced apps should be able to explicitly route some effects through serialized FFI, some through custom typed FFI, and some to Rust-local handlers.

## Desired outcome

Crux should provide an opt-in effect routing API that lets developers define how each effect type is handled outside the `App` implementation.

The `App` should continue to express what effects it needs. The surrounding core/FFI setup should decide where those effects go.

## Acceptance criteria

- Existing bridge-based apps keep working without adopting the new API.
- Middleware has a clear replacement path.
- Effects can be routed by type.
- The default serialized bridge path can be used as one route.
- Custom typed or opaque FFI routes are possible.
- Rust-local handlers are possible.
- Follow-up effects produced after resolving an effect are routed automatically.
- The approach is demonstrated with an example that mixes serialized shell effects and Rust-local/custom effects.
- Documentation explains when to use the simple bridge and when to use effect routing.

nvowell:
@charypar This is a such a great direction, and I agree spot on with the reasoning expressed in the RFC.

I had some time to play with the prototype — very cool. Dropping the serialized lane entirely as you suggested felt really nice. The extra steps needed per effect seem negligible because we're being so deliberate here anyway.

[...]

PR #573, opened 2026-08-10
docs: mark the effect router RFC as adopted

fbartho:
This RFC's proposal-tense self-description never got an adopted marker, unlike command.md and typegen.md next to it, even though the design shipped.

PR #514 added this RFC and implemented it in the same pull request; the maintainer's adoption call is linked below, and issue #525 tracked the follow-up to completion. crux_core::effects::EffectRouter's own module doc cites this RFC as its design rationale.

This adds the same "adopted and implemented" banner as command.md/typegen.md/#572, and sweeps the body's proposal-tense language to match. Design rationale and technical content are untouched, except two forward-looking sentences with no past-tense form (a feedback request, a middleware-deprecation plan that never happened) — removed rather than reworded.

Comparing the RFC against crux_core::effects on master surfaced drift. Offering to reconcile here or later, maintainers' call.

<details><summary>RFC-vs-implementation drift</summary>

- illustrative code names `Router`/`CameraShell`; shipped type is `EffectRouter`
- RFC's three lanes (Serialized/Opaque typed/Core-local) shipped as four routes: `Serialized`, `Parked` (renamed), `Buffer` (new, not in the RFC), plus core-local handling via a `ResolveSink` trait rather than a named route
- "Next steps" plans middleware deprecation once this lands; `middleware` is still public and undeprecated
- open question flags `HashMap` as the prototype registry; shipped registry uses generational typed IDs (`ParkedEffectId`)

</details>

Evidence:
- https://github.com/redbadger/crux/pull/514#issuecomment-4389313280
- https://github.com/redbadger/crux/issues/525
- https://github.com/redbadger/crux/blob/master/crux_core/src/effects/mod.rs

### DIFF
--- a/docs/src/rfcs/effect-router.md
+++ b/docs/src/rfcs/effect-router.md
@@ -1,14 +1,17 @@
 # RFC: Effect Router and shell-owned effect handling
 
-This RFC proposes a new approach for mixed effect handling, based on routing effects
+```admonish
+This RFC has been **adopted** and implemented, it's kept
+for future reference as additional context for the design
+choices.
+```
+
+This RFC proposed a new approach for mixed effect handling, based on routing effects
 by type to explicit handlers.
 
-The main goal, similar to middleware, is to support advanced cases where some effects
+The main goal, similar to middleware, was to support advanced cases where some effects
 should be handled inside the Core, or where request/result data is not serializable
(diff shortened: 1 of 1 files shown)


StuartHarris (approved):
Thank you!

PR #574, opened 2026-08-10
docs: update cargo-nextest installation command with --locked

RWOverdijk:
Reading the [getting started page](https://redbadger.github.io/crux/part-1/getting_started.html) it says:

> For testing, we also recommend to install [cargo-nextest](https://nexte.st/), the test runner we'll be using in the examples.

When trying to install it I get:

```
error: Nextest does not support being installed without --locked. To install nextest from source, run:

       cargo install --locked cargo-nextest

       For more, see https://nexte.st/docs/installation/from-source/.
```

I assume the file I edited now is the file that creates the page.


StuartHarris (approved):
Thank you!

PR #580, opened 2026-09-04
feat: Encode the request kind to the shell

o0Ignition0o:
This changeset changes the request id encoding so it contains its kind: In the requestID u32, the two first bits now encode the operation kind. The sequence number now wraps around the remaining 30 bits. This will allow shells to eventually know what operations can be resolved, and if so how many times, to prevent runtime routing issues.

This is an early draft as I'm still exploring potential ways to expose a reliable API to the various shells. any input would be welcome though!


StuartHarris:
@o0Ignition0o Thank you for this! I think it's a genius solution. And I love the fact that it's backwards compatible for existing shells. We could merge this and gain immediately. 

My worry is that it doesn't do anything to address what I've always thought is the real problem, which is that you can resolve an effect with the wrong variant. The type is too wide and so we have to check it every time, and the type system isn't helping us.

I don't think I've ever seen a situation that would be hampered by having the request kind on the request/response types (rather than lodged in the instance id).

I raised an RFC with a design I think addresses this. It's breaking obvs, but we're still pre-1.0 so I think I'd rather do that and have a better API.

LMK what you think.

StuartHarris (approved):
Thanks Jeremy, We have this commit in the PR stack already (and then changed it slightly, later on, to encode the Effect Kind and the OperationKind). So I can merge this anyway and then merge the stack on top later.

PR #581, opened 2026-09-05
docs: RFC for per-operation types with static request kinds

StuartHarris:
This RFC proposes that each operation a capability can ask the shell to perform becomes its own type, carrying exactly one output type and exactly one request kind (notify, request or stream), declared statically on the type rather than chosen at the call site.

It grew out of the discussion around #580 and the Zulip thread asking whether anyone sends the same operation variant with two different request kinds. The two pain points it addresses:

- A response can be a well-formed value of the wrong variant, so every capability has to match on the response and decide what to do when it does not correspond to the operation that was sent. `crux_kv` and `crux_time` both panic in that case.
- Nothing ties an operation to how many times it will be resolved. The same variant can be notified in one place and requested in another, so the kind cannot be typegen'd for shells.

The proposal moves both facts into the type system: three marker traits bound the `Command` constructors so the wrong kind stops compiling, the core deserializes into the specific `Output` for each operation, and the kind becomes a static property of each `Effect` variant that type generation can emit for shells with no change to the wire format. A later phase can generate a per-effect shell handler API in which the number of resolves is fixed by the method signature.

The RFC keeps the `RequestKind` type and the `kind()` accessors introduced in #580, and supersedes the encoding of the kind into the request id, since both sides would already know the kind statically. It spells out both orderings depending on which lands first.

Rendered: `docs/src/rfcs/per-operation-types.md`. Also adds the entry to the book index.

Open questions are listed at the end of the document; naming, whether notifications should have `Output = ()`, nested effect enums, and how to report a resolve of a notification once the id no longer carries the kind are the ones most worth early opinions.

### DIFF
--- a/docs/src/rfcs/per-operation-types.md
+++ b/docs/src/rfcs/per-operation-types.md
@@ -1,0 +1,559 @@
+# RFC: Per-operation types with static request kinds
+
+```admonish
+This RFC is a **draft** for discussion. Nothing in it is implemented yet.
+```
+
+This RFC proposes that each operation a capability can ask the shell to perform
+becomes its own type, carrying exactly one output type and exactly one request
+kind (notify, request or stream), declared statically on the type rather than
+chosen at the call site.
+
+## Summary
+
+Today a capability implements `Operation` on a coarse enum, and every variant of
+that enum shares one `Output` type and can be sent with any of the three
+`Command` constructors. Two consequences follow, and Crux users run into both:
+
+1. A response can be a well-formed value of the wrong variant. Every capability
+   has to match on the response and decide what to do when it does not
+   correspond to the operation that was sent. `crux_kv` and `crux_time` both
+   panic in that case.
+2. Nothing ties an operation to the number of times it will be resolved. The
+   same variant can be notified in one place and requested in another, so the
+   kind cannot be typegen'd for shells, and a shell that resolves the wrong
+   number of times either crashes the core or hangs a command forever.
+
+The proposal is to move both facts into the type system:
+
+```rust
+// One type per operation. The kind and the output are properties of the type.
+pub struct Get { pub key: String }
+
+impl RequestOperation for Get {
+    type Output = Result<Option<Vec<u8>>, KeyValueError>;
+}
+
+pub struct Publish(pub Vec<u8>);
+
+impl NotifyOperation for Publish {}
+
+pub struct Subscribe;
+
+impl StreamOperation for Subscribe {
+    type Item = Message;
+}
+```
+
+With the kind on the type:
+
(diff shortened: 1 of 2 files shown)


StuartHarris:
I wanted to see what a (non-breaking) implementation of this might look like. The stacked pull requests referenced here represent a candidate for a set of crate releases (including `facet_generate`, which allows us to plug in the generation of resolve handlers) that allow users' existing shells to work as they do today (and migrate at their own pace to the new implementation). None of this needs to be released, but exists to validate the RFC and provide an example implementation).

charypar (approved):
This sounds really good, especially combined with @o0Ignition0o's idea of encoding the type in the ID – we can probably do somehting along the lines of generating an `EffectKind` enum (and `OperationKind` enum) with numbered variants alongside the actual Effect type and then using some bits of the effect ID to store those to be able to detect mistakes early for the bridged interactions. (Hope that makes some sense). 

Tricky question is choosing how many bits to use for which. I'd say 8 bits for effects (256 operations should be enough for everybody, right? 😅) + 1 for request vs. stream. I think we could reserve athe value of `0` as an id for notifications, because responses are not allowed, so they can all share the same one 🤷🏻‍♂️

StuartHarris:
@charypar that's a great idea. It would add runtime defence in depth and the generated code could do the runtime check of the EffectKind. I think the OperationKind would be redundant in the new architecture, because it can be derived (only useful now for legacy shells). This could mean an extra bit for the id.

StuartHarris:
@charypar building on Jeremy's initial id structuring work and your suggestion, there is an implementation in #589 👀

docs/src/rfcs/per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+only stable way to let undeclared operations coexist with declared ones, and
+the `const { assert!(..) }` does reject the wrong constructor. But the error
+only appears on `cargo build`, never in `cargo check` or the editor, and the
+const and the marker are two declarations that a hand-written impl can make
+disagree. Once the legacy default is gone there is no reason to keep either
+wart, so the breaking release moves the kind to an associated type.
+
+## Open questions
+
+1. **`Output` on notifications.** Is `Output = ()` acceptable, or should the base
```
o0Ignition0o, line 602: I'd love to see `Output = !` at least, because a request *can* resolve with a unit in theory while a notification must never resolve

o0Ignition0o (commented):
Looking great!

docs/src/rfcs/per-operation-types.md
StuartHarris: Yes me too. It's only stable in that position from 1.99 and our MSRV is less than that. But we won't need it when the three traits become first class as Notify will have no associated type at all.

PR #582, opened 2026-09-06
feat(core): static request kinds on operations

StuartHarris:
First PR of the implementation of the per-operation types RFC (#581). Stacks on #580 and is intended to merge after it.

**Note on the base:** #580's head branch lives on a fork, so this PR is opened against `master` and currently shows #580's commit as well. Once #580 merges, this rebases to a single commit. The review surface is the top commit only.

## What this does

- **Keeps** the kind enum and the `kind()` accessors on `RequestHandle`, `ResolveSerialized` and `EffectResolver` from #580, and rewrites its doc to lead with the static declaration. The enum is now called `OperationKind` (it was `RequestKind` in #580 and the first round of this PR): the kind is a property of the operation type, which is the point of the RFC, and `RequestKind::Request` read as a tautology. Renamed in a separate commit on each PR of the stack.
- **Removes** the id-bit encoding from #580. `EffectId`, `issue_id` and `resume` are back to their `master` behaviour (byte-identical apart from using `resolve.kind()` to skip storing notifications). The RFC supersedes the encoding: once the kind is static on the type, both sides already know it.
- **Reverts** #580's `ParkedRequest` change so `Parked::park` keeps returning the tuple. It was the one breaking API change, and the `kind` field is redundant with a static kind.
- **Adds** `crux_core::operation::{Notify, Request, Stream}` marker traits. Import the module, not the items: `impl operation::Request for HttpRequest {}`.
- **Adds** `Operation::KIND: Option<OperationKind> = None`. Every existing impl keeps compiling; `None` means "legacy, kind decided by the constructor called".
- **Adds** a compile-time check in `CommandContext::{notify_shell, request_from_shell, stream_from_shell}`: an operation that declares a kind is rejected by the wrong constructor with an E0080 at the call site. Bounds are unchanged, so legacy operations still work with any constructor. The check fires on `cargo build`, not `cargo check`.
- `RenderOperation` declares `Notify`.

## Tests

- Doctest on the new module showing the hand-written form.
- Typed and legacy operations through the constructors in `command/tests/basic_effects.rs`.
- New trybuild harness (`crux_core/tests/trybuild.rs`, `tests/ui/{pass,fail}`) with three E0080 cases and one E0271 case. Note: trybuild only runs `cargo build` when the suite has at least one `pass` case, and the E0080 errors are post-monomorphisation, so `tests/ui/pass/` must never be empty.

## Changelog

`crux_core` `[Unreleased]` entry replaced: the id-bits, `Never`-restored and `ParkedRequest` sections from #580 are gone; the static-kind feature is described.

Next in the stack: `#[derive(Operation)]`, then per-operation rewrites of `crux_kv`/`crux_time`, then typegen, then examples and docs.


crux_core/src/command/tests/basic_effects.rs
```diff
@@ -140,3 +140,140 @@ fn event_can_be_created() {
...
+    use crate::{Request, RequestKind, capability::Operation, operation};
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+    struct ANotification;
+
+    impl Operation for ANotification {
+        type Output = ();
+        const KIND: Option<RequestKind> = Some(RequestKind::Notify);
+    }
```
charypar: I wonder whether we'd save ourselves a lot of work if we didn't bother with `operation::Notify: Operation`. Do we strictly _need_ the `const KIND`?

crux_core/src/core/resolve.rs
```diff
@@ -14,6 +16,27 @@ pub enum RequestHandle<Out> {
...
+/// [`KIND`](crate::capability::Operation::KIND), this is a static property of
+/// the operation type, and every request carrying that operation has the same
+/// kind. An operation that has not declared one leaves the kind to the call
+/// that created the request — `notify_shell`, `request_from_shell` or
+/// `stream_from_shell` — so one such operation type can be notified in one
+/// place and streamed in another.
+#[allow(clippy::unsafe_derive_deserialize)]
+#[derive(Facet, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[repr(C)]
+pub enum RequestKind {
```
charypar, line 31: `OperationKind`?

crux_core/src/command/context.rs
```diff
@@ -40,12 +40,26 @@ impl<Effect, Event> Clone for CommandContext<Effect, Event> {
...
         Effect: From<Request<Op>>,
     {
+        const {
+            assert!(
+                matches!(Op::KIND, None | Some(RequestKind::Notify)),
+                "this operation does not declare RequestKind::Notify; \
+                 send it with request_from_shell or stream_from_shell instead"
+            );
+        }
+
```
charypar, line 62: This is kinda annoying, but I don't see a good way around it without making the huge breaking change of changing the trait bound to `Op: operation::Notify` 🤔 More importantly, the information about the operation kind would not be easily recoverd for an arbitrary operation type `O` in order to generate an ID later, we do need runtime type information, don't we... 

As much as the `KIND` machinery bothers me, I can't see a way around it.

crux_core/tests/trybuild.rs
charypar, line 1: neat.

charypar (approved):
It feels like there's a future simplification once we remove `Operation` entirely, where the `KIND: Option<OperationKind>` and const checking becomes unnecessary, the trait bounds on the operation-kind-specific APIs get tighter (e.g. `notify_shell` would only take a `Op: Notify`), and the three traits all have an automatic implementation of a `fn kind` or `const KIND` which returns to correct kind for runtime kind introspection. 

But I don't see how to do that without making it an openly breaking change that just won't compile, until all capabilities are updated by splitting their operations and classifying them by implementing one of the three new traits. 🤔

crux_core/tests/trybuild.rs
StuartHarris: It is. But it can come out as soon as we don't need to rely on the `const` check and can use trait bounds instead.

crux_core/src/command/context.rs
StuartHarris: Yeah that will come, and this can go

crux_core/src/core/resolve.rs
StuartHarris: Yes. It is a property of the operation type, which is the whole point of the RFC, and `RequestKind::Request` read as a tautology. Renamed to `OperationKind` everywhere, including the generated shell code (`operationKind`, `effectOperationKind`), in a separate commit on each PR of the stack. `Operation::KIND`, the `operation::{Notify, Request, Stream}` markers, `#[operation(...)]` and the typegen `EffectKind` keep their names.

crux_core/src/command/context.rs
charypar: Yeah, I haven't read the rest of the stack when I posted that. I see the trajectory now

PR #583, opened 2026-09-06
feat(macros): #[derive(Operation)]

StuartHarris:
Second PR of the per-operation types RFC (#581), stacked on #582.

## What this does

Adds `#[derive(Operation)]`, re-exported as `crux_core::macros::Operation`, so a capability declares an operation's output and operation kind once, on the type:

```rust
#[derive(Serialize, Deserialize, Operation)]
#[operation(notify)]
pub struct Publish(pub Vec<u8>);

#[derive(Serialize, Deserialize, Operation)]
#[operation(request, output = Option<Vec<u8>>)]
pub struct Get { pub key: String }

#[derive(Serialize, Deserialize, Operation)]
#[operation(stream, output = Message)]
pub struct Subscribe;
```

The derive emits the `Operation` impl (`type Output`, `const KIND = Some(..)`) and the matching `crux_core::operation::{Notify, Request, Stream}` marker, so the two cannot disagree. Struct-only, any shape, generics pass through. `notify` has no output (`Output = ()`); `request` and `stream` require one, and `output` accepts unquoted generic types. There is no `register(..)` argument: facet-generate walks the output's shape itself, so nested types reach the registry without being named (that was only ever needed by the serde-reflection tracer, removed in #591).

Implemented with darling 0.24 and syn 3, the crate's existing dependencies. Unknown arguments get a "did you mean" suggestion.

## Effect macro and type registry

`#[effect(facet_typegen)]` now records each variant's declared kind: the generated `Export` impl calls the new `TypeRegistry::register_effect_kinds("Effect", &[("Http", <HttpRequest as Operation>::KIND), ..])`, carried through `build()` and exposed as `CodeGenerator::effect_kinds()`. This is the data the typegen plugins in the next PRs consume.

## Tests

- Snapshot tests for the five derive shapes and six error cases (`crux_macros/src/operation/tests.rs`).
- trybuild: a derived notify operation passed to `request_from_shell` fails to compile; all three kinds through their constructors plus an `#[effect(facet_typegen)]` enum compile.
- `crux_core/tests/typegen.rs` asserts the recorded kinds, the type each request resolves with, and that nested output types reach the registry without being named.
- Existing effect macro snapshots updated; the diff is only the added `register_effect_kinds` call.

## Notes

- darling's `Option<T>` does not forward `from_invalid_expr`, so unquoted generic outputs needed a small newtype (`MaybeOutput`).
- `crux_core` cannot use its own re-exported derive, so `RenderOperation` stays hand-written.

Next in the stack: per-operation types for `crux_kv`, `crux_time` and `crux_http`.


crux_core/tests/typegen.rs
```diff
@@ -101,3 +101,122 @@ mod test {
...
+    #[repr(C)]
+    pub enum Event {
+        None,
+    }
+
+    #[derive(Facet)]
+    pub struct ViewModel;
+
+    #[derive(Operation, Facet, Debug, Clone, Serialize, Deserialize)]
+    #[operation(request, output = GetResult, register(StoreError))]
```
charypar: Maybe not the best place to comment, but I can't quite see why explicit `register(StoreError)` is necessary here – it should get later discovered on the `GetResult` type? What am I missing? 🤔

crux_core/tests/ui/pass/derive_all_kinds.rs
```diff
@@ -0,0 +1,63 @@
...
+pub struct Publish(Vec<u8>);
+
+#[derive(Operation, Facet, Debug, Clone, Serialize, Deserialize)]
+#[operation(request, output = GetResult)]
+pub struct Get {
+    key: String,
+}
+
+#[derive(Operation, Facet, Debug, Clone, Serialize, Deserialize)]
+#[operation(stream, output = Vec<u8>)]
```
o0Ignition0o, line 22:
```suggestion
#[operation(stream, item = Vec<u8>)]
```

crux_macros/src/operation/tests.rs
```diff
@@ -0,0 +1,187 @@
...
+    let input = parse_quote! {
+        #[operation(notify)]
+        pub struct Subscribe;
+    };
+
+    let actual = operation_impl(&input);
+
+    insta::assert_snapshot!(pretty_print(&actual), @r"
+    impl ::crux_core::capability::Operation for Subscribe {
+        type Output = ();
```
o0Ignition0o, line 18: I think it would be nice to associate type Output = ! for notifications, wdyt?

o0Ignition0o (commented):
Love it, I think it would allow for much safer shell apis! got a couple of comments wrt naming, although they might belong in the rfc conversation instead

crux_macros/src/operation/tests.rs
charypar: I think that since `Operation` is staying mainly for backwards compatibility, we probably shouldn't change the semantics now. I would later look at removing the Operation super trait altogether instead, at which point `Notify` becomes the only trait, and it doesn't even _have_ an associated output type.

charypar (approved):
The implementation seems good. How fragile is the trait structure as it stands? It looks quite possible for someone to `impl Operation` wrongly on a type, e.g. implement both `Notify` and `Operation` with `KIND: StreamKind`, and I don't love that only the macro really prevents that. 

I think we can probably make that inconsistency impossible when we remove `Operation` later, unless we absolutely need it. The combination of needing them to be wery distinct so they don't get mixed up, but also needing them to share a reflection API is vicious – not sure how we're going to write the trait bound for a function which takes an effect and generates it's ID depending on the kind if we remove `Operation` later. 

Maybe something like a blanket impl of a trait for each of the three subtraits? .e.g `impl<T: Notify> OperationReflection for T` setting the `const KIND = NotifyKind`. And that trait can stay private to the core...? 🤔

crux_core/tests/ui/pass/derive_all_kinds.rs
StuartHarris: Yeah I want to use `item` instead of `output`, both here and in the trait. Thought they should match for now (and it has to be `output` until we have independent traits, which will be in the next breaking release)

StuartHarris:
@charypar spot on. It will come out in the wash with the next breaking release

PR #585, opened 2026-09-06
feat(kv,time,http): per-operation types alongside the enum APIs

StuartHarris:
Third PR in the stack implementing the [per-operation types
RFC](https://github.com/redbadger/crux/pull/581): the two capability crates get
per-operation types and new entry points **beside** their existing enum APIs, and
`crux_http` declares the operation kind it has always had.

Stack: #581 (RFC) → #582 (`OperationKind` on operations) → #591 (remove serde typegen) →
#583 (`#[derive(Operation)]`) → **this PR** → typegen plugins → examples, docs and
deprecations.

**Nothing is deprecated here.** `KeyValue`, `KeyValueOperation`, `KeyValueResult`,
`Time`, `TimeRequest` and `TimeResponse` are untouched, and both APIs work side by side.
The deprecations land in #588, once `notes` and `weather` have migrated in #587 — the
examples build with `-D warnings` against path deps, so deprecating earlier would break
their CI.

## Why

`Operation` says one operation has one output. The convention of implementing it on an
enum breaks that: every `KeyValueOperation` variant shares one `KeyValueResult`, so a
`Get` can be answered with a `KeyValueResponse::ListKeys` as far as the type system and
the deserializer are concerned, and the capability has to check at runtime — and decide,
as `crux_kv` and `crux_time` both did, to panic.

## Naming: same names, new modules

The new capabilities keep the names `KeyValue` and `Time` and live in a module each:
`crux_kv::store::KeyValue` and `crux_time::clock::Time`. The root types of the same
name are the enum APIs, deprecated in #588. The two coexist so an app can migrate one
call at a time, and the next breaking release deletes the root types and re-exports
the module ones in their place — so the import you write today keeps working, and
`crux_kv::KeyValue` comes to mean the per-operation type. (Earlier revisions of this PR
called them `KeyValueStore` and `Clock`; the rename follows review.)

## crux_kv

```rust
// before
Command::request_from_shell(KeyValueOperation::Get { key })
    .map(KeyValueResult::unwrap_get) // panics if the shell answered something else

// after
use crux_kv::{operation, store::KeyValue};

#[effect]
enum Effect {
    Get(operation::Get),
    Set(operation::Set),
    Render(RenderOperation),
}

KeyValue::get("key").then_send(Event::Loaded)
```

`store::KeyValue` has the same five methods as the root `KeyValue`, with the same
signatures and the same `DataResult` / `StatusResult` / `ListResult` return types. Its
bounds are per method (`where Effect: From<Request<operation::Get>>`), so an app's
`Effect` only carries the operations it actually uses.

| Operation | Fields | Output | Kind |
| --- | --- | --- | --- |
| `Get` | `key: String` | `ValueResult` | `Request` |
| `Set` | `key: String, value: Vec<u8>` | `ValueResult` | `Request` |
| `Delete` | `key: String` | `ValueResult` | `Request` |
| `Exists` | `key: String` | `BoolResult` | `Request` |
| `ListKeys` | `prefix: String, cursor: u64` | `KeysResult` | `Request` |

```rust
pub enum ValueResult { Ok(Value), Err(KeyValueError) }
pub enum BoolResult  { Ok(bool),  Err(KeyValueError) }
pub enum KeysResult  { Ok(Keys),  Err(KeyValueError) }

pub struct Keys { pub keys: Vec<String>, pub next_cursor: u64 }
```

Concrete `Ok`/`Err` enums rather than `Result`, which does not cross the FFI boundary.
Each converts to and from the `Result` alias an app sees, in both directions, so a shell
that already speaks one API can serve the other.

## crux_time

```rust
// before
let TimeResponse::Cleared { id } =
    ctx.request_from_shell(TimeRequest::Clear { id }).await
else {
    panic!("Unexpected response to TimeRequest::Clear");
};

// after
use crux_time::{clock::Time, operation};

#[effect]
enum Effect {
    NotifyAfter(operation::NotifyAfter),
    Clear(operation::Clear),
}

let (timer, handle) = Time::notify_after(Duration::from_secs(1));
```

`clock::Time` mirrors the root `Time`'s three methods and shares `TimerHandle`,
`CompletedTimerHandle`, `TimerOutcome`, `TimerId`, `Instant` and `Duration` with it.

| Operation | Fields | Output | Kind |
| --- | --- | --- | --- |
| `Now` | — | `Instant` | `Request` |
| `NotifyAt` | `id: TimerId, instant: Instant` | `TimerId` | `Request` |
| `NotifyAfter` | `id: TimerId, duration: Duration` | `TimerId` | `Request` |
| `Clear` | `id: TimerId` | `TimerId` | `Request` |

Every operation is answered with the bare `TimerId` it was given, which the core still
asserts against the timer it started. That includes `Clear`, so clearing behaves exactly
as it does with the enum API: `TimerHandle::clear` sends an `operation::Clear` request
and the timer's future resolves with `TimerOutcome::Cleared` once the shell has
answered. (An earlier revision made `Clear` a notification; review pointed out that
the acknowledgement is a guarantee worth keeping, that nothing on the shell will fire
for that timer after the app sees `Cleared`.)

A timer that fires after the core has cleared it is harmless: the core has stopped
waiting for that request and ignores the late answer, with no event and no error.
There is a regression test for that, because an earlier draft of the migration guide
claimed the opposite.

## crux_http

`HttpRequest` gains `const KIND: Option<OperationKind> = Some(OperationKind::Request)` and
`impl crux_core::operation::Request`, hand-written alongside its existing
`register_types`. No API change and nothing changes on the wire; sending an
`HttpRequest` with `notify_shell` or `stream_from_shell` is now a compile error.

## Tests

- `crux_kv/src/tests/legacy.rs` is the old `tests.rs`, unchanged; `crux_kv/src/tests/store.rs`
  covers the same scenarios through `store::KeyValue`, plus the `From` round-trips, the
  `Set` `Debug` truncation and the JSON shapes of the operations and outputs.
- `crux_time/tests/clock.rs` covers `now`, `notify_at`/`notify_after` resolved with their
  `TimerId`, clearing (a `Clear` request answered with its id, then the `Cleared`
  outcome), a late answer to a cleared timer being ignored, a dropped handle, and the
  JSON shapes of all four operations.
- `crux_core/tests/effect_router_prototype.rs` is migrated from the enum `Time` API to
  `clock::Time` with `TimeNotifyAfter(operation::NotifyAfter)` and
  `TimeClear(operation::Clear)` variants, as a demonstration of an `Effect` enum naming
  third-party operations.

Verified with `cargo fmt --check`, clippy pedantic + nursery `-D warnings` over all
targets and features, `cargo build --workspace`, `cargo nextest run --all-features`,
`cargo test --doc --all-features`, `cargo doc`, and `cargo clippy -p shared --all-targets`
in `examples/notes` and `examples/weather` (plus `web-leptos`, which also depends on
`crux_time`).


charypar (commented):
Nice to see what capabilities look like with the new operations types.

I have a general question about the direction of travel: How does this work? Do we ask people to migrate to `KeyValueStore` and `Clock` and deprecate `KeyValue` and `Time`? If so, should we deprecate them now?

Or do we only have these for testing, and we later switch the original capabilities to the new protocol as a breaking change and remove `Clock` and `KeyValueStore`?

StuartHarris:
Good question, and it changed the stack. The new capabilities now keep the names `KeyValue` and `Time` and live in a module each: `crux_kv::store::KeyValue` and `crux_time::clock::Time`. The root types stay as the enum API, deprecated in #588, and the next breaking release deletes them and re-exports the module types in their place, so code written against the module paths never changes.

The changes are in a separate commit on each PR (this one, #587, #588, #589, #590) so they are easy to read on their own.

charypar (approved):

PR #586, opened 2026-09-06
feat(typegen): emit request kinds and an effect handler API for shells

StuartHarris:
Fourth PR of the per-operation types RFC (#581), stacked on #585. Depends on redbadger/facet-generate#123.

## What this does

Facet type generation now emits, next to the generated `Effect` type in Swift, Kotlin, TypeScript and C#:

- an `OperationKind` type and a per-variant accessor (`effect.operationKind`, `effectOperationKind(effect)` in TypeScript because the union's discriminant is already `kind`), `nil`/`null`/`undefined` for legacy operations that do not declare a kind yet;
- an `EffectHandler` protocol/interface with one method per variant: notifications return nothing, requests return the `Output` (`async` / `suspend` / `Promise` / `Task`), streams take an `EffectSink<Output>`, legacy variants take `(operation, requestId, resolve)`;
- an `EffectDispatcher(handler, resolve)` that calls the right method and resolves the request id never, once, or once per sink item, serializing each output with the generated bincode serializers. `resolve` is the shell's own `(requestId, bytes)` callback around the core FFI.

A shell that implements the handler cannot resolve the wrong number of times or with the wrong type, because there is no `resolve` call for it to get wrong. Shells that prefer a flat match keep working unchanged; the emission is purely additive.

```swift
public protocol EffectHandler: Sendable {
    func render(_ operation: RenderOperation)
    func http(_ operation: HttpRequest) async -> HttpResult
    func subscribe(_ operation: Subscribe, into sink: EffectSink<Message>)
}
```

```ts
export interface EffectHandler {
    render(operation: RenderOperation): void;
    http(operation: HttpRequest): Promise<HttpResult>;
    subscribe(operation: Subscribe, sink: EffectSink<Message>): void;
}
```

## How

- `crux_core/src/type_generation/facet.rs` becomes `facet/{mod,effects}.rs` plus `facet/plugins/{request_kind,handler}/{swift,kotlin,typescript,csharp}.rs`. The plugins implement facet-generate's public `EmitterPlugin` and act in `after_type` for the registered effect only. Nothing Crux-specific is in facet-generate.
- `#[effect(facet_typegen)]` now records each variant through `TypeRegistry::register_effect::<EffectFfi>()?.variant::<Op>("Http")?...finish()`, capturing `Op::KIND` and the `Format` of `Op::Output`; this replaces PR2's interim `register_effect_kinds`.
- `build()` rejects user types named `OperationKind`, `EffectSink`, `EffectHandler`, `EffectDispatcher` (and the C# `I` forms); `CodeGenerator::without_effect_handlers()` opts out of the handler API.

## Dependency

`facet_generate` is pinned to the git revision of facet-generate#123 (`c949b1cc25c2`) so CI can resolve it before 0.21.0 is published. A `TODO(release)` in `Cargo.toml` marks the switch to `facet_generate = "0.21"`. Root and example lockfiles updated.

## Notes

- C#'s `Effect` record is not `partial`, so the accessor is emitted inside the record via `type_body`; the enum still comes from `after_type`.
- The Swift handler and dispatcher carry `@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)` because facet-generate's `Package.swift` declares no platforms and `Task {}` needs them. Packages that declare `platforms:` (WeatherKit does) conform without extra annotation.
- `OperationKind` and the handler API are emitted for the first registered effect only; every effect gets its own kind accessor.

## Tests

Eight per-language plugin snapshots, four real `CodeGenerator` runs into tempdirs (the TypeScript one runs pnpm and tsc), reserved-name and opt-out tests, updated effect macro snapshots and typegen tests. Generated code compiled locally: `swift build` on weather's regenerated Apple package, `tsc` via `just web-nextjs/typegen` for weather and notes, `kotlinc` over weather's regenerated Android types, and `dotnet build` on the counter and fixture C#. Not attempted: xcodebuild, Gradle, WinUI.

Next in the stack: migrate the notes and weather examples to per-operation types, docs, and deprecations.


charypar (approved):
Very cool! 

The one question I have is whether we need to emit the request kind to the shell, given we generate the EffectDispatcher and the handler protocol already makes it impossible to mishandle the effect. How are you thinking it would be used?

PR #587, opened 2026-09-06
feat(examples): move notes and weather to per-operation types

StuartHarris:
Fifth PR of the per-operation types RFC (#581), stacked on #586. Migrates the two pilot examples end to end so readers of the RFC can see real code on both sides of the boundary. Docs, deprecations and version bumps follow in the next PR. The other four examples are untouched and keep using the enum APIs, which proves the two coexist.

## Effect enums

- **notes**: `Time, Render, PubSub, KeyValue` → `Render, Publish, Subscribe, KvGet, KvSet, TimeNotifyAfter, TimeClear`
- **weather**: `Render, KeyValue, Http, Location, Secret, Time` → `Render, Http, KvGet, KvSet, TimeNotifyAfter, TimeClear, IsLocationEnabled, GetLocation, FetchSecret, StoreSecret, DeleteSecret`

Both cores use `crux_kv::store::KeyValue` and `crux_time::clock::Time`, the per-operation entry points from #585, which keep the old names and live in a module each.

Each enum lists only the operations the app uses. Weather's Secret capability is the RFC's best before/after: its three hand-narrowed response types are now simply the outputs of three operations, and the six-variant `SecretResponse` and every `unreachable!()` arm are gone. Location's two operations return a bare `Bool` and `Location?` with no wrapper types.

## Shells

| Shell | Approach |
|---|---|
| notes web-nextjs | Generated `EffectHandler` + `EffectDispatcher`. Nested `matchEffect`/`matchPubSubOperation`/`matchTimeRequest`/`matchKeyValueOperation`, the hand-built response constructors, `unsupported()` and the `subscriptionId` ref are all gone; `page.tsx` holds an `EffectSink<Message>`. |
| weather apple | Generated handler, `nonisolated extension Core: EffectHandler` in a new `handler.swift`; each capability file becomes its handler methods. |
| weather Android | Generated handler; `Core : EffectHandler` delegating to the injected handlers, whose `when` blocks are deleted. One coroutine per request, since `dispatch` is `suspend` and a timer suspends for its whole duration. |
| weather web-leptos | Flat match over the eleven variants in the typed Rust lane; per-operation outputs make each arm's type exact. Adds a `Timeout` registry so `TimeClear` really cancels. |
| weather web-nextjs | Generated handler and dispatcher. |

Every mdBook anchor in the touched files is preserved (the set is byte-identical before and after) and `mdbook build` succeeds. `examples/weather/ARCHITECTURE.md` is updated.

## Verified locally

- Root: fmt, clippy pedantic and nursery with warnings as errors, 304 tests.
- notes: `just shared/ci`, `just web-nextjs/ci`.
- weather: `just shared/ci`, `just web-leptos/ci`, `just web-nextjs/ci`; `swift build` and `xcodebuild -scheme WeatherApp-macOS` succeed with swiftlint clean; Android `assembleDebug`, `lintDebug` and `testDebugUnitTest` succeed.
- Not run end to end: `just apple/ci` and `just Android/ci`, because both begin with `boltffi pack` and this machine lacks swiftformat and the `x86_64-linux-android` target. The equivalent build, lint and test steps ran against freshly regenerated types. CI covers the rest.

## Findings for follow-up

- **Swift `EffectHandler: Sendable` cannot be witnessed by `@MainActor` methods** because facet-generate does not emit `Sendable` on generated types. The shell uses a `nonisolated` extension and hops to the main actor for results. Candidate facet-generate follow-up.
- **A cleared timer can still fire, harmlessly.** Every shell cancels its pending timer on `TimeClear` and answers the clear with the `TimerId`. A timer that fires before the clear reaches the shell answers `TimeNotifyAfter` late, and the core ignores that: the request it belonged to has already been dropped. An earlier revision of this PR believed a late answer panicked and built the shells around suppressing it; that was wrong, and #585 now has a regression test for it.
- Generated `Set` collides with the standard library `Set` in Swift, Kotlin and TypeScript, so shells alias it.


charypar (commented):
Looking at the example shells, I wonder if this reveals a slight weakness of the typegen in that the handler is a single big protocol/interface, rather than a per-capability protocol. (Rather than `EffectHandler`, it would be `HttpEffectHandler` and so on). 

It prvents us from providing drop-in implementations of the official capabilities for the various supported languages in a composable way – users would have to do the wiring from the handler to the provided implementation. 🤔

charypar (approved):
Thinking about the composition some more, I can't actually think of a way to do it, which doesn't require wiring it up anyway. But his is probably the best we can do.

PR #588, opened 2026-09-06
docs: per-operation types guide, deprecate the enum APIs, bump for 0.21

StuartHarris:
Sixth and last PR of the per-operation types RFC stack, on top of #587. Docs,
deprecations, one macro fix, and the version bumps for the release.

Stack: #581 (RFC) → #582 (`OperationKind` on operations) → #583
(`#[derive(Operation)]`) → #585 (`crux_kv` / `crux_time` / `crux_http`) → #586
(typegen, needs redbadger/facet-generate#123) → #587 (examples) → **this PR**.

**The RFC file** (`docs/src/rfcs/per-operation-types.md`) is updated in this PR
to describe what the compat stage actually contains. #581 has merged, so this
is now a diff against master rather than a copy of the file.

## Deprecations

Nothing is removed and nothing changes on the wire. Each item warns and names
its replacement. The replacements keep the old names in a module — `store::KeyValue`,
`clock::Time` — and the breaking release re-exports them at the crate roots once the
enum types are gone; the guide and the RFC's Migration section both say so.

| Item | Since | Use instead |
| --- | --- | --- |
| `crux_kv::KeyValue` | 0.15.0 | `crux_kv::store::KeyValue` |
| `crux_kv::KeyValueOperation` | 0.15.0 | `crux_kv::operation::{Get, Set, Delete, Exists, ListKeys}` |
| `crux_kv::KeyValueResult` | 0.15.0 | `crux_kv::operation::{ValueResult, BoolResult, KeysResult}` |
| `crux_kv::KeyValueResponse` | 0.15.0 | the output type of the operation you sent |
| `crux_time::Time` | 0.19.0 | `crux_time::clock::Time` |
| `crux_time::TimeRequest` | 0.19.0 | `crux_time::operation::{Now, NotifyAt, NotifyAfter, Clear}` |
| `crux_time::TimeResponse` | 0.19.0 | `Instant` for `Now`, `TimerId` for the rest |
| `crux_time::TimerFuture` | 0.19.0 | nothing — an implementation detail of `Time` |

Not deprecated, because both APIs share them: `KeyValueError`, `Value`,
`DataResult`, `StatusResult`, `ListResult`, `TimerHandle`,
`CompletedTimerHandle`, `TimerOutcome`, `TimerId`, `Instant`, `Duration`.

The two `protocol` modules carry a module-level `#![allow(deprecated)]` — a
lint attribute on the item does not reach the code its own derives generate —
and the legacy impls, the `command.rs` re-export shims and the legacy test
files carry `#[allow(deprecated)]`. Root CI runs `-D warnings` and is clean.
Only `notes` and `weather` depend on these crates, and neither uses the enum
APIs any more; the other four examples never did.

## Versions

| Crate | From | To |
| --- | --- | --- |
| `crux_macros` | 0.10.1 | 0.11.0 |
| `crux_core` | 0.20.0 | 0.21.0 |
| `crux_http` | 0.20.0 | 0.21.0 |
| `crux_kv` | 0.14.0 | 0.15.0 |
| `crux_time` | 0.18.0 | 0.19.0 |

Inter-crate requirements moved with them (`crux_core` → `crux_macros` 0.11,
each capability crate → `crux_core` 0.21). Every `[Unreleased]` section became a
dated release-plz-style section for 2026-09-06, keeping the prose the earlier
PRs wrote, plus ⚠️ Deprecated sections in `crux_kv` and `crux_time` and a note
in `crux_core` that facet type generation now needs `facet_generate` 0.21.

Lockfiles were updated with `cargo update --workspace` and a targeted
`cargo update -p crux_*` per example, rather than `just update`, which pulled
in ~1100 lines of unrelated dependency churn.

The `facet_generate` git pin in the root `Cargo.toml` is unchanged and its
`TODO(release)` comment is in place. `docs/STABLE_REF` is deliberately left at
`crux_core-v0.20.0` — it moves after publishing.

## Docs

New:

- **`docs/src/guide/migrate-per-operation-types.md`** — modelled on the
  `crux_http` migration guide. Quick checklist; declaring an operation in all
  three forms (and by hand); replacing a response enum with
  weather's Secret before/after; `crux_kv` and `crux_time` before/after with the
  operation tables; the `Effect` naming convention and the
  `is_` / `into_` / `expect_*` helper renames; regenerating shells and adopting
  the handler API per language with before/after from notes' `core.ts` and
  weather's `core.swift`; keeping a flat match for Rust shells; the traps from
  #587 (a cleared timer that fires anyway, and why that is harmless; the Swift
  `Sendable` / `nonisolated` pattern for `@MainActor` cores; `Set` collisions); the deprecations table;
  and "Coming in the breaking release".
- **`docs/src/rfcs/per-operation-types.md`** — still marked **proposed**; the
  admonish now points at this stack and the migration guide as the reference
  implementation. The "Type generation" section is rewritten to the API the
  plugins actually emit in all four languages (taken from the snapshots),
  including the Swift `@available` and C# non-`partial` record details.
  "Migration" describes what the compat stage contains, and all six open
  questions are answered — `Output = ()` kept, no enum-splitting derive, no
  nested effect enums, `Never` for a resolved notification (structured ids,
  #589), stream termination still separate, `Ok`/`Err` outputs as convention.

Rewritten or edited:

- `part-2/capabilities.md` — new "One output per operation" section, the Secret
  narrowing/`unreachable!()` section deleted, a new notifications-and-streams
  section including notes' `pub_sub.rs` through two new minimal anchors.
- `part-2/effects.md` — eleven variants, the kind on `Operation`, the three
  builders as the three kinds, `AnOperation::One/Two` → two operation types.
- `part-2/shell.md` — new "How many times to resolve" section introducing
  `operationKind`, `EffectHandler`, `EffectSink` and `EffectDispatcher`.
- `part-2/shell/ios.md`, `android.md`, `react.md` — the handler API instead of
  the match, plus the actor-isolation and name-collision notes.
- `part-2/shell/leptos.md` — why a Rust shell keeps a flat match, and the
  cleared-timer warning.
- `part-4/typegen.md` — new "Operation kinds and the effect handler API" section
  with per-language snippets, reserved names, `without_effect_handlers()` and
  the `facet_generate` 0.21 requirement.
- `part-1/shell.md`, `part-3/middleware.md`, `part-4/runtime.md` — small edits.
- `SUMMARY.md` — the guide under Guides, the RFC under RFCs.

`mdbook build` is clean, every `{{#include}}` and anchor resolves, and the
weather/notes anchor set is unchanged apart from the two added to `pub_sub.rs`.

## Verified locally

Root: `cargo fmt --all --check`; clippy pedantic + nursery `-D warnings` over
all targets and features; `cargo build --workspace`; `cargo nextest run
--all-features` (304 passed); `cargo test --doc --all-features`; `cargo doc
--all --no-deps` (two pre-existing `AppTester` intra-doc warnings, untouched
file); `just check-versions`. Examples: `just shared/check` in `notes` and
`weather`, and `just web-leptos/check` in `weather`. Docs: `cd docs && mdbook
build`.

## Release checklist (owner)

1. Merge redbadger/facet-generate#123 and publish `facet_generate` 0.21.0.
2. In root `Cargo.toml`, replace the git pin with
   `facet_generate = "0.21"` (the `TODO(release)` comment marks the line), then
   `just update` and re-run `just ci`.
3. `just publish` — it goes in dependency order: `crux_macros`, `crux_core`,
   `crux_http`, `crux_kv`, `crux_time`.
4. Bump `docs/STABLE_REF` to `crux_core-v0.21.0` in a follow-up PR.


docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+
+Otherwise, in this order:
+
+1. **Your own capabilities** — one struct per operation with
+   `#[derive(Operation)]`, and one output type per operation instead of a shared
+   response enum.
+2. **`crux_kv` and `crux_time`** — swap `KeyValue` for `KeyValueStore` and `Time`
+   for `Clock`, and list the operations you use in your `Effect` enum.
+3. **Your `Effect` enum** — one variant per operation, which renames the
+   generated `is_` / `into_` / `expect_*` test helpers.
```
charypar, line 32: Thank god for AI, this would be the most painful part of the migration 😬

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+4. **Regenerate your shells** and either adopt the generated `EffectHandler`, or
+   widen the match you already have.
```
charypar, line 34: I would recommend adopting the handler, and just mention that if that's impossible for whatever reason, there is a backup option.

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+
+### By hand
+
+If you'd rather not use the derive, declare the kind and the marker together:
+
+```rust,ignore
+impl Operation for Get {
+    type Output = ValueResult;
+    const KIND: Option<RequestKind> = Some(RequestKind::Request);
+}
```
charypar, line 101: When we make a decidion on this, worth adding that this is temporary and `type Output` will move to the `Request` trait.

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+
+The convention the examples follow:
+
+- **Third-party operations** get a capability prefix: `KvGet`, `KvSet`,
+  `TimeNotifyAfter`, `TimeClear`. Without it, `Get` and `Set` on their own read
+  as if they belonged to the app, and `Set` collides with a standard library type
+  in three of the four shell languages.
+- **Your own operations** are verb-first and unprefixed: `Publish`, `Subscribe`,
+  `IsLocationEnabled`, `GetLocation`, `FetchSecret`, `StoreSecret`,
+  `DeleteSecret`.
```
charypar, line 335: IMO it's wise to follow the 3rd-party convention even for local capabilities.

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+- **List only the operations the app uses.** Bounds are per operation, so an app
+  that never lists keys is never asked to serve `ListKeys` — and neither is its
+  shell.
```
charypar, line 338: This is quite a neat side-effect of the refactoring 🙂

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+      // … an arm per variant the app never issues
+    }),
+    // … Time, likewise
+  });
+}
+```
+
+```typescript
+// After — one method per operation, returning its output
+export class Core implements EffectHandler {
```
charypar, line 399: arguably, this should be called `Shell`, no?

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+
+These are the things that actually caught us out migrating the two examples.
+
+### Never resolve a cleared timer
+
+`operation::Clear` is a notification, so by the time the shell sees it the core
+has already stopped waiting for the timer. Resolving the original
+`NotifyAfter` request afterwards is a `NotFound`, which the FFI surfaces as a
+panic. Each shell has to drop or cancel the pending timer rather than let it
+fire:
```
charypar: This sounds like a race condition waiting to happen. Core clears a timer, emits the notification, but before the shell gets to it, it fires the notification anyway. I don't think that should panic, it should just be a no-op.

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+
+- Swift invalidates the `Timer`, which releases the continuation
+  `timeNotifyAfter` is suspended on;
+- Leptos keeps a `HashMap<usize, Timeout>` and removes the entry, since dropping
+  a `gloo_timers::Timeout` cancels it;
+- TypeScript calls `window.clearTimeout` on the stored handle;
+- Kotlin cancels the coroutine running the delay.
+
+Under `Time`, by contrast, the shell answered a `Clear` with
+`TimeResponse::Cleared`. Delete that response.
```
charypar: I can't quite remember, but I think this was intentional and had something to do with the above race condition

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+}
+```
+
+URLSession, Keychain and CoreLocation work does not belong on the main actor
+anyway, so this is usually an improvement. Only `Sendable` values cross back.
+
+### `Set` collides with the standard library
+
+`crux_kv`'s `Set` operation generates a type called `Set` in every language, and
+Swift, Kotlin and TypeScript all have one already. Alias it at the import:
```
charypar, line 587: This feels like foot gun we should really work harder to avoid in the first place 😬

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+```typescript
+import type { Set as SetValue } from "shared_types/app";
+```
+
+### `Clock`'s cleared outcome no longer waits for an ack
+
+`Time::notify_after`'s future resolved `TimerOutcome::Cleared` only once the
+shell had acknowledged the clear. `Clock`'s resolves immediately. If any of your
+core logic relied on the round trip to sequence something after a clear, it now
+happens sooner.
```
charypar: As above – this was intentional, at least in part to avoid the panic on late timer firing, but I think I also needed it once to have a definite confirmation that the timer has now been cancelled on the shell end.

docs/src/guide/migrate-per-operation-types.md
```diff
@@ -0,0 +1,638 @@
...
+## Coming in the breaking release
+
+Written against this release's derive and marker traits, your code does not
+change. What changes:
+
+- **The kind becomes an associated type.** `type Kind: operation::Kind` replaces
+  `const KIND`, with sealed unit types `operation::kind::{Notify, Request,
+  Stream}`, and the markers `operation::{Notify, Request, Stream}` become blanket
+  impls from it. The derive emits `type Kind = kind::Request;` instead of a const
+  plus a marker impl, so only hand-written `impl Operation` blocks need editing.
```
charypar, line 633: Ha, we arrived at the same idea. Ignore my comment on one of the earlier PRs then

charypar (commented):
Sorry, a good half of this feedback really belongs to the respective lower down PRs. 😓

docs/src/guide/migrate-per-operation-types.md
StuartHarris: You were right on both counts, and both are fixed. `Clear` is a request again, answered with its `TimerId`, and the cleared outcome waits for that answer as `Time` always did.

Separately, the guide was wrong: a shell that fires a cleared timer late does not panic. The core drops the request when the timer is cleared and ignores the answer, which I verified through the bridge and pinned with a test in #585 (`answering_a_cleared_timer_late_is_ignored`). The shells no longer contort themselves to suppress the answer. This section of the guide is rewritten in the new fix commit on this PR.

docs/src/guide/migrate-per-operation-types.md
StuartHarris: Not sure. It's not really core or shell. Core is what callers are calling, and what people are used to. But it does now own the loop. So *maybe*?

docs/src/guide/migrate-per-operation-types.md
StuartHarris: reverted from notification to request

mhedgpeth (approved):
This looks great, I'm looking forward to these changes

docs/src/guide/migrate-per-operation-types.md
StuartHarris: Taken, in `b74ef5263` up in #605. The checklist step now says to adopt the handler API and let the generated `Core` drive the loop, with implementing just the handler or widening an existing match named afterwards as the fallback if that can't work; the longer section says outright that the hand-written match "is the fallback, not the recommendation".

docs/src/guide/migrate-per-operation-types.md
StuartHarris: Decided far enough to write up, with one question left for you. In the breaking release `Operation` keeps `type Output` exactly as today and gains `type Kind: operation::Kind`; the kind traits carry the payload under the right word — `operation::Request` gets `type Response`, `operation::Stream` gets `type Item`, and `operation::Notify` declares none — each bounded `Operation<Kind = kind::X, Output = <Self as X>::Response>` (or `::Item`, or `()`), so the two cannot disagree. Because `Output` stays put, `Request<Op>`, the registries, middleware, typegen and the `#[effect]` helpers do not change, and neither does any third-party middleware. The write-up is a new RFC, [One trait per operation kind](https://github.com/redbadger/crux/blob/per-op/14-opaque-request-ids/docs/src/rfcs/operation-kind-traits.md), which supersedes the breaking stage of #581's RFC and reproduces the shape you reviewed there before replacing it; the guide's By hand section now says the `type Output` + `const KIND` pair is transitional and points at it. All on #609 at the top of the stack. It also settles `item =` rather than `output =` for streams (#583) and the `Output = !` question (#581, #583) — `Notify` has no payload type to give a value to.

The open question is yours from #583: you asked for the inconsistency to be impossible rather than checked. The RFC compares two ways of binding the kind traits, with rustc transcripts for each. *Blanket*: the kind traits are blanket-implemented from `Kind` and read the payload off `Operation` — one impl per operation, disagreement unrepresentable, no orphan. *Tied*: an author (or the derive) writes both impls and the supertrait bound rejects a mismatch as E0271. The decisive difference is the wrong constructor: with tied bounds it is an unimplemented trait, so `#[diagnostic::on_unimplemented]` fires and you read "`Get` is not a notification — a request is sent with request_from_shell, a stream with stream_from_shell"; with blanket impls it is an E0271 associated-type mismatch, the attribute does not fire, and you get the compiler's walk through the blanket impl's bounds, which does at least point at your `type Kind` line. The RFC leans tied for that reason, since derived operations cannot disagree or be orphaned either way and the diagnostic is what a first-time author hits. Do you weigh impossibility above the diagnostic? That is the one thing I'd like your answer on before implementing.

PR #589, opened 2026-09-07
feat(bridge): structured request ids

StuartHarris:
Stacks on #588. Part of the per-operation types work (RFC #581).

#582 dropped #580's two-bit kind-in-id encoding. This puts structure back into the request id, but a different structure, now that the kind is static per operation and shells already read it from the generated table.

## Id layout

`EffectId` is still a `u32`, packed most significant first:

| bits | meaning |
|---|---|
| 31–24 | effect variant index (declaration order in the `Effect` enum) |
| 23 | operation kind: set for a stream, clear for a request |
| 22–0 | sequence, starting at 1, wrapping within its own bits and stepping over anything still outstanding |

Id `0` is reserved: every notification is issued that one id and nothing is stored for it. Sequences never reach 0, so 0 can only ever be a notification. The pieces are readable through `EffectId::{effect_index, kind, sequence}`; the layout is internal to the bridge, and a shell resolves with the id exactly as it arrived, never taking it apart.

The variant index comes from `EffectFFI`, which gains a defaulted `variant_index()` method, a `VARIANT_COUNT` constant and a `variant_name(index)` function. `#[effect]` overrides all three and rejects an enum with more than 256 variants. A hand-written `EffectFFI` impl keeps compiling, reports index 0, and its errors number variants instead of naming them. The kind bit comes from the handle at registration, so legacy operations with no static `KIND` get it too.

## Bridge checks

The registry keys outstanding requests by sequence and stores the issued id alongside the resolve, which is what lets a mangled id be reported as mangled rather than as unknown. Before any deserialization, `resume` rejects:

- id 0 with `ResolveError::Never` (a shell resolved a notification)
- an effect index outside the enum with `ResolveError::NoSuchEffect`
- a sequence issued for another effect or another kind with `ResolveError::WrongEffect` / `ResolveError::WrongKind`

The errors name what they reject rather than numbering it (review round 2, in the `fix(bridge)` commit): ids print in hex so the three fields can be read off them, `WrongEffect` carries both sides as `EffectVariant { index, name }`, and `NoSuchEffect` names the effect type.

```text
Request id 0x01000001 names `Render` (variant 1), but request 1 was issued for `Http` (variant 0).
Request id 0x09000001 names variant 9 of `shared::Effect`, which has only 2 variants.
```

`NotFound` now means only never-issued or already-resolved. `ResolveError` is marked `#[non_exhaustive]` (called out in the CHANGELOG; happy to drop that if reviewers prefer).

## Type generation

The id is opaque to the shell. Nothing is generated to take it apart: a request is resolved with the id exactly as it arrived, and the plugins this stack emits — `OperationKind` with its per-effect accessor, and the handler API — read the kind from the effect's metadata, never from the id. The mapping from an id back to its effect lives in Rust, in the bridge's errors above, next to the layout it depends on.

An earlier revision of this PR also emitted an `EffectKind` enum and a `RequestId` decoder for Swift, Kotlin, TypeScript and C#. #609 removes them, on the review comment below, before anything ships: nothing consumed them, and they pinned the bit layout into four languages.

## Docs

- RFC: rewrote the "supersedes the id encoding" paragraph and "the serialized lane and the wire" with the actual layout and why it is not #580's; answered open question 4 (resolving a notification now reports `Never`) and re-checked 5 (the kind bit changes nothing for stream termination).
- `part-4/typegen.md`: a "Reading a request id" subsection, which #609 rewrites as "Request ids" to describe the id as opaque; #609 also restores the pre-existing claim that `without_effect_handlers()` keeps the kind accessor, by making the code do so.
- `part-2/shell.md`: a paragraph on the id with a link.
- CHANGELOG entries inside the existing `crux_core` 0.21.0 and `crux_macros` 0.11.0 sections.

## Verification

| Command | Result |
|---|---|
| `just check` | pass |
| `just ci` (328 tests incl. trybuild) | pass |
| `cargo doc --all --no-deps --all-features` with `-D warnings` | pass |
| `cd docs && mdbook build` | pass |
| notes: `just shared/ci`, `just web-nextjs/ci` | pass |
| weather: `just shared/ci`, `just web-leptos/ci`, `just apple/ci`, `just Android/ci` | pass |
| counter: `just windows/ci` | **not run** (recipe refuses to run off Windows) |

For C#, the counter's generated code was compiled ad hoc with `dotnet build` and produced no errors in the generated file. The real Windows shell CI job will be the first full run.

## Wire change to note

All notifications now share id 0, so a batch of two renders carries two identical ids. Nothing in Crux keys off a notification id, but a shell that did will see the change.


StuartHarris:
@charypar one benefit from having notifications always use id `0` is that only requests and streams now need allocating and even though we only have 23 bits they will go a lot further probably (before wrapping)

crux_core/src/bridge/registry.rs
```diff
@@ -11,48 +11,253 @@ use crate::{EffectFFI, RequestKind, ResolveError};
...
+
+    /// How many times the shell is expected to resolve this request.
+    ///
+    /// This is the kind the request was registered with, which for an
+    /// operation that declares [`Operation::KIND`](crate::capability::Operation::KIND)
+    /// is the kind that operation always has, and otherwise the one the call
+    /// site chose.
+    #[must_use]
+    pub const fn kind(self) -> RequestKind {
+        if self.0 == Self::NOTIFICATION.0 {
```
charypar, line 88: would `if self == Self::NOTIFICATION {` not work? Asking more out of curiosity

crux_core/src/core/resolve.rs
```diff
@@ -84,11 +84,40 @@ impl<Output> RequestHandle<Output> {
...
+    /// effect variant.
+    #[error(
+        "Request id {id} names effect variant {actual}, but request {sequence} was issued for effect variant {expected}."
+    )]
+    WrongEffect {
+        id: u32,
+        sequence: u32,
+        expected: u8,
+        actual: u8,
+    },
```
charypar, line 142: Any way we could actually name the types? The numeric error message will be really painful to map back 🤔

charypar: Something something effect derive macro provides a tranalation helper... or something like that

charypar (commented):
Nice! Good bit of added safety.

IMO we should leave the id opaque to the shells (in line with the suggestion on the other PR that we don't even emit the EffectKind type). Saves us generating some code, and keeps our ability to change our mind about the format. It's just an internal error detection mechanism which is entirely private.

crux_core/src/core/resolve.rs
StuartHarris: Done, in the `fix(bridge)` commit on this PR. `EffectFFI` gains a defaulted `variant_name(index) -> Option<&'static str>` which `#[effect]` implements from the names it already has, the registry keeps it alongside `VARIANT_COUNT`, and the errors use it:

```text
Request id 0x01000001 names `Render` (variant 1), but request 1 was issued for `Http` (variant 0).
Request id 0x09000001 names variant 9 of `shared::Effect`, which has only 2 variants.
```

Ids print in hex so the three fields can be read off them. A hand-written `EffectFFI` without names falls back to "variant 1".

crux_core/src/core/resolve.rs
```diff
@@ -95,24 +121,29 @@ pub enum ResolveError {
...
+        id: u32,
+        index: u8,
+        variants: u16,
+        effect: &'static str,
+    },
     /// The id's sequence is outstanding, but it was issued for a different
     /// effect variant.
     #[error(
-        "Request id {id} names effect variant {actual}, but request {sequence} was issued for effect variant {expected}."
+        "Request id {id:#010x} names {actual}, but request {sequence} was issued for {expected}."
```
charypar, line 135:
```
"Request id {id:#010x} expects {expected}, but response {sequence} carries {actual}."
```

charypar (approved):
👏🏻

StuartHarris:
@charypar you were right about this, and the thread deserved an answer. #609 removes the `EffectKind` enum and the `RequestId` decoder from all four languages. Before removing them I checked what read them: nothing. Not the generated `Core`, `EffectDispatcher` or `EffectHandler`, not the shipped shell handlers, not an example, not a test beyond the plugin's own snapshots. The dispatcher switches on the effect variant and passes the id through as the integer it arrived as, so all the decoder achieved was pinning the bridge's bit layout into Swift, Kotlin, TypeScript and C#.

The crash-report case it was written for is carried by the bridge's errors, which since this PR name the effect: ``Request id 0x01000001 names `Render` (variant 1), but request 1 was issued for `Http` (variant 0).`` That mapping lives in Rust next to the layout it reads, and the shell sees an opaque integer it hands back. None of it has shipped, so it costs nobody anything now, and we keep the freedom to change the format later.

`OperationKind` and the per-effect `operationKind` accessor stay: they are static per operation and come from the effect metadata, not the id, and a shell resolving by hand needs them to know how many times to resolve. Along the way #609 also fixes the gating so `without_effect_handlers()` no longer takes them away with the handler API — the pre-existing docs claim that this PR "corrected" turned out to describe the right behaviour, so the code now matches it.

crux_core/src/core/resolve.rs
StuartHarris: Done in #609, following your expects/carries shape and expected-before-actual order. The three now render as:

- `shared::Effect` has only 2 variants, but response id 0x09000001 carries variant 9.
- Request 1 expects `Http` (variant 0), but response id 0x01000001 carries `Render` (variant 1).
- Request 1 expects the Request kind, but response id 0x00800001 carries the Stream kind.

One deliberate departure from the string as written: I swapped which noun takes which verb. The id in `{id}` is the one that came back from the shell, so it is the thing that *carries* `{actual}`; the outstanding request identified by `{sequence}` is the thing that *expects* `{expected}`. Taken literally the suggestion had them the other way round, and named a response by a sequence number where a response is identified by its id. If you would rather have yours exactly as written, it renders as "Request id 0x01000001 expects `Http` (variant 0), but response 1 carries `Render` (variant 1)." and I will switch to it.

crux_core/src/bridge/registry.rs
StuartHarris: It doesn't compile, because `kind()` is a `const fn` and `==` on `EffectId` goes through the derived `PartialEq::eq`, which isn't const — rustc rejects it with E0015 ("cannot call non-const operator in constant functions"). Comparing the wrapped `u32`s is the same test expressed in terms the const evaluator accepts. `matches!(self, Self::NOTIFICATION)` would also work here, since the type derives `PartialEq + Eq` and so is structural-match, but the field comparison reads more like the bit tests on the other two branches.

PR #590, opened 2026-09-07
feat(typegen): generate a shell-side Core

StuartHarris:
Stacked on #589 (structured request ids). Part of the per-operation types stack #584; RFC #581.

## What this does

With `EffectHandler` and `EffectDispatcher` doing the resolving, what every shell still wrote by hand was the same loop: serialize the `Event`, call the FFI `update`, deserialize the `Requests`, re-read the view on `Render`, dispatch the rest, and on each resolve call the FFI `resolve` and loop over what comes back. Weather's Swift shell spent about 140 of its 171 glue lines on it; every TypeScript shell carried a byte-identical hand-rolled `deserializeRequests`.

Type generation now emits that loop as a `Core` class in Swift, Kotlin, TypeScript and C#, next to the handler API:

- **`CoreBridge`** (`ICoreBridge` in C#): three byte-level methods, `update`, `resolve`, `view`. The shell implements it around BoltFFI's `CoreFfi` in a few lines, or with a fake for previews and tests. Typegen never imports BoltFFI's output.
- **`Core`**: `update(event)`, `process(requests)`, `process(bytes)` (for middleware callbacks), and the current `view`. It intercepts the variant carrying `crux_core::render::RenderOperation`, re-reads the view, holds it and notifies through an `onView` callback (a `StateFlow` in Kotlin). Holding the view is deliberate: it is where difficient-style diffs would be applied later without moving the app-facing surface.
- **`EffectHandler.render` gains a default no-op** in all four languages, so app handlers no longer implement it.

No `Render` variant, no `Core`. `CodeGenerator::without_core()` turns it off and leaves the handler API; `without_effect_handlers()` turns off both. `Core`, `CoreBridge`, `ICoreBridge` join the reserved names. `TypeRegistry::register_app` now records the app's `Event` and `ViewModel` as `AppMeta` (`CodeGenerator::app()`), and `EffectVariantMeta` gains `render: bool`.

## Examples

The notes (web) and weather (Apple, Android, web) shells adopt it. In each, the old hand-written `Core` splits into a `LiveBridge` adapter, a `WeatherHandler`/`NotesHandler`, and the generated `Core`. Counter stays hand-rolled because Part I of the book teaches the manual loop. Two things worth a look:

- Swift: the generated `Core` carries the same `@available` as the dispatcher, so it can't be `@Observable`; weather keeps a tiny `ViewStore` that `onView` writes into. `LiveBridge` is `nonisolated` and `@unchecked Sendable` (the target defaults to `MainActor` isolation; `CoreFfi` is a class over a mutex-guarded `Bridge`).
- Android: the generated `build.gradle.kts` declares `kotlinx-coroutines-core`, but weather pulls the generated sources in via `srcDirs`, so the `shared` module now declares it as an `api` dependency (version in `libs.versions.toml`).

## Docs

New proposed RFC `docs/src/rfcs/generated-core.md`. Part II shell chapters (iOS, Android, React), the lifecycle chapter's include, the typegen chapter (new "The generated Core" section), the migration guide and the CHANGELOG are updated. The per-operation types RFC's admonish and status sentences are put back into proposal tense; that text was introduced in #588, so tell me if you'd rather I squash that hunk down there.

## Verification

- `cargo test -p crux_core --features facet_typegen`, `just check`, `just ci` (341 tests), `RUSTFLAGS="-D warnings" cargo doc --all --no-deps`. 8 insta snapshots changed on purpose: 4 new `core_*`, 4 `handler_*` gaining the defaulted `render`.
- Regenerated output compiles: `swift build` on the generated package, `kotlinc` against `kotlinx-coroutines-core`, `tsc` for both web shells, `dotnet build` for counter's C# output (the typegen integration test also compiles TS).
- Shell CI locally: `just web-nextjs/ci` (notes, weather), `just apple/ci` (weather), `just Android/ci` (weather, `lintDebug` + `assembleDebug` + unit tests). Weather web was also driven end to end under Node: initial view read without an `onView` call, `Start` produced the secret and kv requests, each `Render` fired `onView`.
- `cd docs && mdbook build` clean; no unresolved `{{#include}}` in the output.
- Not clean: `swiftformat --lint` in `WeatherKit`'s own `just ci`, with swiftformat 0.63.0, fails on 10 files that all fail identically on the parent commit; none are files this PR adds or edits in the flagged lines. GitHub CI doesn't run swiftformat.


crux_core/src/type_generation/facet/plugins/snapshots/crux_core__type_generation__facet__plugins__tests__core_swift.snap
```diff
@@ -0,0 +1,60 @@
...
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+@MainActor
+public final class Core {
+    /// The view model as of the last `Render`.
+    public private(set) var view: ViewModel
+    private let bridge: any CoreBridge
+    private let onView: @MainActor (ViewModel) -> Void
+    private var dispatcher: EffectDispatcher!
+
+    public init(bridge: any CoreBridge, handler: any EffectHandler, onView: @escaping @MainActor (ViewModel) -> Void) {
```
charypar, line 27: Rather than a callback for view, I think the `view` property should be `@Observable` (or the SwiftUI equivalent I can't quite remember)

charypar (commented):
I've read this quite quickly as I'm not super familiar with all the languages, but two things that stood out are

* I think we should work a little harder on making the view property of the core integrate nicely with the foreing language's (or even UI framwork's) expectations on observability. A callback is probably good in TS, but not that idiomatic in Swift. It will also be generally insufficient if/when we switch to sending patches over the FFI rather than full view models
* Is there a technical limitation forcing the examples to implement the bridging to the generated FFI by hand? I was hoping this could all be fully automatic, and the shell would just get a `Core` class it can instantiate, which expects an effect hanlder to be provided and exposes a `viewModel`.

StuartHarris:
Thanks Viktor. Two answers.

**Observability.** Agreed, and done in #594 stacked on this one: the Swift `Core` is now `@Observable` (with `view` its one observed property), C# gets `INotifyPropertyChanged`, Kotlin keeps `StateFlow` and TypeScript keeps the callback, which is what a React state setter is. Only `Core` moves to iOS 17 / macOS 14; the handler API and dispatcher stay at iOS 13, so an older shell still has them. Weather's `ViewStore` is gone; views read `@Environment(Core.self)`.

On patches: `@Observable` on a class whose `view` is a struct invalidates exactly what a callback would, every view that reads `view`. Finer-grained updates need the view model *types* themselves to be observable, which is a difficient-time design question. What this stack does for it is keep the previous value inside generated code, which is what applying a patch needs. I've rewritten the RFC's open question 1 to say that.

**Automatic bridging.** No hard limitation, but two facts. The shell side is only configuration: typegen doesn't know BoltFFI's module and class names (`Shared`/`CoreFfi` in Swift, from `boltffi.toml`; `Data` vs `[UInt8]`; a package dependency the generated `App` package would need on `Shared`), and the middleware examples have a different FFI shape. I'd like to add an optional typegen config naming the FFI module and class, from which typegen emits a `Core(handler:)` convenience that wraps `CoreFfi`, keeping `CoreBridge` for fakes and middleware. Follow-up PR, if you're happy with that shape.

The Rust side is the real blocker. I checked BoltFFI 0.30.1 with fixtures: it still discovers exports by scanning source, so a `crux_core` macro emitting `#[boltffi::export]` is silently invisible. 0.30 does pull exports from dependency crates into the root bindings, but only path dependencies, so `crux_core` from crates.io can't host `CoreFfi`. The fix is BoltFFI's per-invocation metadata capture (boltffi/boltffi#665, approved, integration branch stalled since July, not in 0.30). When that ships, `crux_core::export_ffi!(App)` becomes possible and the whole thing is automatic.

charypar:
> I'd like to add an optional typegen config naming the FFI module and class, from which typegen emits a Core(handler:) convenience that wraps CoreFfi, keeping CoreBridge for fakes and middleware. Follow-up PR, if you're happy with that shape.

That sounds good! Ideally a Swift engineer looking at this should just see a `class` that wants an effect delegate (we could name it that, it's a very common pattern in Cocoa) and they can almost be totally oblivious to the fact there's Rust happening in there.

StuartHarris:
Done in #597, stacked on #594.

The codegen is told where BoltFFI put its output and type generation emits the bridge: `boltffi(BoltFfi::new().swift("Shared").kotlin().typescript("shared", ..))`, each language opted in on its own. For a named language the generated module gains an `FfiBridge` and `Core` gains a constructor that takes only the handler — `Core(handler: WeatherHandler())` in Swift, `Core(handler, scope)` in Kotlin, `await Core.create(handler, setView)` in TypeScript (the wasm module loads asynchronously, and this awaits it so no shell casts the untyped `initialized` promise any more), `new Core(handler)` in C#. Weather's and notes' `LiveBridge` files are gone; `CoreBridge` stays for the preview fake and for the middleware examples, whose `CoreFfi::new` takes a callback.

Two things it costs, both only when the option is set: the generated Swift package depends on BoltFFI's, so it needs a `platforms:` floor (a new `Config` setting), and the generated TypeScript package depends on the wasm package, so `boltffi pack wasm` runs before typegen. Both needed small additions to facet-generate (redbadger/facet-generate#124), including the companion-file hook the shipped-handlers RFC wanted anyway.

On "effect delegate": I've kept `EffectHandler` and `Core(handler:)`. The name is in the merged per-operation types RFC and across four languages in this stack, and what a Swift reader sees is the property you described — a class that takes one collaborator, with nothing about Rust in view — under either name. Happy to revisit in the breaking release if you feel strongly.

charypar (approved):
Ticking off this one, given https://github.com/redbadger/crux/pull/597 addresses the feedback

PR #591, opened 2026-09-08
feat(core,macros)!: remove the serde-generate type generation

StuartHarris:
Removes the serde-reflection / serde-generate type generation backend, leaving `facet_typegen` as the only one.

Stacked on #582; #583 now bases on this. Part of the per-operation types stack (#584).

## Why here

The `#[derive(Operation)]` PR (#583) originally added `register(A, B, ..)` to name types the serde tracer couldn't reach through an output — nested enums, anything behind a `Vec` or `Option`. facet-generate doesn't have that problem: `add_type` walks the shape statically and finds nested containers itself. So `register(..)` existed purely to keep a backend alive that the docs stopped describing in 0.19 ("the facet-based typegen remains the supported path"), that no example uses, and that the [typegen RFC](https://redbadger.github.io/crux/rfcs/typegen.html) lists retiring as its final step. Rather than ship the derive with a serde-only knob in 0.21 and remove it later, this removes the backend below the derive, and the PRs above no longer carry `register(..)` at all. A regression test in `crux_core/tests/typegen.rs` checks that a nested output type reaches the registry without being named.

## What goes

- `crux_core`: the `typegen` feature; `type_generation::serde` (693 lines) and the `crux_core::typegen` alias; `Operation::register_types`; the bundled `typegen_extensions/` runtimes; `serde-generate`, `serde-reflection` and `include_dir` as direct dependencies.
- `crux_macros`: the `Export` derive; `#[effect(typegen)]`, which is now a compile error naming `#[effect(facet_typegen)]`; the `typegen` feature.
- `crux_http`, `crux_kv`, `crux_time`: the `typegen` feature and their serde `register_types` impls.
- `doctest_support`, `renovate.json`: the matching feature and the `typegen-npm` group.

The facet path is untouched in API and naming (`register_types_facet` keeps its name; a rename is breaking for hand-written impls and belongs in the breaking release).

## Breaking

Only for crates on the serde `typegen` feature. Each affected crate's CHANGELOG has a 💥 entry with the migration: switch to `facet_typegen`, derive `Facet` on bridge types, `#[effect(facet_typegen)]`, generate via `TypeRegistry`/`CodeGenerator` per [the book](https://redbadger.github.io/crux/part-4/typegen.html).

## Verification

`just ci` green (all features, trybuild `ui` cases, insta, doctests); `cargo tree -i serde-generate` finds nothing; `cargo check --all-features` in `examples/counter/shared` and `examples/notes/shared`.


charypar (approved):
RIP

PR #594, opened 2026-09-15
feat(typegen): make the generated Swift Core observable

StuartHarris:
Stacked on #590 (generated shell-side `Core`). Part of the per-operation types stack #592; RFC #581.

## What this does

Viktor's review of #590 asked for the Swift `view` to be `@Observable` rather than published through a callback. This PR does that, and gives C# its equivalent.

- **Swift**: the generated `Core` is `@Observable`. `view` is its one observed property; `bridge` and `dispatcher` are `@ObservationIgnored`. The `onView` initializer parameter is gone: `Core(bridge:handler:)`. Only `Core` is raised to `@available(macOS 14.0, iOS 17.0, tvOS 17.0, watchOS 10.0, *)`; `CoreBridge`, `EffectHandler`, `EffectDispatcher` and the rest of the module keep their iOS 13 / macOS 10.15 bar, so a shell with an older deployment target still has the handler API and can drive the loop itself. The generated file now imports `Observation` (only when a `Core` is emitted).
- **C#**: `Core` implements `INotifyPropertyChanged` and raises `PropertyChanged` when `View` is replaced; the `Action<ViewModel>` constructor parameter is gone.
- **Kotlin** (`StateFlow`) and **TypeScript** (`onView` callback, which is exactly a React state setter) are unchanged.

## Examples

The weather Apple shell deletes `ViewStore`. `WeatherApp` keeps the `Core` in `@State` and injects it with `.environment(core)`; views read `@Environment(Core.self)`. Previews build a genuine `Core` over a `PreviewBridge` (fixed view model, no requests) and a `PreviewHandler` whose methods never run, in a new `PreviewCore.swift`. The book's `start` anchor is unchanged.

## Docs

RFC `generated-core.md`: publication paragraph, Swift concurrency note, non-goal on UI frameworks, a new drawback (iOS 17 floor for `Core`), migration step 3, and open question 1 is replaced by the real remaining question, whether the view model *types* should become observable once diffs arrive. The BoltFFI non-goal now records what 0.30 does and doesn't change. The iOS shell chapter, typegen chapter, migration guide and CHANGELOG follow.

## Verification

- `cargo test -p crux_core --features facet_typegen`, `just check`, `just ci` (333 tests), `RUSTFLAGS="-D warnings" cargo doc --all --no-deps`. Two snapshots changed on purpose: `core_swift`, `core_csharp`. New test: `observation_is_imported_only_when_a_core_is_emitted`.
- Regenerated output compiles: `swift build` on the generated package and on `WeatherKit`; `dotnet build` of counter's generated C# project (no new warnings; the new `_view` field raises no CS8618).
- `cd examples/weather && just apple/ci`: `BUILD SUCCEEDED`, swiftlint clean on real sources. `WeatherKit`'s own `just ci` still fails at `swiftformat --lint` on the same 10 pre-existing files as the parent; `PreviewCore.swift` is formatted and adds none.
- `cd docs && mdbook build` clean; no unresolved includes.


crux_core/src/type_generation/facet/plugins/snapshots/crux_core__type_generation__facet__plugins__tests__core_swift.snap
```diff
@@ -15,18 +15,20 @@ public protocol CoreBridge: Sendable {
 /// requests that come back, and resolves them through the bridge.
 ///
 /// `Render` is handled here, so an `EffectHandler` need not implement it.
-@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+///
+/// `view` is observable, so a SwiftUI view that reads it is invalidated
+/// whenever the core renders.
+@available(macOS 14.0, iOS 17.0, tvOS 17.0, watchOS 10.0, *)
+@Observable
```
charypar, line 22: I trust the robot it knows what it's doing, but if I'm not mistaken, it's using the macro version of `@Observable` rather than the property... deco..rator (?) on the `view` field which I think is more widely supported? My Swift is a bit rusty tbh.

More importantly – we should ask people for feedback on the generated bindings when they test the RC

charypar (approved):

PR #596, opened 2026-09-16
RFC: shell handlers shipped with capabilities

StuartHarris:
Proposes that a capability crate can ship the shell side of its protocol — Swift, Kotlin, TypeScript and C# implementations of its effect handlers — as source embedded in the crate, which type generation emits into the app's generated package when the app asks for it.

The rules for turning a platform response into an operation's output belong to whoever defined the protocol: which `URLError` is a timeout, what a cleared timer answers with, that a late answer after a clear is harmless. Today every shell re-derives them from the book. Which HTTP client performs the request is a different question, and that one stays with the shell. So this is a reference implementation a shell can adopt, not a default it inherits.

Stacked on #597 because it builds on the per-operation types and generated `Core` RFCs, and because the companion-file hook it uses arrived with the BoltFFI bridge. The implementation follows in #604 (the mechanism) and #605 (the three capabilities and the migrated examples).

### The shape

```rust
// crux_http/src/lib.rs
pub static HTTP: ShellHandler = ShellHandler::new("Http")
    .types(register_types)
    .swift(ShellSource::stdlib(include_str!("../shell/swift/Http.swift")))
    .kotlin(ShellSource::stdlib(include_str!("../shell/kotlin/Http.kt")))
    .typescript(ShellSource::stdlib(include_str!("../shell/typescript/http.ts")))
    .csharp(ShellSource::stdlib(include_str!("../shell/csharp/Http.cs")));
```

```rust
// the app's codegen binary
registry.shell_handler(&crux_http::HTTP)?.shell_handler(&crux_time::TIME)?;
let typegen = registry.build()?;
```

```swift
// the app's handler
let http = URLSessionHttpHandler.shared
func http(_ operation: HttpRequest) async -> HttpResult { await http.request(operation) }
```

Type generation copies the source into the module — a companion file in Swift, Kotlin and C#, appended after the types in TypeScript — and generates nothing else. `EffectHandler`, `EffectDispatcher` and `Core` are untouched, and nothing generated calls the shipped code. Registering a handler also registers every operation and output its sources name, because the shipped file implements the whole capability even where the app's `Effect` carries only part of it.

### Two revisions since the first draft

Both came out of @charypar's review, and both earlier shapes are written up under Alternatives so the next reader doesn't retread them.

1. **The default is gone.** The first draft had `ShellSource` carry a `default` expression and each operation name its shipped method in `#[operation(..)]`, so an app that wrote nothing got the shipped implementation. That made the core decide what the shell does, emitted source into shells that never asked, and made a capability's property required on one platform and optional on another.
2. **The generated adapter is gone too.** The second draft kept opt-in wiring: a per-capability `HttpHandling: EffectHandler` protocol supplying the method bodies. Counted out on the weather shell it saved almost nothing — five delegating lines became three conformances and three properties — while costing a method listing per capability, `TypeId` plumbing and four adapter emitters. The delegating lines turn out to be the better behaviour anyway: when a capability gains an operation, the compiler asks for a line instead of a regenerated file wiring it silently.

What survives from both is the one-declaration configuration they were there to provide, now as ordinary code the shell already knows how to write.

### Registration is deliberate

It is the one line an app has to know to write. It stays because type generation has to discover the static somehow, and the only alternative on stable Rust is an attribute on every operation in the capability crate, which puts the decision back in the core. It is also what keeps a capability's dependencies out of apps that did not ask: foreign compilers strip unused types, but they do not strip manifest entries.

### Needs from facet-generate

Nothing new. Module companion files — written whenever the module is, unlike `runtime_files`, which installers skip when the serde runtime is external — arrived with the generated `Core`'s BoltFFI bridge in #597.

### Open questions in the text

How shipped handlers should let a shell see their traffic in its own logs; which implementation each capability should ship per platform; and whether a typegen fixture should compile every shipped file rather than relying on the example shells to do it.


docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+After the per-operation types RFC, a shell implements the generated
+`EffectHandler`: one method per effect variant, each with a typed operation in
+and a typed output out. After the generated `Core` RFC, that is all a shell
+writes. But for the operations of a published capability — `crux_http`,
+`crux_kv`, `crux_time` — what it writes is the same in every app: turn an
+`HttpRequest` into a `URLRequest`, map the response and its errors into an
+`HttpResult`; keep a table of timers keyed by id and cancel one on `Clear`.
+The capability author is the one who knows the protocol's rules, and today
+those rules are re-derived by every shell author from the book and the
+examples.
```
charypar: Not _necessarily_. I think this is an opt-in – you may still want to bring your own implementation of HTTP with some silly hardened security proxy thing like banks do. 

I think we should find a way that lets the code gen give you an implementation which you're free to use or not use depending on needs

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+    }),
+    typescript: Some(ShellSource {
+        source: include_str!("../shell/typescript/http.ts"),
+        default: Some("fetchHttpHandler"),
+    }),
+    csharp: Some(ShellSource {
+        source: include_str!("../shell/csharp/Http.cs"),
+        default: Some("HttpClientHttpHandler.Shared"),
+    }),
+};
```
charypar: Nice! I'd personally put it in `lib.rs` of the capability, but... unimportant.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+
+#[derive(Operation, Facet, Serialize, Deserialize)]
+#[operation(request, output = HttpResult, shell(HTTP, method = "request"))]
+pub struct HttpRequest { /* … */ }
```
charypar: That feels like overkill – the author can keep the core and shell sides compatible without us getting involved, right? It's still the same level of type safety as custom capabilities, should be more than enough.

Reading on, I don't love the fact that the core makes a decision about the shell implementation here – if you declare that as your shell handler, the shell _cannot_ override it, because its effect handler now doesn't expose the hooks for HTTP 🤔

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+```
+
+## Why?
+
+### What is left in the shell is mostly protocol, not platform
+
+The weather example's Swift shell, after adopting the generated `Core`, has
+193 lines of handler code. Of those, 153 implement `crux_http`, `crux_kv` and
+`crux_time` — `http.swift` (52), `keyValue.swift` (44) and `time.swift`
+(57) — and would be identical in any other app using those capabilities. The
```
charypar: This is the bit I think I disagree with.

docs/src/rfcs/shipped-shell-handlers.md
charypar: This would also mean that apps choosing to use effect routing would still get an unused HTTP handler generated into the shell code. 

IMO this is not the right place to make the "linking" decisions.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+listening. That last rule was settled in the per-operation types RFC's
+[review](./per-operation-types.md) and is written into a doc comment in the
+weather shell's `time.swift`. It should be written once, by the capability
+author, and shipped.
+
+### The wiring already exists for one operation
+
+The generated `Core` RFC gave `render` a default implementation in the
+generated `EffectHandler`, because `Core` handles that variant itself. That
+default is a protocol extension in Swift, a default body in Kotlin and C# and
```
charypar: Agreed, and I feel _slightly_ uneasy about that, but at minimum I'd treat that as an exception (which it already is, special `view` function, etc...)

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+
+### The default has to be easy to leave and easy to return to
+
+A generated default that is hard to override is worse than no default. Real
+shells configure their HTTP client (certificate pinning, a shared
+`URLSession`, an `OkHttpClient` with interceptors), put their key-value store
+somewhere specific, and sometimes replace a whole capability's implementation
+with one backed by a platform library. Each of those has to be one
+declaration away from the zero-configuration path, and the compiler has to
+point at what is missing when configuration is required.
```
charypar: Unless I'm missing something (which is very possible 😅), I think the current proposal violates this

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+    override suspend fun isLocationEnabled(operation: IsLocationEnabled) = location.enabled()
+}
+```
+
+**Configure.** Override the property with a configured instance of the shipped
+implementation, or any other type conforming to the shipped protocol.
+
+```kotlin
+override val httpHandler = OkHttpHttpHandler(client)          // app's own conformer
+override val keyValueHandler = FileKeyValueHandler(context.filesDir)
```
charypar: I see! 

Okay this wasn't clear to me reading above. I would still say this should not be an override, it should be what the Shell code explicitly does in either case. In other words `httpHandler` becomes a property of the `Handler` which the Shell implementation needs to supply, and it can choose to supply the one delivered by the capability crate.

And that way we don't need extra config on the operation declarations.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+`EffectHandler`; a stateful implementation guards its state with an actor or
+a lock, as the weather shell's timer table does today with `@MainActor`.
+Kotlin implementations are `suspend` and choose their own dispatcher, as the
+weather shell's `withContext(Dispatchers.IO)` does. Nothing about the generated
+`Core` or `EffectDispatcher` changes.
+
+### Rules for shipped source
+
+These are constraints on capability authors, checked by review and by the
+examples' CI rather than by type generation:
```
charypar: Uh yea... this is something we'll need to be quite careful about.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,553 @@
...
+directory — with the module's header prepended by the installer. Unlike
+`runtime_files`, which the installers skip when the serde runtime comes from
+an external package, companion files are written whenever the module itself
+is. They are the right home for shipped sources in Swift, Kotlin and C#, where
+one file per capability is what a reader expects to find. TypeScript's
+single-file modules do not need it.
+
+Everything else uses hooks that exist: `after_type` for the wiring, as the
+handler and `Core` plugins do, and `manifest_dependencies` for the rare
+library.
```
charypar: Super handy, useful for other future features!

charypar (commented):
Nice! I suggest some minor changes, mainly to avoid polluting the core code with considerations that should be shell side

mhedgpeth:
looking through things it looks like y'all have this handled, let me know if you want any careful feedback...in my mind this is a done deal, it's what I saw yesterday. I might be missing something.

charypar:
@mhedgpeth This goes one step further from yesterday – it allows `crux_http` and friends to ship with an actual Swift/C#/Kotlin _implementation_ of HTTP, so you don't need to even bridge to the platform implementation yourself, if you're fine with the default one.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: Agreed. I added another commit with the changes. 

The core now offers a reference implementation and the shell decides whether to emit it (a registration during codegen) and whether to wire it (a conformance on its handler type). The summary now separates the two things the old text ran together: the rules (error mapping, timer semantics), which are protocol and belong with the capability author, and the client, which is the shell's.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: Yep.  The static is now in lib.rs

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: > That feels like overkill 

Yep you're right. 

The binding has moved off the operation entirely. ShellHandler now carries methods: `&[ShellMethod::new::<HttpRequest>("request")]`, so the derive and the trait are untouched and the operation declaration says nothing about shells. Some binding is still needed, because EffectHandler's method names come from the app's variant names, which the capability can't know. But it's protocol metadata in the static, not a decision on the operation.

On the second point: you were right that the default made the core decide. There is no default now. The generated `EffectHandler` is unchanged (every method abstract); what's generated alongside is an adapter protocol (`HttpHandling`) the app may conform to. Conform and you supply httpHandler; don't and the compiler asks for http(_:) as it does today.

StuartHarris: > This would also mean that apps choosing to use effect routing would still get an unused HTTP handler generated into the shell code.
> 
> IMO this is not the right place to make the "linking" decisions.

Fixed by making emission opt-in per capability: .shell_handler(&crux_http::HTTP) on the CodeGenerator, next to .boltffi(...). An app that doesn't register it gets no source, no adapter and no manifest dependency. Registering one together with without_effect_handlers() is an error, mirroring check_boltffi.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: Reframed. The section is now "The rules are protocol; the client is platform", and it says explicitly that which HTTP client performs the request is the shell's decision and that a shell with a hardened proxy or pinned session should keep it. What the capability author owns is the mapping onto HttpResult and the timer semantics. The line counts stay as evidence that those rules are currently re-derived per shell, not as a claim that every shell wants the same client.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: Agreed. The section arguing from the render default is gone; it's now listed under Alternatives as "Generalising the render default", rejected with your reasoning.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: It did. With the default gone, the tiers are: conform and supply the shipped instance (Use); supply a configured or app-owned instance (Configure); implement one method over the conformance (Replace one method); drop the conformance (Replace the capability); don't register it (Do not use). Each is one declaration, and the compiler names what's missing at each step.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: This is essentially wha's in the new commit, with one refinement. Rather than putting httpHandler on EffectHandler itself, which would make an app that implements http by hand declare a dead property, it goes on a generated HttpHandling: EffectHandler adapter that carries the delegating body. The app opts in by conforming and must supply the property; an app that doesn't conform never sees it. And yes: nothing on the operation declarations. The binding is in the ShellHandler static alongside the source.

One open question I've left in the RFC: whether HttpHandling should default httpHandler to the shipped implementation so conforming alone is enough. Swift could do it from the shipped source; Kotlin and C# would need the expression back in the static. I've leaned towards keeping the property required for uniformity, but interested in your view.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: Agreed. I've added why it matters (a broken shipped file breaks every app that registers it), a rule that configuration goes through the implementation's initialiser rather than globals, and made "compiled by an example shell on every change" an explicit rule rather than an aside. The dependency rule is now tied to ShellSource::dependencies, so a library only reaches manifests of apps that registered the handler.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,622 @@
...
+
+```rust
+pub struct ShellHandler {
+    /// UpperCamelCase. Names the shipped protocol (`HttpHandler`,
+    /// `IHttpHandler` in C#), the generated adapter (`HttpHandling`,
+    /// `IHttpHandling`) and the property the adapter requires (`httpHandler`).
+    pub name: &'static str,
+    /// Which operations the shipped protocol performs, and through which
+    /// method.
+    pub methods: &'static [ShellMethod],
```
charypar: I wonder if this is going to become quite tricky and we should accept the compromise of just generating the lot. I would think the foreign languages will compile the unused code out anyway?

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,622 @@
...
+The app names the shipped handlers it wants on the built `CodeGenerator`,
+next to `boltffi`:
+
+```rust
+let typegen = TypeRegistry::new()
+    .register_app::<Weather>()?
+    .build()?
+    .boltffi(BoltFfi::new().swift("Shared").kotlin())
+    .shell_handler(&crux_http::HTTP)
+    .shell_handler(&crux_time::TIME);
```
charypar: This is becoming really nice 🙂

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,622 @@
...
+
+## Open questions
+
+1. **A default within the adapter.** Should `HttpHandling` supply
+   `httpHandler` itself, defaulting to the shipped implementation, so that
+   conforming alone is enough and the property is only written to configure?
+   Swift could do it from the shipped source, with an extension on the
+   generated protocol; Kotlin and C# would need the default expression back
+   in the static, because it has to appear in the interface declaration.
+   Uniformity across languages argues for leaving the property required.
```
charypar: I agree, and explicit is better IMO.

docs/src/rfcs/shipped-shell-handlers.md
```diff
@@ -0,0 +1,622 @@
...
+2. **Logging.** Shells want to see HTTP traffic and timer activity in their own
+   logs. A shipped protocol could carry an optional observer, or the shipped
+   implementations could take a logger in their configuration. Left to the
+   first implementation.
```
charypar, line 594: hmm, IMO this could be be an optional behaviour of the default implementation later on. Wouldn't worry about it just yet. The effect observability in the core will also change the equation for this

charypar (approved):
Nice, this feels right now.

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: ahh good point. yes let's leave it for later

docs/src/rfcs/shipped-shell-handlers.md
StuartHarris: I think I agree with you, so I've stripped it back quite a lot (which is probably a good thing, and it's more explicit, less magical).

In the weather shell the adapter turned three conformances and three properties into five delegating lines and three properties – it saved almost nothing, and cost a method listing in every capability, TypeId plumbing, registration checks and four adapter emitters. Not worth it.

The RFC is now source-only. ShellHandler is a name and four optional sources. Typegen copies the shipped file into the generated module (and adds its manifest dependency, if it has one) but doesn't generate anything else (EffectHandler, EffectDispatcher and Core are untouched). The shell holds an instance of the shipped implementation and delegates, one line per operation:
```
  let http = URLSessionHttpHandler.shared
  func http(_ operation: HttpRequest) async -> HttpResult { await http.request(operation) }
```

Those lines are also the better behaviour: when a capability gains an operation the compiler asks for a line, instead of a regenerated file wiring it silently.

The one thing I've kept is the registration in the codegen binary (.shell_handler(&crux_http::HTTP), next to .boltffi(...)). Typegen has to discover the static somehow, and the only alternative on stable Rust is shell = HTTP on every #[operation] in the capability crate, which puts the decision back in the core. 

Whilst foreign compilers may strip unused types, they won't strip manifest dependencies, so the registration line keeps libraries a capability might ship out of apps that didn't ask for it. Instead there's one line per capability, shell-side, and the checks reduce to a name collision.

Both earlier shapes are written up under Alternatives, with your reasoning, so the next reader doesn't retread them.

PR #597, opened 2026-09-16
feat(typegen): bridge the generated Core to BoltFFI's CoreFfi

StuartHarris:
Stacked on #594. Follows up the review of #590, where @charypar asked why the examples bridge the generated `Core` to BoltFFI's `CoreFfi` by hand ("the shell would just get a `Core` class it can instantiate, which expects an effect handler").

## What this does

Type generation cannot discover BoltFFI's output, but it can be told where it is. `CodeGenerator::boltffi(BoltFfi)` names the exported class and, per language, the module, package or namespace it lives in:

```rust
TypeRegistry::new().register_app::<Weather>()?.build()?.boltffi(
    BoltFfi::new()
        .swift("Shared")   // import Shared; the package is at ../Shared
        .kotlin()          // CoreFfi is in the generated package
        .typescript("shared", PackageLocation::Path("../pkg".into())),
)
```

For each named language the generated module gains an `FfiBridge` (`CoreBridge` over `CoreFfi`, with the Swift `Data` conversion and the `@unchecked Sendable` declaration where they belong) and `Core` gains a constructor that takes only the handler:

| | |
|---|---|
| Swift | `Core(handler: WeatherHandler())` |
| Kotlin | `Core(handler, scope)` |
| TypeScript | `await Core.create(handler, setView)` |
| C# | `new Core(handler)` |

TypeScript's is an async factory because the wasm module loads asynchronously; `Core.create` awaits BoltFFI's untyped `initialized` promise once, in generated code, instead of every shell casting it. `CoreBridge` and the two-argument constructors are still emitted, so previews (weather's `PreviewBridge`) and the middleware examples, whose `CoreFfi::new` takes a callback, keep writing their adapters. A `boltffi(..)` with no `Core` to bridge is an error, not a no-op. `FfiBridge` is reserved.

`FfiBridge` is written as a companion file in the module (Swift, Kotlin, C#), so the module's main file never imports the FFI; TypeScript's single-file module gets it inline.

## Build consequences

- **Swift.** The generated package now depends on BoltFFI's (`.package(path: "../Shared")` plus the product on the target) and so needs a `platforms:` floor no lower than that package's. That is a new `Config::builder(..).platform(".iOS(.v16)")` setting, useful with or without the bridge. The weather app target stops linking `Shared` directly.
- **TypeScript.** The generated `package.json` depends on the wasm package, so `boltffi pack wasm` runs before typegen; the weather and notes web Justfiles were reordered (Android already packed first).

## facet-generate

Needs redbadger/facet-generate#124 (stacked on #123): a `companion_files` plugin hook, Swift `manifest_dependencies`/`target_dependencies`, `Config.platforms`, and TypeScript `manifest_dependencies`. The pinned rev is bumped here. The companion-file hook is the one the shipped-shell-handlers RFC (#596) proposes; that RFC's text is amended on top of this PR to say it exists.

## Examples and docs

Weather (Swift, Kotlin, TypeScript) and notes (TypeScript) lose their `LiveBridge` files. The typegen chapter gains a "Bridging to BoltFFI" section; the iOS, Android and React chapters and the migration guide show the one-argument constructors; the generated-core RFC records the bridged constructor as an opt-in, the alternatives it displaces and the drawbacks above.

## Naming

Viktor suggested "effect delegate" for the Swift surface. This keeps `EffectHandler` and `Core(handler:)`: the name is in the merged per-operation types RFC and across the stack in four languages, and the property a Swift reader wants — a class that takes one collaborator and nothing about Rust in view — holds either way.

## Not here

The Rust side of the FFI (`ffi.rs`) stays hand-written until BoltFFI can see macro-emitted exports (boltffi/boltffi#665).


crux_core/src/type_generation/facet/plugins/snapshots/crux_core__type_generation__facet__plugins__tests__core_swift_with_boltffi.snap
```diff
@@ -0,0 +1,67 @@
...
+                self.process(bytes: self.bridge.resolve(id, bytes))
+            }
+        }
+    }
+
+    /// Build a `Core` over the `CoreFfi` the Rust crate exports — the shell
+    /// writes the handler and never sees the bridge.
+    public convenience init(handler: any EffectHandler) {
+        self.init(bridge: FfiBridge(), handler: handler)
+    }
```
charypar, line 45: Neat 👌🏻

charypar (approved):
Now we're talking! 🙌🏻

PR #608, opened 2026-09-18
fix(typegen): reject duplicate generated type names

### LINKED ISSUE #601: Type registry silently lets a same-named type replace another
Registering two root-namespace types that generate under the same name does not error: one silently wins and the other is never emitted.

Found while implementing shipped shell handlers (#596 stack): registering `crux_kv`'s full operation set alongside the weather example's own `secret::Delete` produced a single generated `Delete` (the secret one, with `value: String`); `crux_kv::operation::Delete { key }` vanished from every generated module with no diagnostic. The shells then fail to compile in ways that point nowhere near the cause.

Expected: `TypeRegistry::build` (or facet-generate's registry) reports the collision and names both Rust types, with the `#[facet(rename = "...")]` fix, the way reserved-name collisions are reported today.

TayfurYldz:
Fixes #601

## Problem
Registering two distinct root Rust types that generate to the same qualified type name lets one silently replace the other in the generated registry.

## Root cause
`facet_generate::RegistryBuilder` keys both its processed set and registry by the generated qualified name. Once one root type claims that name, a later distinct Rust type with the same generated name is treated as already processed, so Crux loses the second type without a diagnostic.

## Fix
Track the generated qualified name for each root type registered through `TypeRegistry` and reject a second distinct Rust type claiming the same name. The error names both Rust types and points to `#[facet(rename = "...")]` / namespaces as the resolution. The tracking state is cleared when `build()` consumes the current registry.

## Tests
- Current `origin/master` (`30337489b85f43ee24a4db69b94ad9c94c305da5`) + regression test: fails as expected (`duplicate generated names should be rejected`).
- `cargo test -p crux_core --test typegen duplicate_generated_type_names_are_rejected -- --nocapture`
- `cargo test -p crux_core --test typegen -- --skip facet_test::generates_typescript` — 9 passed; TypeScript generation test skipped because `pnpm` is not installed in this environment.
- `cargo test -p crux_core --lib` — 67 passed.
- `cargo clippy -p crux_core --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`


StuartHarris:
Thanks for this @TayfurYldz, and for the clear root-cause write-up. You found exactly where the drop happens: `RegistryBuilder` keys its processed set by generated name, so a second Rust type with the same name looks like a recursive visit and is skipped.

I have taken the fix to that spot rather than guarding it in `TypeRegistry`, because the guard here only sees types passed to `register_type` directly. A collision reached through a field or a variant, which is the shape that bit us in #601 (the kv operations arrive through the handler's type set, the app's `Delete` through the app), still vanishes with this PR applied. The builder is the only place that sees every type, and it already has what it needs to tell "same type again" from "different type, same name", which is facet's `Shape::id`.

The upstream change is redbadger/facet-generate#137. Every generated name now records the Rust type that claimed it, and a different type claiming it returns `Error::DuplicateTypeName`, naming both types by Rust path with the two fixes (`#[facet(rename = "...")]` or `#[facet(fg::namespace = "...")]`). It covers direct and transitive registrations, root and named namespaces, and collisions through `rename`, `type_tag` and field-level namespace overrides. The same identity check fixes two neighbours you may have noticed while you were in there: a `rename` on one type no longer applies to an unrelated type with the same identifier, and `a::Foo<A>` beside `b::Foo<B>` is reported as a name collision rather than an unsupported generic.

crux picks it up in #611, which bumps the pinned rev and adds two regression tests in `facet_clash_test`. One of them is your test shape, two root-registered `Delete`s in sibling modules, with credit to this PR in the description. The other is the transitive shape, a registered type whose field holds the colliding `Delete`.

So I am going to close this one in favour of #611. The diagnosis was right, and it is what pointed the fix at the correct layer.

Thanks again for your contribution.

